PR #5, opened 2014-09-10
Updated instructions for Debian (and derivates)

naoliv:
A more up-to-date instruction to install osmose-backend


jocelynj:
frodrigo just updated README.md with instructions and command lines for installation. So I think it is no longer necessary to add another file with instructions.

PR #6, opened 2014-09-10
Update utilities path and config

naoliv:
Make it possible to configure all utilities inside modules/config.py only.
Use system apps by default (with alternative examples when the user don't have the packages installed)


jocelynj:
I've ported part of these commits on 6bdb81d21296939d050df1c1e00ffe4a0b8d8c09.
The remaining part is how to use system binaries in priority - I'm wondering if putting this code in modules/config.py is the right place.

PR #42, opened 2014-12-22
wait much longer for osmosis import lock

tkas:
For paralell import, it is OK to wait quiet long for import lock.


jocelynj:
Waiting for more than 2 hours to get access to osmosis database is a bit long. I won't apply this commit except if it is really necessary.

PR #55, opened 2015-05-24
Allow maxspeed=implicit if traffic_sign=maxspeed

tyndare:
plugins/Number.py

According to the wiki, maxspeed=implicit is a valid value
when used with traffic_sign=maxspeed:
http://wiki.openstreetmap.org/wiki/Key:traffic_sign

### DIFF
--- a/plugins/Number.py
+++ b/plugins/Number.py
@@ -36,7 +36,12 @@ class Number(Plugin):
         for i in self.tag_number:
             if i in tags:
                 m = self.Number.match(tags[i])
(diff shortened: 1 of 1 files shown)


jocelynj:
Is it possible to rebase this change on master ?

frodrigo:
This patch is already on master

jocelynj:
Thanks !

it4workflow:
according to https://wiki.openstreetmap.org/wiki/Speed_limits#Signage `maxspeed=default` is a valid value, too

PR #60, opened 2015-07-10
Move Mexico to North America

naoliv:


jocelynj:
Thanks, I committed the change on 831a54c8812e831f78819e04e5043002de0072e7.

PR #64, opened 2015-07-18
Create Name_PoorlyWrittenWayType_ca.py

jmontane:


jmontane:
Fixed id numbers. Thanks!

jocelynj:
It doesn't seem to pass tests. The first issue is that the class name is incorrect. I fixed this, but it fails on another error:
```shell
$ nosetests plugins/Name_PoorlyWrittenWayType_ca.py
E
======================================================================
ERROR: test (plugins.Name_PoorlyWrittenWayType_ca.Test)
----------------------------------------------------------------------
  File "/home/jocelyn/osm/osmose/backend/plugins/Name_PoorlyWrittenWayType_ca.py", line 59, in test
    a.init(None)
  File "/home/jocelyn/osm/osmose/backend/plugins/Name_PoorlyWrittenWayType_ca.py", line 45, in init
    self.ReTests[(207, u"Plaça")]    = self.generator(u"^([P][Ll]([aA][çÇzZ][aA])?\.?) .*$")
  File "/home/jocelyn/osm/osmose/backend/plugins/Name_PoorlyWrittenWayType.py", line 29, in generator
    (p1, p2) = p.split("|")
ValueError: need more than 1 value to unpack

----------------------------------------------------------------------

```

frodrigo:
I think the part detecting Spanish and converting must go else where. This analyser must deals only with Catalan.
Nevertheless, there is an other analyser that check default language is the same in name and name[default] when available.
So, let start without the converting part, activate all the language analyser with your other pull request, and then add a new specific analyser latter if needed.

PR #65, opened 2015-07-27
use names for SK kraje without diacritics to avoid troubles

tkasp:
Using diacritics will cause troubles, I suggest using names without diacritics the same as for CZ.

### DIFF
--- a/osmose_config.py
+++ b/osmose_config.py
@@ -894,14 +894,14 @@ class sk_kraj(default_country):
         default_country.__init__(self, "europe", "slovakia/" + province, polygon_id, analyser_options,
                                     download_repo, download_country)
 
-sk_kraj("trnavský", 388266)
-sk_kraj("trenčiansky", 388267)
-sk_kraj("prešovský", 388271)
+sk_kraj("trnavsky", 388266)
+sk_kraj("trenciansky", 388267)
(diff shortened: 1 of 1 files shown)


jocelynj:
What kind of issues do you see with diacritics ?

I have hesitated between using diacritics or not, but choose to use them. The reason is that we will soon add Russia, and I think it makes more sense to use Cyrillic alphabet instead of a transliteration.

tkasp:
2015-07-27 18:16 GMT+02:00 jocelynj <notifications@github.com>:

> What kind of issues do you see with diacritics ?
>
> I have hesitated between using diacritics or not, but choose to use them.
> The reason is that we will soon add Russia, and I think it makes more sense
> to use Cyrillic alphabet instead of a transliteration
>
Then I would strongly recomend to go without diacritics as well as with
latin alphabet usage for foreign countries (e.g. china, japan etc.). The
names of country parts transferes into several file names (extracts, logs
etc.) which you have to work with on the server. If you are not able to
write those characters on the server, how would you do that. eg.


ls -l czech_republic-* vs. ls -l česká_republika_*

tail -n2 polska_

*Łódzkie.log *

*etc.*

jocelynj:
You're right, it is better to use ASCII only here.

PR #94, opened 2016-02-09
Correct and add to German translations

sleske:
Some corrections and additions to the German translations.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -56,7 +56,7 @@ msgstr "Verschobener Vermessungspunkt"
 
 #: ../analysers/analyser_merge_geodesie.py:65
 msgid "Missing survey site"
-msgstr ""
+msgstr "Fehlende Gruppe ('site') von Vermessungspunkten"
 
 #: ../analysers/analyser_merge_hydrant_point_CH_lausanne.py:28
 msgid "Hydrant not integrated"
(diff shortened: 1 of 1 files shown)


jocelynj:
Thanks for the translation.

We use Transifex to handle translations in fact. Can you add your translations to https://www.transifex.com/openstreetmap-france/osmose/dashboard/ instead ?

sleske:
Translations added in Transifex. I just added them using the online editor.
Do I need to do anything else? I did not see any "publish" button or similar.

sleske:
And BTW, maybe you could inform contributors that you do not accept pull requests for translations.
You could add a CONTRIBUTING.md, as suggested by Github ( https://help.github.com/articles/setting-guidelines-for-repository-contributors/ ). Or just mention it in README.md .

jocelynj:
Thanks!

I check translations at least once per month, or when somebody directly requests an update. It is true that there is no button on Transifex to request an update.

Your CONTRIBUTING.md file looks a good idea. The only information on contribution that we currently have is on http://osmose.openstreetmap.fr/en/translation , which is available on the Help menu.

jocelynj:
Your changes were pushed on ca2251af78d96f6dbb6cf3f9de7402c70b46ed85.

sleske:
@jocelynj Thanks, that was quick.

PR #124, opened 2016-08-07
Fix insertion into official tables (*_off)

otrichet:
The commit 2b47b7b87043c broke the insertion of data into so called "official table" because Analyser_Osmosis.run0() can not be called from its callback.

The consequence is that inexistent error get reported.
E.g. http://osmose.openstreetmap.fr/fr/map/#zoom=17&lat=50.5476813&lon=3.0393326&item=7150&level=3

http://osmose.openstreetmap.fr/fr/errors/graph.png?source=7406&item=7150&class=2

### DIFF
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -539,6 +539,7 @@ class Load(object):
             osmosis.run(sql_schema % {"schema": db_schema})
             osmosis.run(sql00 % {"schema": db_schema, "official": tableOfficial})
             giscurs = osmosis.gisconn.cursor(cursor_factory=psycopg2.extras.DictCursor)
+            giscurs_getpoint = osmosis.gisconn.cursor(cursor_factory=psycopg2.extras.DictCursor)
             def setLonLat(lonLat):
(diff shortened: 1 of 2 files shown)


frodrigo:
Good catch !

I think it's linked to this issue:
https://lists.openstreetmap.org/pipermail/talk-fr/2016-August/081661.html

About "Avoid useless call to polygons.openstreetmap.fr", the call is mandatory as it generates the seconde ressource if it's not already in cache.

otrichet:
Hello,

What is the next step ?
* Do you want me to change this PR to only include the "insert into official tables" fix ?
* Do you prefer to make the method Analyser_Osmosis.run0() reentrant ?
* You just have no time right now to review/test/push this change ?

frodrigo:
I picked up the first commit and add some improvement.
https://github.com/frodrigo/osmose-backend/commit/40c044b9a2daa8b4780b51d2ca8aa50d42640aa3
https://github.com/frodrigo/osmose-backend/commit/500a2a6337acc3b171949ad81b7fd49776730729
We are waiting @jocelynj to merger all of this into master.

Issue #136, opened 2016-09-01
mhs:inscription_date: please check date format


treinen:
Hello,

osmose doesn't seem to check the format of the date in the field mhs:inscription_date,
for instance here:

http://www.openstreetmap.org/way/200452259

at least this one, and other ones in the format YYYY/MM/DD, are not listed under 3090-date

frodrigo:
Yes, It does not. And I don't think it will.

treinen:
Why? I thought all dates should be ISO 8601, according to

http://wiki.openstreetmap.org/wiki/Editing_Standards_and_Conventions#Date_and_time

frodrigo:
> Why?
Cause it's too much complicated to do this, IMHO.

Feel free to study the code, and make a pull request yourself if you want.
https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_merge_heritage_FR_merimee.py

frodrigo:
Hint : https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Merge.py#L876

jocelynj:
@treinen: can you check which item the error you have seen has ?  Your fix on #138 would fix this issue only if error is on item 3090. If error is on item 8011 or 8012, then it is file  analyser_merge_heritage_FR_merimee.py or Analyser_Merge.py that should be fixed.

frodrigo:
@jocelynj for 801x it's in #137

treinen:
when there is something like "mhs:inscription_date=2010/12/31" then osmose *should* raise a 3090,
but currently it doesn't 

In order to fix #137 we need that the date is in standard format,so that we can extract the year from the date. So, in some sense this is a prerequisite for fixing #137.

PR #138, opened 2016-09-02
plugin/Date: also apply to field mhs:inscription_date

treinen:
Hello,

this is untested (I didn't yet figure out how to set up a test environment) but this should solve issue #136. Just include mhs:inscription_date in the list of tags checked by the Date plugin.

-Ral.f


frodrigo:
@jocelynj OK, can be merged

PR #154, opened 2016-10-21
update postgis functions to be compatible with postgis 2.1+

tkas:
Old names are still available in 2.3.0, but issue a warning and seems to be very suboptimal as of performance.

### DIFF
--- a/analysers/analyser_osmosis_broken_highway_level_continuity.py
+++ b/analysers/analyser_osmosis_broken_highway_level_continuity.py
@@ -119,7 +119,7 @@ FROM
 WHERE
     o1.nid != o2.nid AND
     o1.level = o2.level AND
-    ST_Distance_Sphere(o1.geom, o2.geom) < 1000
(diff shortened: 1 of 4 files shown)


frodrigo:
@jocelynj This change will break compatibility with PostGis 1.5, I think we still have backend on it, doesn't ?

tkas:
If so, then leave it just for the moment it could be merged into mainstream. I will keep it localy if necessary for own installation.

jocelynj:
We still have some backends using postgis 1.5, but we should be able to upgrade them.

Just wondering: are these changes required by an existing version postgis ? Otherwise, will they be required soon ?

tkas:
As I wrote, they are required by postgis version 2.1 and newer. There is some emulation but it does not work well.

stephankn:
Docker image is using postgis 2.4, warnings are still around. Any reason to still keep legacy names?

frodrigo:
Yes, because we are still sticky with postgis 1.5 on some backbend servers.

PR #160, opened 2016-11-10
Update analyser_osmosis_building_shapes.py

sidjy:
Several round buildings are dovecote...Osmose should propose this as a choice


frodrigo:
@jocelynj ok for me

PR #167, opened 2016-12-19
Analyser merge_public_transport_FR_stif

nlehuby:
ZDE stands for "Zone d'Embarquement : endroit où le voyageur attend, monte et descend du véhicule". 
This is the exact definition of a public_transport = platform.

And some wording improvements.


frodrigo:
@jocelynj ok for me

jocelynj:
Thanks for the update. It should appear in the following days on osomse.openstreetmap.fr

PR #172, opened 2016-12-23
 analyser_merge_street_number: normalize for better matching.

tyndare:
Normalize addr:housenumber France's format for better matching
when doing merge of open data with OSM data.

The reason is the multiplicity of "valid" format (e.g. "bis" or "B", with space separator or not)

https://lists.openstreetmap.org/pipermail/talk-fr/2016-December/082920.html


matkoniecz:
@osm-fr Can you consider either merging or rejecting or reviewing this PR?

PR #201, opened 2017-05-21
add analyse for public_transport:version tag on pt route relations

nlehuby:
I did not test it. I copy the implementation logic from other analyses from this directory.

Additional questions : 
* How I am supposed to choose the class id (I pick 9 totally randomly)
* Do I need to edit the translation files too ?

### DIFF
--- a/plugins/Public_transport_route.py
+++ b/plugins/Public_transport_route.py
@@ -1,0 +1,74 @@
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank you. I fixed the author name, the item and class id and check if route and type are existing tags.
The commit is now here:
https://github.com/frodrigo/osmose-backend/commit/be6a2b2697e9ce2d80f6c376e53448bc8fac5481

Note: you can run the unit test like this:
```
$ nosetests Public_Transport_Toute.py
```

PR #210, opened 2017-06-14
Add sax test on public_transport tag

nlehuby:


plugins/Public_Transport_Stop.py
```diff
@@ -0,0 +1,65 @@
...
+    def node(self, data, tags):
+        err = []
+        if u"public_transport" not in tags:
+            if "highway" in tags and tags["highway"] == "bus_stop":
+                err.append({"class": 21411, "subclass": 0,
+                            "text": T_("Bus stop without public_transport tag"),
+                            "fix": {"public_transport": "platform"}
+                            })
+            if "railway" in tags and tags["railway"] == "tram_stop":
+                err.append({"class": 21411, "subclass": 0,
```
frodrigo: You can set subclass to 1 here to distinguate with previous one.

plugins/Public_Transport_Stop.py
```diff
@@ -0,0 +1,65 @@
...
+###########################################################################
+
+from plugins.Plugin import Plugin
+
+
+class Public_Transport_Stop(Plugin):
+
+    def init(self, logger):
+        Plugin.init(self, logger)
+        self.errors[21411] = {"item": 2141, "level": 3, "tag": ["tag", "public_transport", "fix:chair"], "desc": T_(
```
frodrigo: We can stay with 2140 here. We try to group classes on menu items on the front.

nlehuby:
I still mix up class, subclass and item :confused:

frodrigo:
For me also it was not so easy ;-).

Categories are group of items.
Categories and items are for display and filter purpose on the frontend.

Subclass are subelement of class (yes really).
Class and subclass are for distinguish the analyses results.

Class are set into item by thematic, one anayser can produce different classes set in different items.

frodrigo:
@jocelynj PR good for me

jocelynj:
Thanks for the merge request - I've just merged it to master.

PR #214, opened 2017-07-04
Make osmose-backend runnable in Docker

dvenza:
This pull request adds the possibility of running osmose inside a Docker container.
It is useful for:

* testing, because the Docker image guarantees that osmose is run in an always fresh environment
* running concurrently, each container comes with its own postgres instance

The image is built with `osmose_run.py` as the entrypoint. This means that the container can be used as `osmose_run.py` itself: you pass the same command-line options at the end of the `docker run` command and they will be passed on to osmose.
See the examples at the end of the README file for a short how-to.


frodrigo:
Nice to see this.
You also need to map the password file into the container to be able to upload the result to the frontend.

dvenza:
I added a command line variant that passes the password file to the container to the README instructions.

We can also have the image autoamtically built at every commit on this repository by the Docker Hub.

Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM ubuntu:14.04
```
jocelynj: Is there a reason to use an old version of Ubuntu ?

This means that postgresql is using an old version too, which could be more recent.

frodrigo: The last LTS is 16.04

dvenza: The only reason is habits, it's the image I use more often. I will test with 16.04 and add a new commit.

frodrigo:
@jocelynj ok for me

PR #215, opened 2017-07-10
fix analyser relation_public_transport for missing route_master

prhod:
This PR adds errors for `route` relation included in a non `route_master` relation.


frodrigo:
@jocelynj ok for me

PR #216, opened 2017-07-23
update public_transport analysers

nlehuby:
* public transport modes unification on both analysers
* wording enhancements


frodrigo:
@jocelynj OK for me

PR #217, opened 2017-07-23
Add analyser for public transport stop without legacy tag

nlehuby:


frodrigo:
Please, fix the `def way` if needed.

nlehuby:
I don't think it is needed.

public_transport = platform on ways are usually not bus_stop or tram_stop (but actual platform)
and highway = bus_stop on ways are usually over-mapped shelters (building = yes + wall = no). And the wiki said that this tag should not be used on ways.

jocelynj:
Thanks, I've just pushed this change to all backends with tests and translation updates.

PR #237, opened 2017-10-24
install git in the image

stephankn:
The version reported to the front-end ist constructed from git output. Without git being available it would be missing, so include it.


jocelynj:
Thanks, this looks perfectly fine. I will merge this soon.

Pinging @dvenza in the meantime.

dvenza:
Yes, looks good to me too!
Thanks

PR #242, opened 2017-11-03
Analyser Merge Transgironde - fix

nlehuby:
most stops are on the side of the road (and not in the middle of the road) so they seems to represent platforms


frodrigo:
Thank you.

PR #243, opened 2017-11-05
PT osmosis Analyser - Stop too far

nlehuby:
change distance threshold to limit false positive


nlehuby:
With these threshold, I have no false positives at all :) (I've tested the whole ghana region, and a few ones in france_ile_de_france)

PR #248, opened 2017-11-19
Analyser Merge TBM - fix

nlehuby:
most stops (from the opendata source and from OSM) are on the side of the road (and not in the middle of the road) so they seems to represent `public_transport`= `platform`


frodrigo:
@jocelynj @nlehuby whnat to display first the result on a map (so the support of passwordless and Docker).

PR #252, opened 2017-12-13
Analysers for open data toilets in France

PanierAvide:
Added other cities to merge toilets open data : Lyon, Nantes, Toulouse, Angers, Le Havre, Montpellier


analysers/analyser_merge_public_equipment_FR_lehavre_toilets.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8180", "class": 7, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Le Havre toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "https://data.agglo-lehavre.fr/",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Ville du Havre", millesime = "12/2017",
+                    fileUrl = "https://data.agglo-lehavre.fr/api/v1/file/data/159/SANITAIRE/json", zip = "OD.SANITAIRE.json"),
+                extractor = lambda json: json['features']),
+            Load("geometry", "geometry",
+                xFunction = lambda g: g and g.coordinates and g.coordinates[0],
```
frodrigo: the "." is not a valid syntax in python.
Use something like g['coordinates']

PanierAvide:
Fixed the syntax

analysers/analyser_merge_public_equipment_FR_lehavre_toilets.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class Analyser_Merge_Public_Equipment_FR_LeHavre_Toilets(Analyser_Merge):
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8180", "class": 7, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Le Havre toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "https://data.agglo-lehavre.fr/",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Ville du Havre", millesime = "12/2017",
+                    fileUrl = "https://data.agglo-lehavre.fr/api/v1/file/data/159/SANITAIRE/json", zip = "OD.SANITAIRE.json"),
```
frodrigo: Please, can you switch to CVS, the JSON multi-level level not works: g['coordinates'][0]

frodrigo: Maybe we add a geojson parser
https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Merge.py#L398

analysers/analyser_merge_public_equipment_FR_lyon_toilets.py
```diff
@@ -0,0 +1,50 @@
...
+        self.missing_official = {"item":"8180", "class": 3, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Lyon toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "https://data.grandlyon.com/equipements/toilettes-publiques-sur-le-territoire-du-grand-lyon/",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Métropole de Lyon", millesime = "12/2017",
+                    fileUrl = "https://download.data.grandlyon.com/wfs/grandlyon?SERVICE=WFS&VERSION=2.0.0&outputformat=GEOJSON&request=GetFeature&typename=gin_nettoiement.gintoilettepublique&SRSNAME=urn:ogc:def:crs:EPSG::4326"),
+                extractor = lambda json: json['features']),
+            Load("geometry", "geometry",
+                xFunction = lambda c: c and c['coordinates'] and c['coordinates'][0],
+                yFunction = lambda c: c and c['coordinates'] and c['coordinates'][1]),
```
frodrigo: Same, json multi level not supported

analysers/analyser_merge_public_equipment_FR_montpellier_toilets.py
```diff
@@ -0,0 +1,52 @@
...
+        Analyser_Merge.__init__(self, config, logger,
+            "http://data.montpellier3m.fr/dataset/toilettes-publiques-de-montpellier",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Montpellier Mediterranée Métropole", millesime = "12/2017",
+                    fileUrl = "http://data.montpellier3m.fr/sites/default/files/ressources/MMM_MTP_WC_PUBLICS.json", encoding = "ISO-8859-15"),
+                extractor = lambda json: json['features']),
+            Load("geometry", "geometry",
+                where = lambda res: res.enservice == u"En Service",
+                xFunction = lambda g: g and g['x'],
+                yFunction = lambda g: g and g['y']),
```
frodrigo: json, same again.

analysers/analyser_merge_public_equipment_FR_nantes_toilets.py
```diff
@@ -0,0 +1,51 @@
...
+        self.missing_official = {"item":"8180", "class": 5, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Nantes toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "http://data.nantes.fr/donnees/detail/toilettes-publiques-de-nantes-metropole/",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Nantes Métropole", millesime = "11/2016",
+                    fileUrl = "http://data.nantes.fr/api/publication/24440040400129_NM_NM_00170/Toilettes_publiques_nm_STBL/content/?format=json"),
+                extractor = lambda json: json['data']),
+            Load("_l", "_l",
+                xFunction = lambda c: c and c[1],
+                yFunction = lambda c: c and c[0]),
```
frodrigo: CSV is the best choice ;)

PanierAvide:
Well, I already took all possible CSV from various websites, the ones using JSON are because open data site doesn't offer CSV, or CSV is missing some information like coordinates. Implementing a GeoJSON parser seems right, I will look to do so. Still some data sources offers JSON which aren't GeoJSON, needing to handle multi-level attributes reading. I will try to find how to manage this also.

PanierAvide:
I wrote the GeoJSON parser and used it on Le Havre. As I can't test results on my side (only checking syntax errors), I'm not sure this will do the job.

PanierAvide:
Thanks to the server access you opened to me, I was able to test and fix all the analysers. They are now working properly. I introduced GeoJSON reader, which is more or less like JSON one but also adding in properties coordinates as `geom_x` and `geom_y` columns (only works for Point features). And for JSON analysers based on custom data formats, I used `json.loads` method (plus edits in JSON reader) in order to handle multi-level properties.

analysers/Analyser_Merge.py
```diff
@@ -409,14 +414,46 @@ def __init__(self, source, extractor = lambda json: json):
 
     def header(self):
         self.json = self.extractor(json.loads(self.source.open().read()))
-        return self.json[0].keys()
+        columns = map(lambda c: unaccent(c), self.json[0].keys())
+        return columns
```
frodrigo: Why do you need to use unaccent ?

PanierAvide: Because one of the JSON dataset has a field named "Pôle", with the ô causing some encoding trouble.

analysers/analyser_merge_public_equipment_FR_montpellier_toilets.py
```diff
@@ -0,0 +1,53 @@
...
+        self.missing_official = {"item":"8180", "class": 6, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Montpellier toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "http://data.montpellier3m.fr/dataset/toilettes-publiques-de-montpellier",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Montpellier Mediterranée Métropole", millesime = "12/2017",
+                    fileUrl = "http://data.montpellier3m.fr/sites/default/files/ressources/MMM_MTP_WC_PUBLICS.json", encoding = "ISO-8859-15"),
+                extractor = lambda json: json['features']),
+            Load("geometry", "geometry",
+                where = lambda res: json.loads(res['attributes'])['enservice'] == u"En Service",
+                xFunction = lambda g: g and json.loads(g)['x'],
```
frodrigo: Can we flat data in load ? Having column like "namelvel1.sublevel", in place of running lots of json parsing ?

PanierAvide: Maybe, I can use a flattenjson implementation like this one : https://stackoverflow.com/a/28246154

analysers/Analyser_Merge.py
frodrigo: It should work. We already use lots raw column names.
Renaming the column is not in the spirit of the process. It's make the the code of the analyzer less understandable and more magic.

analysers/analyser_merge_public_equipment_FR_montpellier_toilets.py
frodrigo: Yes. (I personally find the "." more intuitive.)

analysers/Analyser_Merge.py
PanierAvide: Well, I am not a Python expert, maybe the encoding error thrown could be easily avoided. But using it as is wasn't working. And the input file seemed to be in UTF-8, which was making this error even more weird.

frodrigo: Python with UTF-8 IS weird ! Ok I will look at that.

PanierAvide: Thank you :-)

frodrigo: Which one contains "ô" ?

PanierAvide: Nantes file : http://data.nantes.fr/api/publication/24440040400129_NM_NM_00170/Toilettes_publiques_nm_STBL/content/?format=json

frodrigo:
```diff
diff --git a/analysers/Analyser_Merge.py b/analysers/Analyser_Merge.py
index eba6ce5..3564f0e 100644
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -275,9 +275,6 @@ WHERE
     official.tags1 - osm_item.tags - 'source'::text != ''::hstore
 """
 
-def unaccent(s):
-   return ''.join(c for c in unicodedata.normalize('NFD', s) if unicodedata.category(c) != 'Mn')
-
 class Source:
     def __init__(self, attribution = None, millesime = None, url = None, name = None, encoding = "utf-8", file 
= None, fileUrl = None, fileUrlCache = 30, zip = None, filter = None):
         """
@@ -415,16 +412,15 @@ class JSON(Parser):
 
     def header(self):
         self.json = self.extractor(json.loads(self.source.open().read()))
-        columns = map(lambda c: unaccent(c), self.json[0].keys())
+        columns = self.json[0].keys()
         return columns
 
     def import_(self, table, srid, osmosis):
-        self.json = self.json or self.extractor(json.loads(self.source.open().read))
-        insert_statement = u"insert into %s (%%s) values %%s" % table
+        self.json = self.json or self.extractor(json.loads(self.source.open().read()))
         for row in self.json:
-            columns = map(lambda c: unaccent(c), row.keys())
-            values = map(lambda column: unicode(json.dumps(row[column])) if row[column] != None else None, row.keys())
-            osmosis.giscurs.execute(insert_statement, (psycopg2.extensions.AsIs(u",".join(map(lambda c: "\"%s\"" % c, columns))), tuple(values)))
+            osmosis.giscurs.execute(u"insert into \"%s\" (\"%s\") values (%s)" %
+                (table, u'", "'.join(row.keys()), (u'%s, ' * len(row.keys()))[:-2]),
+                map(json.dumps, row.values()))
 
 class GeoJSON(Parser):
     def __init__(self, source, extractor = lambda geojson: geojson):
```

PanierAvide: Thanks for the solution, I will commit this in the same time as flatten fix.

PanierAvide:
This should be OK now, using flattenjson in order to have multi-level properties handled simply, plus removed unaccent method and using your fix.

frodrigo:
Ok. Thank you. One last thing, since you flatten the JSON, It still makes sense to have a specific geojson parser?

PanierAvide:
I think so, because GeoJSON coordinates are in an array, so you will still need to do a bit of parsing for retrieving geometry, whereas here they are mapped into two distinct columns, so analysers just have to name these columns. Also, as the format is standardized, properties are already mapped into columns without the need to prefix every attribute with `properties.`. Less text = clearer code :-)

analysers/Analyser_Merge.py
```diff
@@ -33,7 +33,9 @@
 import zipfile
 import tempfile
 import json
+import geojson
 import re
+import unicodedata
```
frodrigo: Please remove this line.

analysers/Analyser_Merge.py
```diff
@@ -395,28 +397,71 @@ def import_(self, table, srid, osmosis):
     def close(self):
         self.f.close()
 
+def flattenjson(b):
```
frodrigo: Can you add a comment with the source of this code, please.

analysers/Analyser_Merge.py
```diff
@@ -395,28 +397,71 @@ def import_(self, table, srid, osmosis):
...
         self.extractor = extractor
 
         self.json = None
 
     def header(self):
-        self.json = self.extractor(json.loads(self.source.open().read()))
-        return self.json[0].keys()
+        self.json = map(flattenjson, self.extractor(json.loads(self.source.open().read())))
+        columns = self.json[0].keys()
+        return columns
```
frodrigo: Please, keep the original return line, since it's the same.

analysers/Analyser_Merge.py
```diff
@@ -395,28 +397,71 @@ def import_(self, table, srid, osmosis):
...
+        insert_statement = u"insert into %s (%%s) values %%s" % table
+        for row in self.geojson.features:
+            row.properties = flattenjson(row.properties)
+            columns = row.properties.keys()
+            values = map(lambda column: row.properties[column] if row.properties[column] != None else None, columns)
+            columns.append(u"geom_x")
+            columns.append(u"geom_y")
+            values.append(row.geometry.coordinates[0])
+            values.append(row.geometry.coordinates[1])
             osmosis.giscurs.execute(insert_statement, (psycopg2.extensions.AsIs(u",".join(map(lambda c: "\"%s\"" % c, columns))), tuple(values)))
```
frodrigo: Please, use the same as in JSON Parse, is more robust.

analysers/analyser_merge_public_equipment_FR_montpellier_toilets.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8180", "class": 6, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Montpellier toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "http://data.montpellier3m.fr/dataset/toilettes-publiques-de-montpellier",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Montpellier Mediterranée Métropole", millesime = "12/2017",
+                    fileUrl = "http://data.montpellier3m.fr/sites/default/files/ressources/MMM_MTP_WC_PUBLICS.json", encoding = "ISO-8859-15"),
+                extractor = lambda json: json['features']),
+            Load("geometry.x", "geometry.y",
+                where = lambda res: res['attributes.enservice'] == u'"En Service"'),
```
frodrigo: Use "select" in place of "where".
select: {u'attributes.enservice': u'"En Service"'}

analysers/analyser_merge_public_equipment_FR_nantes_toilets.py
```diff
@@ -0,0 +1,52 @@
...
+        self.missing_official = {"item":"8180", "class": 5, "level": 3, "tag": ["merge", "public equipment"], "desc": T_(u"Nantes toilets not integrated") }
+        Analyser_Merge.__init__(self, config, logger,
+            "http://data.nantes.fr/donnees/detail/toilettes-publiques-de-nantes-metropole/",
+            u"Toilettes publiques",
+            JSON(Source(attribution = u"Nantes Métropole", millesime = "11/2016",
+                    fileUrl = "http://data.nantes.fr/api/publication/24440040400129_NM_NM_00170/Toilettes_publiques_nm_STBL/content/?format=json"),
+                extractor = lambda json: json['data']),
+            Load("_l", "_l",
+                xFunction = lambda c: c and json.loads(c)[1],
+                yFunction = lambda c: c and json.loads(c)[0]),
```
frodrigo: Still json.loads here ?

frodrigo:
I think we are close. Thank you for your work.

analysers/analyser_merge_public_equipment_FR_nantes_toilets.py
PanierAvide: Yes because it's an array, better than manual text parsing.

PanierAvide:
Thanks for your comments. As I will not be available, I will edit this at start of january.

PanierAvide:
I committed the requested changes, also corrected values mappings in JSON analysers, as they are imported in table with `"` character.

vinber:
Hello,
I suppose we could add this dataset on Bordeaux: http://opendata.bordeaux.fr/content/toilettes-publiques ?

PanierAvide:
There is already an analyser for Bordeaux : http://osmose.openstreetmap.fr/fr/errors/?item=8180&class=1

vinber:
oups, sorry :( Happy new year !!!

PanierAvide:
No problem ;-) Happy new year to you too 👍

frodrigo:
Thank you. Please can you rebase on master mainly to fetch this commit 137f7b230644956cd6680a2493b4e6e57188f447 and apply it to your new code.

PanierAvide:
This is done ;-)

frodrigo:
Thank you alot for your contribution and patience along the process.

PanierAvide:
Thanks for your patience also and all the code review you have done :-)

frodrigo:
Remove geojson deps as is not available as Debian package on last stable.

frodrigo:
The results are here:
http://osmose.openstreetmap.fr/fr/errors/?item=8180

Unfortunately the JSON change breaks other analyses:
http://buildbot.osmose.openstreetmap.fr/builders/france_local_db/builds/340/steps/osmose_run.py/logs/stdio

PanierAvide:
Are you working on this, or would you like that I fix the broken analysers ?

frodrigo:
Please can you try to fix it ?

PanierAvide:
No problem, I will work on this ASAP.

PanierAvide:
The bug was fixed in https://github.com/PanierAvide/osmose-backend/commit/1471e0687f04f900b3ce6e738878a69b4c0d5476

jocelynj:
Thanks - can you create a pull-request against master ?

PR #271, opened 2018-02-03
handling of external MapCSS translations

stephankn:
* include installation instructions in placeholder
* add external content to .gitignore


jocelynj:
That's a good idea, even if it duplicates some information from the main README.

Can you also rename the file from placeholder to README ?

stephankn:
Yes, having it named README makes it even clearer what this directory is about

PR #289, opened 2018-04-17
docker: upgrade after installing packages

stephankn:
As even a just updated base image can contain upgradeable packages do force a package upgrade to have latest security fixes in the image.


frodrigo:
Ok. But should be done before `apt-get install`, doesn't it ?

stephankn:
Not sure whether it would make a difference. Maybe swapping it to before in case due to updated versions install dependencies change. I'll push a change.

PR #343, opened 2018-10-09
Docker: install new python dependencies

stephankn:


frodrigo:
Right. In fact it's the plan. Please can you remove all the pyhton dependencies from apt and use only pip.

stephankn:
only installing python-pip had resulted later in pip complaining about not finding setuptools and wheel, so breaking the pip install...

frodrigo:
I talk about removing
```
                                python-polib \
                                python-poster \
                                python-psycopg2 \
                                python-shapely \
                                python-regex \
                                python-geojson \
                                python-requests \
```
from apt install, but install them using only pip.

stephankn:
Ah, now I understand. Yes, sounds like a good idea. At least geojson is missing in the requirements.txt. I will clean it up and push a new PR for that if this is OK for you, to not add too many things in a single change. the requirements.txt is not related to docker.

So this one would make docker work with the new requirements. A future one will clean up the requirements.txt.

stephankn:
I figured out geojson was never imported. So it did boil down to a docker cleanup. For imposm.parser and regex some dependencies had been needed in Ubuntu.

frodrigo:
Well done. Thank!

PR #345, opened 2018-10-09
package build-essential is required for pip later

stephankn:


jocelynj:
Just for information: pip and build-essential are not strictly necessary if you use only packages, like the one I generated for python-lark-parser.

stephankn:
@jocelynj that section of the README is imho specific to installing dependencies using pip. Do you think the readme should be restructured to make a clear distinction between installing packages or installing from pip?

jocelynj:
On Tue, Oct 09, 2018 at 10:41:59PM -0700, Stephan wrote:
> @jocelynj that section of the README is imho specific to installing dependencies using pip. Do you think the readme should be restructured to make a clear distinction between installing packages or installing from pip?

You're right - this section is about installing with pip.

I'm not sure how to restructure the README, but it might be clearer if we had
two sub-sections, one for installing through debian packages, the other with a
virtualenv. And we have the issue that debian packages no longer contain all
dependencies, so a note should be added that lark-parser has to be installed
manually.

PR #351, opened 2018-10-13
Merge Police - update with last Gendarmerie file

nlehuby:
* new file url
* use a conflation key
* update a few tags in prevision of https://wiki.openstreetmap.org/wiki/FR:Project_of_the_month/Gendarmerie_nationale

### DIFF
--- a/analysers/analyser_merge_police_FR.py
+++ b/analysers/analyser_merge_police_FR.py
@@ -25,23 +25,32 @@ from .Analyser_Merge import Analyser_Merge, Source, CSV, Load, Mapping, Select, 
 
 class Analyser_Merge_Police_FR(Analyser_Merge):
     def __init__(self, config, logger = None):
-        self.missing_official = {"item":"8190", "class": 1, "level": 3, "tag": ["merge"], "desc": T_(u"Police not integrated") }
+        self.missing_official = {"item":"8190", "class": 1, "level": 3, "tag": ["merge"], "desc": T_(u"Police Gendarmerie not integrated") }
+        self.possible_merge   = {"item":"8191", "class": 3, "level": 3, "tag": ["merge"], "desc": T_(u"Police Gendarmerie, integration suggestion") }
+        self.update_official  = {"item":"8192", "class": 4, "level": 3, "tag": ["merge"], "desc": T_(u"Police Gendarmerie update") }
+
         Analyser_Merge.__init__(self, config, logger,
-            "http://www.data.gouv.fr/fr/dataset/liste-des-points-d-accueil-de-la-gendarmerie-nationale-avec-geolocalisation",
-            u"Liste des points d'accueil de la gendarmerie nationale avec géolocalisation",
-            CSV(Source(attribution = u"data.gouv.fr:Ministère de l'Intérieur", millesime = "02/2016",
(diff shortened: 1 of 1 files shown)


analysers/analyser_merge_police_FR.py
```diff
@@ -26,22 +26,31 @@
...
+            CSV(Source(attribution = u"data.gouv.fr:Ministère de l'Intérieur", millesime = "10/2018",
+                    fileUrl = "https://www.data.gouv.fr/fr/datasets/r/d6a43ef2-d302-4456-90e9-ff2c47cac562"),
                 separator = ";"),
             Load("geocodage_x_GPS", "geocodage_y_GPS"),
             Mapping(
                 select = Select(
                     types = ["nodes", "ways"],
                     tags = {"amenity": "police"}),
-                conflationDistance = 1000,
+                conflationDistance = 500,
```
frodrigo: It's required to lower the distance ?
There is existing Gendarmerie closer than 1000m ?

analysers/analyser_merge_police_FR.py
```diff
@@ -26,22 +26,31 @@
...
+                osmRef = "ref:data.gouv.fr",
                 generate = Generate(
                     static1 = {
                         "amenity": "police",
-                        "operator": "Gendarmerie Nationale"},
+                        "name": "Gendarmerie nationale",
+                        "operator": "Gendarmerie nationale"},
                     static2 = {"source": self.source},
-                    mapping2 = {"phone": "telephone"},
+                    mapping1 = {"ref:data.gouv.fr": "identifiant_public_unite"},
```
frodrigo: It's a stable ref ?

nlehuby: We can hope so, its purpose is to ease the update for the data user: https://twitter.com/ericfreyss/status/1045750462690070530

frodrigo: Ok. But not better ref name? this one is not specific for this data source.

analysers/analyser_merge_police_FR.py
nlehuby: the geoloc does not look that bad, and I would like to reduce the risk to conflate with police office that are not Gendarmerie.
Do you prefer I change it back to 1 km ?

analysers/analyser_merge_police_FR.py
nlehuby: What do you suggest ? `ref:FR:GendarmerieNationale` ?

frodrigo: yes

nlehuby: ok

frodrigo:
http://osmose.openstreetmap.fr/fr/errors/?item=8190,7250,8191,8192&useDevItem=true

frodrigo:
I change the wording for homogeneity with other.
But there is some kind of bug, item=8190 have no content.

frodrigo:
I finally found the 8190 bug, it was on front of the keyboard.
The good URL is:
http://osmose.openstreetmap.fr/fr/errors/?item=8190,7250,8191,8192&useDevItem=all
http://osmose.openstreetmap.fr/fr/errors/graph.png?source=6388&item=8190&class=1

frodrigo:
In fact we should not set the missing_osm item=7250 because it reports on all amenity=police the missing ref:FR:GendarmerieNationale. But in France amenity=police can be "Gendarmerie" or "Police".

PR #375, opened 2018-10-28
Handling hotel chains in wikidata brand tag fix

PanierAvide:
Hi,
Just changed a bit the URL in order to also find hotel chains for this tag fix.


frodrigo:
Thank

PR #385, opened 2018-11-07
docker: create missing locale en_US.UTF-8

### LINKED ISSUE #384: creation of database fails due to invalid locale
creating database cluster inside docker fails due to invalid locale.

version: 3491240

```
# pg_createcluster 10 main
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
Error: The locale requested by the environment is invalid:
  LANG: en_US.UTF-8
```

locale is missing inside the container.

stephankn:
fixes #384

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,8 @@ RUN apt-get update && \
 
 RUN mkdir -p /data/work/osmose && \
     useradd -s /bin/bash -d /data/work/osmose osmose && \
-    chown osmose /data/work/osmose
+    chown osmose /data/work/osmose && \
+    locale-gen en_US.UTF-8
 
 ADD requirements.txt /opt/osmose-backend/requirements.txt
 WORKDIR /opt/osmose-backend


frodrigo:
In order to better cache usage can you add this to the first RUN please.

stephankn:
What are you trying to optimize? If you create another docker image with slightly different locale you would break caching by adding it to the package RUN statement. To reduce the number of layers the two RUN lines could be merged.

I do not see a strong benefit for merging or moving around the locale generation. From a system perspective creating a locale is more related to creating users and setting the ENV than to installing packages.

Let me know if you really insist of having it moved, then I push a modified patch.

frodrigo:
Sorry I misread the Dokerfile. You are right, it will change nothing.

Issue #402, opened 2018-11-22
No updates are reported as errors (code 4)


stephankn:
My backend instance often fails due to OSM extracts not being updated and so no updated data delivered to the frontend.
This makes it failing with error code 4 in case a single analyzer could not send fresh data.

https://github.com/osm-fr/osmose-backend/blob/master/osmose_run.py#L292

As I monitor my instance with Zabbix I get an alert for every country/state where this fails. Often these sum up to over hundred alerts.
These seriously clutter the monitoring system and make it hard to distinguish these from real problems.

I suggest to turn this into a warning and not returning an error for it.

What needs to be discussed is how a situation could be detected where extracts gone stale and do not update at all for a longer time.
Using the error code 4 for this purpose sounds wrong, so the monitoring for stale data updates should be done separate of this.
Would it be possible to provide the "delay" of a single country by API? then I could monitor that and again warn if it exceeds a threshold.

I am going to push some changes to frontend/backend to implement the warning handling. Please comment on it.

frodrigo:
We now have this in the frontend API:
https://github.com/osm-fr/osmose-frontend/blob/master/control.py#L294
The idea in the backend is to get the timestamp before running the analysis, and if there will be something to update, then run the analysis. But this is still not done.

stephankn:
This sounds then like a suitable interface to check for stalled processing.

How about changing the current FAIL into a WARNING as proposed in my PR above? So osmose_run will no longer report an error is such situations?

frodrigo:
3fdac9144d4136bbeee4b8c1d36344f21f9cda46
ab2bda1ae5b058f2689d24e44ba64ae07cb7ef5e
Does not report error if "FAIL: Already up to date" happen after a timeout.

frodrigo:
Now it's OK.

PR #403, opened 2018-11-22
In case of a warning do not return error code

stephankn:
see #402


stephankn:
any chance of getting this merged and frontend adjusted? Having analyzers "fail" because of some expected condition is worth a warning, but not an error. It makes it difficult to monitor that everything works well.

frodrigo:
We still not look at this. We was focused on other issues. But now our infrastructure is more stable.

frodrigo:
From my point of view, we must prevent `FAIL: Already up to date` to append.
We must not run the analyser.

jocelynj:
I think that we have a different objective:
  - we want to know when backend or frontend is failing, meaning that this error is useful for us.
  - you only want to know when a failure is linked to your backend, or data from your countries.

As a compromise, what do you think of using a specific error code for update error, and you can filter errors that have a different error code ?

stephankn:
@jocelynj :
I would like to avoid treating some "specific" errors as non-errors.
If this would be the route to go, then I would suggest to reserve some of the least significant bits to describe warnings. Then treating everything larger than e.g. 32 would indicate an error.

While I understand the desire to ge feedback of a failing backend, I wonder how having this reported as an error is useful to you. Is the error code also meaningful to you, I wonder how you get it. I do not see code which uploads the error code at the end of the analysis.

This might also be an option: Collect all errors in a verbose structure and upload that at the end. But right now I did not see such code.

So the only one noticing such result codes is the person running the backend code.
I monitor the the backends with zabbix. And for each failing analyzer I get an error reported and check the logs what was broken. These "Already up to date" results are currently a false alert.

@frodrigo : can you please detail a bit further about what exactly triggers this FAIL? I see that it is sent when the update throws an exception here:
https://github.com/osm-fr/osmose-frontend/blob/master/control.py#L292

What **exactly** is checked here? Is it really that case that something did not change? Or is it a problem with some processes on the frontend? 
https://github.com/osm-fr/osmose-frontend/blob/3a9f71307e339320983271978b3e7cc825fa241b/tools/update.py#L405

In case this FAIL indicates actually a problem on the front-end, should I then report such cases? I also not yet understand what you mean than the analyzer is not supposed to run beforehand. What exactly is meant to be the trigger here? Or does it all boil down to the OSM files not been updated? But why is then only a single analyzer failing and not all of that country?

jocelynj:
We are using buildbot to get all information, including full log and error code. See here for an example:
https://buildbot.osmose.openstreetmap.fr/builders/central_african_republic/builds/393/steps/osmose_run.py/logs/stdio

There is also a filter in buildbot to mark some error code as "error" or "warning"


About the fail "Already up to date" on update, they are triggered when frontend is receiving results for a country+analyser with the same timestamp than the previous result. It happens in at least two scenarios:
  - an analyser sending several results for different checks with the same analyser name. This error was fixed recently, by using python class name instead of python file name to generate the analyser name sent to frontend. This error should no longer happen.
  - results was sent twice to frontend, because the previous upload failed. We don't know yet why frontend is sending an error, but still importing the results in its database. This case has to be fixed.
  - there might be other cases that we need to know about.

As a compromise, it might be possible to modify backend to not send an error if previous update got a "504: Gateway Timeout", as results was still uploaded to frontend.

frodrigo:
3fdac9144d4136bbeee4b8c1d36344f21f9cda46
ab2bda1ae5b058f2689d24e44ba64ae07cb7ef5e
Does not report error if "FAIL: Already up to date" happen after a timeout.

PR #405, opened 2018-11-22
have git describe add -dirty in case of uncommitted local changes

stephankn:


frodrigo:
```
(venv3)fred@osm110:/home/fred/frodrigo-osmose-backend/analysers$ git describe
v1.0-2018-11-20-9-g9e1a027
(venv3)fred@osm110:/home/fred/frodrigo-osmose-backend/analysers$ git describe --dirty
v1.0-2018-11-20-9-g9e1a027-dirty
```

@jocelynj what do you think of this?

jocelynj:
I don't have any objection to this change, and I see how it can be useful to find if anything is not commited.

Issue #404, opened 2018-11-22
calling git describe on non-git directory in Docker


stephankn:
osmose_run.py tries to call `git describe`. This is useless since osmose-backend inside Docker is no git repository. This folder is explicitly excluded from being added (see `.dockerignore`).

https://github.com/osm-fr/osmose-backend/blob/master/osmose_run.py#L68

frodrigo:
The idea is to mount the source code as a volume and got back .git again. Mounting the source avoid to rebuild the image at each update.
BTW, I have a branch changing most on the Docker part : https://github.com/frodrigo/osmose-backend/commits/docker

stephankn:
Uh, sounds complicated. Why not simply passing in the `git describe` value as an ARG to later assign this to ENV inside docker? Sounds much cleaner than importing sources into the image and running git inside the container.
Versions will not change after the image was created. So no need to determine the version multiple times.

jocelynj:
In fact, we might need two versions:
  - the one from origin/master
  - the one from HEAD

This is because HEAD might not be up-to-date with origin/master.

frodrigo:
Now, It's OK for you? Can we close this?

PR #406, opened 2018-11-22
add version from Docker image via environment

stephankn:
improves  #404 ,

call 

    docker build --build-arg GIT_VERSION=`git describe --dirty`

to pass the current version as environment to the image.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -34,6 +34,8 @@ ENV PYTHONPATH /opt/osmose-backend
 
 ADD . /opt/osmose-backend/
 
+ARG GIT_VERSION
+ENV OSMOSE_VERSION ${GIT_VERSION:-(unknown)}
 ENV LANG en_US.UTF-8
 ENTRYPOINT ["/opt/osmose-backend/tools/docker-entrypoint.sh"]
 CMD bash
--- a/osmose_run.py
+++ b/osmose_run.py
@@ -65,12 +65,14 @@ class analyser_config:
   pass
 
 def get_version():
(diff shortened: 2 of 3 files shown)


osmose_run.py
```diff
@@ -65,13 +65,13 @@ class analyser_config:
   pass
 
 def get_version():
-    cmd  = ["git", "describe"]
+    cmd  = ["git", "describe", "--dirty"]
```
frodrigo: Remove this change from this PR, please.

osmose_run.py
```diff
@@ -65,13 +65,13 @@ class analyser_config:
...
     try:
         proc = subprocess.Popen(cmd, stdout=subprocess.PIPE)
         version = proc.stdout.readlines()[0].strip()
     except:
         version = "(unknown)"
-    return version
+    return os.getenv('OSMOSE_VERSION', version)
```
frodrigo: If there is `OSMOSE_VERSION` do not call to git before.

docker/Dockerfile
```diff
@@ -34,6 +34,8 @@ ENV PYTHONPATH /opt/osmose-backend
 
 ADD . /opt/osmose-backend/
 
+ARG GIT_VERSION
```
frodrigo: Can we make the `git describe` in Dockerfile?

stephankn: This is not possible, as the .git file is already excluded inside the Docker build environment

stephankn:
do you need further changes to have this merged?

frodrigo:
It's OK. Thank you.

PR #409, opened 2018-11-24
Improve Paris bicycle parking OpenData integration

Phyks:
Hi,

This is an improvement on the latest bicycle parking in Paris OpenData analysis, based on first feedbacks I have:

* Dataset is actually quite precise for the locations, 20m should be
enough as conflation distance.
* Shared motorcycle / bicycle parkings should be matched against both
motorcycle_parking and bicycle_parking.

Thanks!


analysers/analyser_merge_bicycle_parking_FR_paris.py
```diff
@@ -122,8 +122,9 @@ def __init__(self, config, logger = None):
             Mapping(
                 select = Select(
                     types = ["nodes"],
-                    tags = {"amenity": "motorcycle_parking"}),
-                conflationDistance = 50,
+                    tags = [{"amenity": "motorcycle_parking"},
+                            {"amenity": "bicycle_parking"}])
```
frodrigo: `tags = {"amenity": ["motorcycle_parking", "bicycle_parking"]}`
Use this, it is more efficient.

Phyks: Oh, I missed that this syntax was actually valid. This should be fixed :)

PR #434, opened 2019-01-11
Fix analyser_merge_street_number wording for Rennes

PanierAvide:
Small fix of wording for street numbers in Rennes, which is quite broken for now.

capture d ecran_2019-01-11_08-56-00


frodrigo:
Thank you.

PR #442, opened 2019-02-06
clarify minimum versions of docker/docker-compose

stephankn:
at least compose 1.19.0 is needed due to this bug:

Fixed a bug where tmpfs volumes declared using the extended syntax in
Compose files using version 3.2 would be erroneously created as anonymous
volumes instead


frodrigo:
Thank.

Issue #433, opened 2019-01-09
opening hours parser is incomplete and produces lots of false positives


simonpoole:
I would suggest disabling the check globally as it is far from correct and highlighting lots of issues with perfectly good OH strings.

Example: it complains about opening_hours = Mo-Sa 07:00-19:00; PH Su off 

Which is perfectly OK, see https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:weekday_selector

Note: it is likely that the mapper in question actually should have used PH,Su but that is not the point. 

Other example: the relatively commonly used weekday in month syntax is completely missing from the BNF for the parser in OSMOSE.

See https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:weekday_range:nth_entry

frodrigo:
@rezemika are you able to fix your parser?

rezemika:
Yes, it's done. Check the `0.1.8` version of oh_sanitizer. Sorry for the inconvenience.

rezemika:
I'll check this as soon as possible, sorry for the delay (I am in exam period).

frodrigo:
Thank you. For now I take the 0.1.9.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/graph.png?item=3250&class=32501&start_date=2019
Look good. It removes 10% of issues.

dex2000:
Other example:

    We-Sa 18:00+

is marked as error.
Like: http://osmose.openstreetmap.fr/de/error/22862069211

rezemika:
I am currently working on a project which will be able to parse and correct almost all fields. However, it will take a long to complete. So, in the mean time, it may indeed be relevant to disable the OH checking.
Sorry for the inconvenience.

frodrigo:
@rezemika a completely different project? Should I remove OH checking from Osmose or just disable it?

simonpoole:
@rezemika why? There are a number of fairly to very complete parsers including my javacc one (that in principle could generate c++ code too).

rezemika:
Not really, it would be a next major version of `humanized_opening_hours` (but with a quite good retro-compatibility), which will be able to parse, sanitize and understand almost all valid fields. However, it's probably a matter of weeks before a stable release (my free time is so rare), so it may be relevant for you to find an alternative...

frodrigo:
OK I disable it. For Osmose there is no need of retro-compatibility.

[...]

PR #444, opened 2019-02-11
Improve the oh_sanitizer module

fgouget:
This patchset fixes the unit tests, existing functionality, and provides some extensions so Osmose can more often suggest fixes.
Importantly it adds support for time spans with no end time or an open-ended end time (+) as per the opening_hours specification. This should reduce the number of errors (false positives actually) quite significantly (I'd guess at least 30%).


frodrigo:
Oh. Thank to work on this subject.
@rezemika is dooing similar work in parallel
#433

frodrigo:
Thank you

PR #445, opened 2019-02-11
Website: A plugin to check the URLs.

fgouget:
This verifies that the URLs contain a valid scheme and no invalid
character (just spaces for now as they are the most likely offenders).

Note that this new plugin is untested!

### DIFF
--- a/plugins/Website.py
+++ b/plugins/Website.py
@@ -1,0 +1,110 @@
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Francois Gouget fgouget free.fr 2019                       ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from plugins.Plugin import Plugin
+
+
+class Website(Plugin):
+
+    URL_TAGS = set((u"contact:website", u"contact:webcam", u"url",
+                    u"website", u"website:mobile", u"website:stock"))
+
+    def init(self, logger):
+        Plugin.init(self, logger)
+
+        import re
+        # From RFC 1738 paragraph 2.1
+        self.HasScheme = re.compile(r"^[a-zA-Z0-9.+-]+://")
+
+        self.errors[30931] = {"item": 3093, "level": 2,
+                              "tag": ["value", "fix:chair"],
+                              "desc": T_(u"The URL contains a space")}
+        self.errors[30932] = {"item": 3093, "level": 2,
+                              "tag": ["value", "fix:chair"],
+                              "desc": T_(u"The URL does not have a valid scheme")}
+
+    def _bad_url(self, tag, tags):
+        return T_("Bad URL %(k)s=%(v)s", {"k": tag, "v": tags[tag]})
+
+    def check(self, tags):
+        for tag in self.URL_TAGS:
+            if tag not in tags:
+                continue
+            url = tags[tag]
+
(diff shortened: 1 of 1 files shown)


matthewdarwin:
Thanks for this!   I hope this is just a start and you will extend it to check hostname is valid (do a DNS lookup for either A or AAAA record), the web page exists and if the URL is HTTP and HTTPS also works then to suggest changing the URL to HTTPs!

fgouget:
I am not sure those checks would be appropriate for Osmose.
* DNS lookups and (even partial) page downloads are very very slow, because of potential timeouts. In particular trying to access a non-existant https site will likely bock for 30 seconds or more. Given the number of URLs Osmose has to check doing those checks would take days.
* Even if Osmose is threaded, those checks are not CPU-bound. So it would likely require a different approach to threading to start more than one thread per core. How many is best is anybody's guess.
* That could greatly increase the network usage.
* And finally, having this type of URL validity check on a much smaller scale, this type of check is a major pain. There's all these URLs that work just fine but where the web server blocks you because it does not like your user-agent. Then those that are more subtle and redirect you to a different page still because of your user-agent. It does not help matters that redirects come in multiple forms (HTTP 302, http-equiv tag buried after kilobytes of garbage, or even javascript!) and that you should follow them to detect loops (often caused by bad URLs). I could go on and on...

Finally it's not because an OpenStreetMap object contains a syntactically valid URL and has not been modified for years that the URL it points to is still valid. So where Osmose rechecks objects only after modifications, verifying a URL is still valid should be done periodically on every URL in the database.
The same is true for switching to https. A website that's only accessible through http today may become accessible through https at any time in the future.

So to me all this means deep URL validation is a job for a different service (and one I'm unlikely to attempt).

frodrigo:
Thank @fgouget ! We will review it.
I'm agree with @fgouget, checking for remote web URL is not appropriate for standard analyse. It may be done, but with something dedicated.

Note: we have a mode for checking diff only, but not enabled, yet.

fgouget:
I updated the patch to take into account the previous remarks:
* The plugin now loops over URL_TAGS and for each of them checks if it is set.

* In cases where we don't know how to fix the issue I now set the text field in the error.

* I did not go the subclass route. Instead I went the Phone.py way and defined two error types: 30931 if the URL contains spaces, and 30932 if it does not have a valid scheme.

frodrigo:
Thank you.

frodrigo:
Please, next time, avoid the return at the start of the test ;-)
https://github.com/frodrigo/osmose-backend/commit/33e9dff15a186f89af9b2d7b38a8617e2b241952#diff-4a1e0409090fcc2436d15efa86ced998L92

frodrigo:
http://osmose.openstreetmap.fr/fr/errors/?item=3093&useDevItem=true

fgouget:
I only noticed two potential issues.

The first one is the more important one and happens a lot in Germany. The scheme check seems intent on turning things like "http://www.goethegymnasium-demmin.de/" into "http://http://www.goethegymnasium-demmin.de/". That seems to represent the bulk of the issues there and yet does not seem to happen in other countries like France for instance. I also cannot reproduce this issue here.


The other potential issue has to do with the space check because some people have put a list of URLs in the website tag. However:
* I saw no mention of this being allowed in the docs.
* The way it's done is very random: "url1; url2" "url1 + url2" "url1 e url2" (where 'e' is for 'and' in Italian), etc.
* Semi-colons are sometimes used as an alternative parameter separator for http and https URLs. So using semi-colons to separate URLs would make those unparseable.
* I only found one company in Alsace which may legitimately have multiple URLs because it has one website per domain of activity each with different content.

frodrigo:
Le 21/02/2019 à 19:59, fgouget a écrit :
>
> I only noticed two potential issues.
>
> The first one is the more important one and happens a lot in Germany. 
> The scheme check seems intent on turning things like 
> "http://www.goethegymnasium-demmin.de/" into 
> "http://http://www.goethegymnasium-demmin.de/". That seems to 
> represent the bulk of the issues there and yet does not seem to happen 
> in other countries like France for instance. I also cannot reproduce 
> this issue here.
>
I already fixed it. It's remaining data.


> The other potential issue has to do with the space check because some 
> people have put a list of URLs in the website tag. However:
>
>   * I saw no mention of this being allowed in the docs.
>   * The way it's done is very random: "url1; url2" "url1 + url2" "url1
>     e url2" (where 'e' is for 'and' in Italian), etc.
>   * Semi-colons are sometimes used as an alternative parameter
>     separator for http and https URLs. So using semi-colons to
>     separate URLs would make those unparseable.
>   * I only found one company in Alsace which may legitimately have
>     multiple URLs because it has one website per domain of activity
>     each with different content.
>
May we disable fix suggestion when we are in such suspicious case?

frodrigo:
http://osmose.openstreetmap.fr/fr/errors/graph.png?item=3093&start_date=2019-02

PR #459, opened 2019-02-24
Python 3 fixes for the opening hours validator

fgouget:
oh_sanitizer: Fix some test failures that happened only in Python 3

### DIFF
--- a/modules/oh_sanitizer/main.py
+++ b/modules/oh_sanitizer/main.py
@@ -80,9 +80,11 @@ class SanitizerTransformer(_lark.Transformer):
         return "24/7"
     
     def selector_sequence(self, args):
-        if len(args) == 1:  # "time_selector"
+        if len(args) == 1:  # time_selector
             return args[0]
-        else:  # "range_selectors time_selector"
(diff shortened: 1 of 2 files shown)


modules/oh_sanitizer/main.py
```diff
@@ -269,7 +271,7 @@ def time(self, args):
     
     def hms(self, args):
         combined = int(args[0])
-        h = combined / 100
+        h = int(combined / 100)
```
frodrigo: Just use integer div , replace "/" by "//"

fgouget:
```suggestion
        h = combined // 100
```

fgouget:
I used Github's suggestion option to hopefully modify the commit to use an integer division.
This means losing compatibility with some old Python 2 versions but I guess that's ok.

frodrigo:
The min pyhton 2 version of Osmose is 2.7, And it's work.

The plan is to do changes in phton2.7 compatible python3.

frodrigo:
Thank

PR #460, opened 2019-02-24
plugins/Website: Continue fixing tags after stripping spaces from a URL

fgouget:
Signed-off-by: Francois Gouget <fgouget@free.fr>


fgouget:
Unless maybe there was a reason not do do so in this case?

frodrigo:
You are right.

PR #461, opened 2019-02-24
Phone

fgouget:
Some minor phone number validation improvements.

### DIFF
--- a/plugins/Phone_fr.py
+++ b/plugins/Phone_fr.py
@@ -41,6 +41,8 @@ class Phone_fr(Plugin):
         # Regular numbers must not have a 0 after +[code]
         self.BadInter = re.compile(r"^([+]%s *0)([0-9 ]{%s,})$" % (self.code, self.size))
 
+        self.DupSpaces = re.compile(r" {2,}")
+
         # National numbers to internationalize. Note that in addition to
         # short numbers this also skips special numbers starting with 08
         # or 09 since these may or may not be callable from abroad.
@@ -56,6 +58,9 @@ class Phone_fr(Plugin):
                               "tag": ["value", "fix:chair"],
                               "desc": T_(u"Missing international prefix")}
 
+    def fix(self, tag, str):
+        return {tag: self.DupSpaces.sub(' ', str)}
+
     def check(self, tags):
         err = []
         for tag in self.PHONE_TAGS:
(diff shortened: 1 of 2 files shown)


frodrigo:
Hum. I made a concurrent change : https://github.com/frodrigo/osmose-backend/commit/2cd10504669fd52dc7b200127f8897b135ecc1e4

plugins/Phone_fr.py
```diff
@@ -117,6 +118,7 @@ class father:
                             (u"+33  3631", u"3631"),
                             (u"0102030405", u"+33 102030405"),
                             (u"01 02 03 04 05", u"+33 1 02 03 04 05"),
+                            (u"01.02.03.04.05", u"+33 1 02 03 04 05"),
```
frodrigo: Please, keep input format. It's not really a fault, data still usable. It will raise lot of not interesting issue.

frodrigo:
Notice : https://lists.openstreetmap.org/pipermail/talk-nc/2019-February/thread.html

plugins/Phone_fr.py
fgouget: The OpenStreetMap wiki says the phone fields should follow ITU-T E.123 which does not allow dots in the phone numbers. (*)

Also some Android phones don't work if there are dots in the phone number (I had the case with a Chinese vanilla Android 5 device). And on my device (Samsung Android 8), the Samsung GUI removes all dots as soon as you start editing a phone number. Maybe because the software author is a control freak who does not like dots, or more likely (imho) because they had reports of people putting dots in their phone numbers and running into trouble with one device or another (the contact list is shared across devices after all).

So I'd say dots should really go.


(*) The other references (ITU-T E.164, DIN 5008 and RFC 3966/NANP) don't say anything about formatting except RFC 3966 which says no spaces for phone URIs.

fgouget:
Wikipedia has two nice resources:
https://en.wikipedia.org/wiki/Telephone_numbers_in_New_Caledonia
https://en.wikipedia.org/wiki/National_conventions_for_writing_telephone_numbers

plugins/Phone_fr.py
frodrigo: I know, but Osmose must be fault tolerant at some point.

By experience, you can't flood contributors with issues asking to change data with low value added.

frodrigo:
The commit a04c9554fb46b27f7adc2bbfed670a70ce55118f plugins/Phone_fr: Check all phone numbers in one go  is now on master.

frodrigo:
Sorry. I need to close it now.

PR #480, opened 2019-04-13
Mapcss Colour - also check colour tag

nlehuby:
As mentioned in https://github.com/Jungle-Bus/transport_mapcss/pull/30 add checks for colour tag


frodrigo:
Thank.

PR #485, opened 2019-04-27
Respect conditional speed limits

zzptichka:
[Sign example](https://www.mapillary.com/app/?focus=photo&pKey=B7HnjvewARqzSzTebVuXdQ&signs=true)
Typically can be seen in school zones.
Corresponding road should be tagged with `maxspeed:conditional=40 @ (Mo-Fr 07:00-09:30,14:00-17:00;PH off;SH off)`
This PR should remove false positive warnings for such signs next to the roads tagged with this tag.

### DIFF
--- a/merge_data/mapillary-traffic-signs.mapping.json
+++ b/merge_data/mapillary-traffic-signs.mapping.json
@@ -26,6 +26,10 @@
       {
         "highway": null,
         "maxspeed:backward": "5{speed_limit_unit}"
+      },
+      {
+        "highway": null,
+        "maxspeed:conditional": null
       }
     ],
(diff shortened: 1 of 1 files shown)


frodrigo:
This issue is not US/CA specific. It can be done for all maxspeed.
It's not yet supported, be we can easily add the support of LIKE or regex match : `"maxspeed:conditional": "40 @ %"`

zzptichka:
For some reason `maxspeed:conditional` was defined only for non-US 80 and 100. I thought it was by design. Added this tag for all signs with `"maxspeed:conditional": null` for now, will change to regex once it's supported.

PR #486, opened 2019-04-28
Analyser merge with P+R parkings in IDF

nlehuby:


analysers/analyser_merge_parking_FR_IDF.py
```diff
@@ -0,0 +1,47 @@
...
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from .Analyser_Merge import Analyser_Merge, Source, CSV, Load, Mapping, Select, Generate
+
+class Analyser_Merge_Parking_FR_IDF_park_ride(Analyser_Merge):
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8130", "class": 751, "level": 3, "tag": ["merge", "parking"], "desc": T_(u"CAPP parking disabled not integrated") }
```
frodrigo: Please, change the "desc".

analysers/analyser_merge_parking_FR_IDF.py
```diff
@@ -0,0 +1,47 @@
...
+                select = Select(
+                    types = ["nodes", "ways"],
+                    tags = {
+                        "amenity": "parking",
+                        "park_ride": None}),
+                    conflationDistance = 300,
+                generate = Generate(
+                    static1 = {"amenity": "parking", "park_ride": "yes"},
+                    static2 = {"source": self.source},
+                    mapping1 = {"capacity": lambda res: int(float(res["NB_PL_PR"])) if res["NB_PL_PR"] != "0" else None} )))
```
frodrigo: just cast to int, not float.

nlehuby: the opendata value represents a float actually, and the `int()` don't know how to handle it

frodrigo: OK

frodrigo:
I tested the analyser and produce 0 result. Do you have already integrated all the P+R?

nlehuby:
nope :thinking: I've tested in IDF_seine_et_marne last week, and I got a few results.
testing again.

nlehuby:
6 results with ```docker-compose -f docker-compose.yml -f docker-compose-dev.yml --project-name comoros run --rm backend ./osmose_run.py --country=france_ile_de_france_seine_et_marne --analyser=merge_parking_FR_IDF```

frodrigo:
OK ok france_ile_de_france_seine_et_marne there is results.

frodrigo:
Thank you.

frodrigo:
http://osmose.openstreetmap.fr/en/map/#item=8130&zoom=9&lat=48.855&lon=2.43&level=1%2C2%2C3&tags=&fixable=

nlehuby:
hum, actually that's not really convincing. Most of the time, there is already a parking, it only lacks the park_ride tag (and maybe the capacity one too)
I'm not sure about the better way to handle that:
* with a `possible_merge` class instead / in addition to the `missing_official` class ?
* by searching for existing parking (without checking the park_ride tag) ?

what do you think ?

frodrigo:
Yes we can add `possible_merge` and maybe `updae_official`.

frodrigo:
But there is so few to integrate. Do you really thing, it's required ?

nlehuby:
I think it would ease the maintenance (to check and add P+R directly from osmose frontend, even in the future when new elems  will be added in the ref dataset). But tell me, if you think it's not valuable enough, i can indeed live without :yum:

frodrigo:
There is down side. The computation required to do it. We need to code it. Then run it every day for ever.

nlehuby:
ok

PR #496, opened 2019-05-01
Analyser for missing gauge tag on railway=rail in France

PanierAvide:
Hello,
As majority of railway=rail in France has the gauge=* tag defined, it would be interesting to have this analyzer to locate rails lacking this information. This tag is particularly interesting when performing routing over rail network. Default value is 1435 when rail is connected to main railway network, but as this analyze is pretty basic, maybe it's not a good idea to offer a direct fix. If we want to go further, we need some network analysis, for example using PgRouting (but I'm not sure if it's possible to use it in current Osmose setup).
Regards,
Adrien.


frodrigo:
Item 2150 is about maxspeed.

Note there is already tow hidden items about railsways. It will be better to improve this mapcss, or create a new one : http://osmose.openstreetmap.fr/fr/errors/?item=9015,9016&useDevItem=all

PanierAvide:
I can't see where the MapCSS file for OpenRailwayMap is located, is it in the same repository ?

frodrigo:
https://josm.openstreetmap.de/wiki/Rules

PanierAvide:
OK, so I created [this ruleset](https://josm.openstreetmap.de/wiki/Rules/FranceSpecificRules), now I have to figure out how to integrate this in Osmose.

PR #497, opened 2019-05-01
MapCSS rules for France (from JOSM)

PanierAvide:
So I used the tool to convert from the https://josm.openstreetmap.de/wiki/Rules/FranceSpecificRules MapCSS file, and used item 9016 for this gauge on railway rule.

### DIFF
--- a/plugins/Josm_fr.py
+++ b/plugins/Josm_fr.py
@@ -1,0 +1,58 @@
+#-*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import modules.mapcss_lib as mapcss
+import regex as re
+
+from plugins.Plugin import Plugin, with_options
+
+class Rules_FranceSpecificRules(Plugin):
+
+
+    def init(self, logger):
+        Plugin.init(self, logger)
+        tags = capture_tags = {}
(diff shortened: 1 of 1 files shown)


frodrigo:
I'm impressed, you build a mapcss files!

Unfortunately most steps was necessary. I update the mapcss source file to add -osmose attributes. Then rename osmose assert attributes, and finally regenerated the python file and changing generator configuration.

PanierAvide:
Great, thanks for having done the last steps, that makes clearer for me what is necessary to have a complete MapCSS analyzer for Osmose :-)

frodrigo:
Result
http://osmose.openstreetmap.fr/en/map/#item=&useDevItem=all&zoom=7&lat=46.019&lon=1.472&level=1%2C2%2C3&tags=&fixable=

Item 2160 should be activated

PanierAvide:
Yes I saw that this morning, that's great :-) Thanks for the fix and integration !

didier2020:
I made a modification to  https://josm.openstreetmap.de/wiki/Rules/FranceSpecificRules :
remove "ref" in nodes'rule highway milestone.
Can you do an osmose update for this mapcss rule ?

frodrigo:
@didier2020 maybe wait a bit more as we are agree?

didier2020:
yes i can...
those who think that ref should not be a tag, do not offer an alternative

PR #509, opened 2019-05-17
OH bug: bad semicolon between weekdays

HubMiner:
Fix bug where semicolon is suggested between days of the week.

add to separator tweaking logic: check if previous key is hour (or sunset, etc).


frodrigo:
Thank.

PR #535, opened 2019-06-16
Remove material from simple value tag

ThomasG77:
Material can contain multiple values, for instance when using key bicycle_parking


frodrigo:
Thank.

PR #536, opened 2019-06-16
docker: clarify doc

bilboed:
It's not obvious that you have to compile the module when you use
docker as a development environment.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -91,6 +91,9 @@ running. You can stop them with `docker-compose stop`.
 Note: if you don't need the frontend just remove `-f
 docker-compose-frontend.yml`.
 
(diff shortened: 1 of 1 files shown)


jocelynj:
I don't use docker files, but isn't this compilation supposed to be done here?
https://github.com/osm-fr/osmose-backend/blob/master/docker/Dockerfile#L32

bilboed:
Yes. When you  build the container. But in "dev" mode you overwrite the directory with the one from the host. And that one doesn't contain the compiled module

frodrigo:
Better solution should be found, but for now, let's doc it.

frodrigo:
Thank

PR #537, opened 2019-06-16
Name_UpperCase: Add whitelist of some names

bilboed:
Some names are meant to be in uppercase. Instead of complaining,
just whitelist them per country.

### DIFF
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -22,6 +22,10 @@
 from plugins.Plugin import Plugin
 import regex as re
 
+# Whitelist of allowed capitals by country code
+UpperCase_WhiteList = {
+    "FR": ["CNFPT", "COSEC", "EHPAD", "MACIF", "MEDEF", "URSSAF"]
+}
 
 class Name_UpperCase(Plugin):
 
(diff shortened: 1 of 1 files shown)


matthewdarwin:
Great! Would it be possible to keep the list somewhere outside the code so it is more easily to be edited by non-programmers?

I would like to add some exceptions for Canada.

bilboed:
It's a rather small list for now which doesn't require a separate file. The format is rather easy for non programmers also (quite similar to json).

plugins/Name_UpperCase.py
```diff
@@ -22,6 +22,10 @@
 from plugins.Plugin import Plugin
 import regex as re
 
+# Whitelist of allowed capitals by country code
+UpperCase_WhiteList = {
+    "FR": ["CNFPT", "COSEC", "EHPAD", "MACIF", "MEDEF", "URSSAF"]
+    }
```
frodrigo: Please, can you align to start of the line.

plugins/Name_UpperCase.py
```diff
@@ -49,14 +63,21 @@ def way(self, data, tags, nds):
...
         a.init(None)
         for t in [{u"name": u"COL TRÈS HAUTTT"},
+                  {u"name": u"EHPAD MAGEUSCULE"},
                   {u"name": u"AÇǱÞΣSSὩΙST"},
                  ]:
             self.check_err(a.node(None, t), t)
             self.check_err(a.way(None, t, None), t)
 
         for t in [{u"name": u"Col des Champs XIIVVVIM"},
+                  {u"name": u"EHPAD Les Petits Vieux"},
```
frodrigo: Please, can you change this name?

frodrigo:
@matthewdarwin let's keep it simple and move to a JSON file latter if required.

plugins/Name_UpperCase.py
bilboed: fixed

plugins/Name_UpperCase.py
bilboed: fixed

frodrigo:
Thank.

PR #546, opened 2019-06-28
Update plugin Phone Spain

JorgeSanzS:
Update the phone plugin to be compatible with Spain.

### DIFF
--- a/plugins/Phone.py
+++ b/plugins/Phone.py
@@ -252,3 +252,29 @@ class Test(TestPluginCommon):
         # Verify we got error for other correct numbers
         for bad in (u"3631", u"(123) 123-4567", "+1 123 1234567"):
             assert p.node(None, {"phone": bad}), ("phone='%s'" % bad)
+    def test_ES(self):
+        p = Phone(None)
+        class _config:
+            options = {"country": "ES", "phone_code": "34", "phone_len": 9, "phone_format": r"^[+]%s[- ][0-9]{3}[- ][0-9]{3}[- ][0-9]{3}$"}
+        class father:
+            config = _config()
+        p.father = father()
+        p.init(None)
+
+        for (bad, good) in (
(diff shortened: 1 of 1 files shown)


jocelynj:
@JorgeSanzS: We just added some automatic tests on osmose-backend, and it seems that your new tests are failing. Do you need help to fix them or run tests?

plugins/Phone.py
```diff
@@ -252,3 +252,29 @@ class father:
         # Verify we got error for other correct numbers
         for bad in (u"3631", u"(123) 123-4567", "+1 123 1234567"):
             assert p.node(None, {"phone": bad}), ("phone='%s'" % bad)
+    def test_ES(self):
```
frodrigo: Plase, skip a line before def.

frodrigo:
Tu run the test, just run
```sh
nosetests plugins/Phone.py
```

plugins/Phone.py
```diff
@@ -252,3 +252,29 @@ class father:
...
+        p = Phone(None)
+        class _config:
+            options = {"country": "ES", "phone_code": "34", "phone_len": 9, "phone_format": r"^[+]%s[- ][0-9]{3}[- ][0-9]{3}[- ][0-9]{3}$"}
```
frodrigo: Here is only the test settings. The real setting are in `osmose_config.py`

frodrigo:
Cherry-picked and fixed in 88055ef20359e456617ac0a9c6d416b70c39dfca

frodrigo:
It's done. The results are available.

Please report here any false positive.

I have disable the check of the number format. After reading Wikipedia, there is many possible formats.

frodrigo:
Spain is now the biggest country not ussing international prefix
http://osmose.openstreetmap.fr/en/errors/?item=3092

PR #575, opened 2019-07-29
configure 1G /dev/shmem size for postgresql

stephankn:
Default docker limit of 64M is too small and PostgreSQL faces out of memory errors when executing large queries. Increasing limit to 1G to have enough headroom for PostgreSQL.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -2,6 +2,7 @@ version: '3.3'
 
 services:
   postgis:
+    shm_size: 1g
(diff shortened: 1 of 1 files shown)


frodrigo:
@jocelynj ?

jocelynj:
I've just checked the size of /dev/shm on the backends we have, and it is always bigger than 2G. Increasing it for docker seems fine.

frodrigo:
Ok. Merged.
Thank.

PR #582, opened 2019-08-05
Update main.py

fansanelli:
Improve regex to fix a sanity test


frodrigo:
Thank you.

HubMiner:
The change looks reasonable, is there a matching unit test change?

fansanelli:
@HubMiner I fear no.. I was reading the code to identify what can cause my error of day parsed as time and got a spurious result before the commit. BTW if you receive 99:77 you don't want to parse it as time, but just throwing an exception... I could add a unit to verify this.

Another issue that I think I have identified, even if I am not sure it can easily be solved... I'm getting '  24/7  ' to be parsed as: '24:00,07:00'.
I fear because the / is also a valid separator for time in the EBNF file, so the lexer the first time doesn't match the always open, but get a time rule follow another time rule with valid separator. The missing dow should be enforced to prevent this case

PR #601, opened 2019-09-02
Split Quebec

frodrigo:


frodrigo:
cc @jocelynj
cc @matthewdarwin

Waiting for extract to be generated.

jocelynj:
I'm seeing that Quebec relation, in addition to "Nord du Québec" doesn't contain a few islands along the coast: https://www.openstreetmap.org/relation/61549#map=11/56.4560/-76.5895

Are these islands in another Canadian Province?

frodrigo:
Yes Nunavut
https://www.openstreetmap.org/relation/390840

frodrigo:
@jocelynj the extract was generated yesterday, but the spited Quebec is not here.

Issue #603, opened 2019-09-04
duration syntax checker


Bibi56:
source | niv | item | cl | scl | E | pos | élmts | sous-titre | ✘/✔
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
  |   |   |   |   |   |   |   |  
  |   |   |   |   |   |   |   |  
407663 | 2 | 9014             transport en commun | 9014022 | 1468186952 | E | -4.50 48.39 | r          6674962                    (j) | La durée est invalide (essayer un nombre en minutes) | ✘/       ✔

On https://www.openstreetmap.org/relation/6674962, the duration is PT2M, this is a possible value according to the wiki https://wiki.openstreetmap.org/wiki/Key:duration

Of course I can change to 00:02, as this format is preferred, but a least the error message (and the level of error) is wrong.

frodrigo:
cc @nlehuby

frodrigo:
The issue should be reported to https://github.com/Jungle-Bus/transport_mapcss
But keep it here.

frodrigo:
@nlehuby it is feasible with mapcss/regex or not ?

nlehuby:
I won't have time to properly dig into the ISO 8601 spec before a few weeks, but I guess it's feasible with a regexp if anyone want to give it a try.

However, I'm still to sure if it is worth it as the usage is very low and this format is not supported by any known OSM data users 🤔

frodrigo:
So we remove this rule?

Bibi56:
@nlehuby, something like 
`^P(?!$)((\d+Y)|(\d+\.\d+Y$))?((\d+M)|(\d+\.\d+M$))?((\d+W)|(\d+\.\d+W$))?((\d+D)|(\d+\.\d+D$))?(T(?=\d)((\d+H)|(\d+\.\d+H$))?((\d+M)|(\d+\.\d+M$))?(\d+(\.\d+)?S)?)??$`
?
Found on [StackOverflow](https://stackoverflow.com/questions/32044846/regex-for-iso-8601-durations).

nlehuby:
The support of the ISO 8601 format for duration has been removed from the wiki a few month ago and the usage in the data is still very low.
I think we can remove the regexp added on last october.
Does it sound ok for you ?

frodrigo:
Remove sound very good to me!

frodrigo:
ISO format support removed.

PR #612, opened 2019-09-11
Add plugin for checking interval:conditional tag

PanierAvide:
Created a plugin to check `interval:conditional` tag using [transport-hours-py](https://github.com/Jungle-Bus/transport-hours-py) library :
- Syntax of tag should be correct regarding wiki doc
- Defined hour ranges should fit inside `opening_hours` ones if existing

This is mainly used on public transport routes, and this analyze will be useful for Jungle Bus team @nlehuby 

By the way, thanks for working on Docker README, this is really helpful and comprehensive :+1:

### DIFF
--- a/plugins/TagFix_IntervalConditional.py
+++ b/plugins/TagFix_IntervalConditional.py
@@ -25,31 +25,31 @@ class TagFix_IntervalConditional(Plugin):
 
     def init(self, logger):
         Plugin.init(self, logger)
-        self.errors[32502] = { "item": 3250, "level": 3, "tag": ["value", "fix:chair", "public_transport"], "desc": T_(u"Invalid Conditional Intervals") }
+        self.errors[32502] = { 'item': 3250, 'level': 3, 'tag': ['value', 'fix:chair', 'public_transport'], 'desc': T_(u'Invalid Conditional Intervals') }
         self._th = Main()
 
     def check_tags(self, tags):
-        if "interval" not in tags and "interval:conditional" not in tags:
+        if 'interval' not in tags and 'interval:conditional' not in tags:
             return
 
         # Analyse all tags related to transport hours (interval, interval:conditional, opening_hours)
         parsedData = self._th.tagsToHoursObject(tags)
 
         # Check validity of interval tag
-        if parsedData["defaultInterval"] == "invalid":
-            return { "class": 32502, "subclass": 1, "text": T_(u"Invalid interval tag format") }
+        if parsedData['defaultInterval'] == 'invalid':
+            return { 'class': 32502, 'subclass': 1, 'text': T_(u'Invalid interval tag format') }
 
         # Check validity of interval:conditional tag
-        if parsedData["otherIntervals"] == "invalid":
-            return { "class": 32502, "subclass": 2, "text": T_(u"Conditional intervals tag is not valid") }
+        if parsedData['otherIntervals'] == 'invalid':
+            return { 'class': 32502, 'subclass': 2, 'text': T_(u'Conditional intervals tag is not valid') }
 
         # Check presence of interval tag if interval:conditional is defined
-        if type(parsedData["otherIntervals"]) is list and parsedData["defaultInterval"] == "unset":
-            return { "class": 32502, "subclass": 3, "text": T_(u"An interval:conditional tag is defined without a default interval tag") }
+        if type(parsedData['otherIntervals']) is list and parsedData['defaultInterval'] == 'unset':
+            return { 'class': 32502, 'subclass': 3, 'text': T_(u'An interval:conditional tag is defined without a default interval tag') }
 
         # Check combination of opening_hours, interval and interval:conditional
-        if "opening_hours" in tags and type(parsedData["opens"]) is dict and parsedData["allComputedIntervals"] == "invalid":
-            return { "class": 32502, "subclass": 4, "text": T_(u"Conditional intervals does not fit into opening hours") }
+        if 'opening_hours' in tags and type(parsedData['opens']) is dict and parsedData['allComputedIntervals'] == 'invalid':
+            return { 'class': 32502, 'subclass': 4, 'text': T_(u'Conditional intervals does not fit into opening hours') }
 
 
     def relation(self, data, tags, members):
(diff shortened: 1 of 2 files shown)


frodrigo:
This PR solve this issue ? https://github.com/osm-fr/osmose-backend/issues/603

frodrigo:
@nlehuby https://github.com/Jungle-Bus/transport_mapcss/blob/master/transport.validator.mapcss#L321-L330

Need to remove this check ?

PanierAvide:
No, this one actually only checks `opening_hours` + `interval:conditional` tags. But the used library can check the syntax of `duration` tag, which is the same as `interval` tag. So if needed I can offer a PR for this too.

frodrigo:
Yes, in the same plugin ?

PanierAvide:
Didn't saw that duration tag has advanced formats like `P00010604` and so on. So this would either need an adaptation of used library, or be part of another PR.

frodrigo:
Ok about duration.

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,77 @@
...
+
+class Test(TestPluginCommon):
+    def test(self):
+        a = TagFix_IntervalConditional(None)
+        a.init(None)
+
+        self.check_err(a.relation(None, {'interval:conditional': 'Mo-Sa 15:00-17:00'}, None))
+        self.check_err(a.relation(None, {'interval:conditional': 'BLA @ (Mo-Sa 15:00-17:00)'}, None))
+        self.check_err(a.relation(None, {'interval:conditional': '00:15 @ (Mo-Sa 15:00-17:00); OUPS'}, None))
+        self.check_err(a.relation(None, {'interval:conditional': '15 @ (Mo-Sa 15:00-17:00)', 'opening_hours': 'Mo-Sa 16:30-18:30'}, None))
```
nlehuby: this one won't work if you don't specify the `interval` tag too.

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,77 @@
...
+        if 'interval:conditional' not in tags:
+            return
+
+        th = Main()
+
+        # Check interval:conditional by itself
+        try:
+            interval_cond_obj = th.intervalConditionalStringToObject(tags['interval:conditional'])
+        except Exception as e:
+            return {"class": 32502, "subclass": 1, 'text': {'en': str(e)}}
```
frodrigo: Does not report issues on exception. Let the analyser fails or ignore it.

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Check combination of interval:conditional + opening_hours
+        if 'opening_hours' in tags:
+            parsedData = th.tagsToHoursObject(tags)
+            if parsedData['allComputedIntervals'] == 'invalid':
+                return {"class": 32502, "subclass": 2, 'text': {'en': "Conditional interval definition doesn't fit in opening hours"}}
```
frodrigo: Make the text translatable with T_(u'')

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,77 @@
...
+##                                                                       ##
+###########################################################################
+from plugins.Plugin import Plugin
+
+try:
+    from transporthours.main import Main
+    module_transporthours = True
+except ImportError as e:
+    print(e)
+    module_transporthours = False
```
frodrigo: Simply import the module, an not almost silent fail.

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,77 @@
...
+        if not module_transporthours:
+            return False
+        Plugin.init(self, logger)
+        self.errors[32502] = {"item": 3250, "level": 3, "tag": ["value", "fix:chair"], "desc": T_(u"Invalid Conditional Intervals")}
+
+    def check_tags(self, tags):
+        if 'interval:conditional' not in tags:
+            return
+
+        th = Main()
```
frodrigo: Can you init the transporthours.Main in the init() and reuse it? Better perf.

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,77 @@
...
+    print(e)
+    module_transporthours = False
+
+class TagFix_IntervalConditional(Plugin):
+
+    def init(self, logger):
+        if not module_transporthours:
+            return False
+        Plugin.init(self, logger)
+        self.errors[32502] = {"item": 3250, "level": 3, "tag": ["value", "fix:chair"], "desc": T_(u"Invalid Conditional Intervals")}
```
nlehuby: maybe add `public_transport` to the tag list

PanierAvide:
Thanks for your feedback, I will improve the analyser in order to make more exhaustive checks on interval, interval:conditional and opneing_hours.

plugins/TagFix_IntervalConditional.py
PanierAvide: Should be fixed as now

plugins/TagFix_IntervalConditional.py
PanierAvide: Done

plugins/TagFix_IntervalConditional.py
PanierAvide: Done, I was using `TagFix_Opening_Hours.py` as a template, might not be a good one.

plugins/TagFix_IntervalConditional.py
PanierAvide: Done

plugins/TagFix_IntervalConditional.py
PanierAvide: Done

PanierAvide:
So I updated the plugin according to your feedbacks, this cover more cases and has more appropriate syntax.

plugins/TagFix_IntervalConditional.py
frodrigo: It is the more buggy contrib of Osmose... issues still open and nobody able to fix it

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,74 @@
...
+
+class Test(TestPluginCommon):
+    def test(self):
+        a = TagFix_IntervalConditional(None)
+        a.init(None)
+
+        self.check_err(a.relation(None, {"interval": "not minutes"}, None))# Invalid interval=*
+        self.check_err(a.relation(None, {"interval": "01:30", "interval:conditional": "Invalid conditional"}, None)) # Invalid interval:conditional=*
+        self.check_err(a.relation(None, {"interval:conditional": "15 @ (Mo-Sa 10:00-18:00)"}, None)) # interval:conditional without interval
+        self.check_err(a.relation(None, {"opening_hours": "Mo-Fr 05:00-23:00", "interval": "15", "interval:conditional": "5 @ (Mo-Sa 12:00-15:00)"}, None)) # interval:conditional don"t fit in opening_hours
```
frodrigo: Comment, simple quote.

plugins/TagFix_IntervalConditional.py
```diff
@@ -0,0 +1,74 @@
...
+        if parsedData["defaultInterval"] == "invalid":
+            return { "class": 32502, "subclass": 1, "text": T_(u"Invalid interval tag format") }
+
+        # Check validity of interval:conditional tag
+        if parsedData["otherIntervals"] == "invalid":
+            return { "class": 32502, "subclass": 2, "text": T_(u"Conditional intervals tag is not valid") }
+
+        # Check presence of interval tag if interval:conditional is defined
+        if type(parsedData["otherIntervals"]) is list and parsedData["defaultInterval"] == "unset":
+            return { "class": 32502, "subclass": 3, "text": T_(u"An interval:conditional tag is defined without a default interval tag") }
```
nlehuby: @frodrigo  we already have this check in mapcss (cl 9014023, https://github.com/Jungle-Bus/transport_mapcss/blob/master/transport.validator.mapcss#L335) but it totally makes sense here. Do we add a `-osmoseItemClassLevel: none;` to the mapcss one ?

frodrigo: It will be better to have it in mapcss, since mapcss code is reusable on other tools.

plugins/TagFix_IntervalConditional.py
PanierAvide: Done

plugins/TagFix_IntervalConditional.py
PanierAvide: I removed this check

PanierAvide:
Done the requested changes ;-)

frodrigo:
Please, can you look at Travis output.
You can ignore the fails about the reference OSM file.
But plugins.TagFix_IntervalConditional.Test also fails.

PanierAvide:
Done, this was the test for checking if interval is present when interval:conditional is set, I forgot to remove test while I removed code for this check.

frodrigo:
Merged.
Thank all.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?item=3250&class=32502

frodrigo:
The result from 1 server are here.

PanierAvide:
Yes I saw that, looks good ! :+1:

PR #617, opened 2019-09-17
4 first analysis for defibrillators

wadouk:
There is already 4 analysis. 
I would be interested by some feedback if i'm on the right path or some things to fix.
There is almost 10 more analysis to do.

### DIFF
--- a/analysers/analyser_merge_defibrillateurs_FR_basrhin.py
+++ b/analysers/analyser_merge_defibrillateurs_FR_basrhin.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Nicolas Bétheuil 2019                                      ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from .Analyser_Merge import Analyser_Merge, Source, GeoJSON, Load, Mapping, Select, Generate
+
+
+class Analyser_merge_defibrillateurs_FR_basrhin(Analyser_Merge):
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8321", "class": 10, "level": 3, "tag": ["merge"], "desc": T_(u"Defibrillator not integrated") }
+
+        Analyser_Merge.__init__(self, config, logger,
+            u"https://www.data.gouv.fr/fr/datasets/donnee-thematique-localisation-des-defibrillateurs-automatiques-externes-dae-bas-rhin/",
+            u"DONNEE THEMATIQUE : Localisation des Défibrillateurs Automatiques Externes (DAE) - Bas-Rhin",
+            GeoJSON(Source(attribution = u"data.gouv.fr:Service Départemental d'Incendie et de Secours du Bas-Rhin", millesime = "23/04/2018",
+                    fileUrl = u"https://www.data.gouv.fr/fr/datasets/r/40f84199-8b19-492d-9f61-45156ac73b63"),
+                    ),
+            Load("geom_x", "geom_y"),
+            Mapping(
+                select = Select(
+                    types = ["nodes"],
+                    tags = {"emergency": "defibrillator"}),
+                conflationDistance = 500,
+                generate = Generate(
+                    static1 = {
+                        "emergency": "defibrillator",
+                    },
+                    static2 = {"source": self.source},
+                text = lambda tags, fields: {"en": u"%s, %s, %s" % (fields["LOCALISATI"], fields["PRECISIONS"], fields["HORAIRES"])}
+                )
+            )
+        )
--- a/analysers/analyser_merge_defibrillators_FR_paris.py
+++ b/analysers/analyser_merge_defibrillators_FR_paris.py
@@ -44,4 +44,4 @@ class Analyser_Merge_defibrillators_FR_paris(Analyser_Merge):
                         "emergency": "defibrillator",
                     },
                     static2 = {"source": self.source},
-                text = lambda tags, fields: {"en": u"%s, %s, %s" % (fields["Type"], fields["Nom"], fields["Addresse"])},)))
+                text = lambda tags, fields: {"en": ', '.join(filter(lambda x: x, [fields["Type"], fields["Nom"], fields["Addresse"]]))},)))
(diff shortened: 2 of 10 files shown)


wadouk:
Hum, don't understand the build error. nothing to deal with those changes.

osmose_config.py
```diff
@@ -355,12 +357,14 @@ class gen(default_country):
 france_departement("ile_de_france/paris", 71525, "FR-75", include=include_ile_de_france + [
     # Paris
     'merge_bicycle_parking_FR_paris',
+    'merge_defibrillateurs_FR_paris'
```
frodrigo: Please, keep "," at the end.

osmose_config.py
```diff
@@ -532,6 +537,7 @@ class gen(default_country):
 france_local_db.analyser["merge_pitch_FR"] = "xxx"
 france_local_db.analyser["merge_police_FR_gn"] = "xxx"
 france_local_db.analyser["merge_police_FR_pn"] = "xxx"
+france_local_db.analyser["merge_defibrillateurs_FR_basrhin"] = "xxx"
```
frodrigo: Adds it directly to départements.

analysers/analyser_merge_defibrillateurs_FR_basrhin.py
```diff
@@ -0,0 +1,50 @@
...
+                conflationDistance = 500,
+                generate = Generate(
+                    static1 = {
+                        "emergency": "defibrillator",
+                    },
+                    static2 = {"source": self.source},
+                text = lambda tags, fields: {"en": u"%s, %s, %s" % (fields["LOCALISATI"], fields["PRECISIONS"], fields["HORAIRES"])}
+                )
+            )
+        )
```
frodrigo: I know it is not usual, but can you, follow the closing schema of other merge source files, please.

analysers/analyser_merge_defibrillateurs_FR_basrhin.py
```diff
@@ -0,0 +1,50 @@
...
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from .Analyser_Merge import Analyser_Merge, Source, GeoJSON, Load, Mapping, Select, Generate
+
+
+class Analyser_merge_defibrillateurs_FR_basrhin(Analyser_Merge):
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8321", "class": 10, "level": 3, "tag": ["merge"], "desc": T_(u"Defibrillator not integrated") }
```
frodrigo: Use a new class at each analyser, 10, 20, 30...

frodrigo: Use item 8370

wadouk:
I will fix those quickly thanks for all comments

analysers/analyser_merge_defibrillateurs_FR_basrhin.py
wadouk: only for basrhin or for all analysis ?

wadouk: what is the meaning of class ? I have put the same on all analysis, in each country there will be a different, maybe some data set will cover back some others, one class by analyser ?

frodrigo:
All analysers

frodrigo:
Class must be unique by source (analyser/country). But it easier to have
one by analyser/item.

analysers/analyser_merge_defibrillateurs_FR_basrhin.py
frodrigo: Please inline more ) and }. It is for consistency.

analysers/analyser_merge_defibrillateurs_FR_basrhin.py
frodrigo: Since the analyse is not the same for each area, you can reuse the same class. But for practical it does not help to reuse the class.

Please, use 10, 20, 30, not 1, 2, 3, 4. We keep the last digit for other usage in merge analyser (missing_official, missing_osm, update...).

frodrigo: For all, it is now OK. An item is an entry on the frontend menu.

analysers/analyser_merge_defibrillators_FR_paris.py
```diff
@@ -0,0 +1,48 @@
...
+class Analyser_Merge_defibrillators_FR_paris(Analyser_Merge):
+    def __init__(self, config, logger = None):
+        self.missing_official = {"item":"8370", "class": 3, "level": 3, "tag": ["merge"], "desc": T_(u"Defibrillator not integrated") }
+
+        Analyser_Merge.__init__(self, config, logger,
+            u"https://www.data.gouv.fr/fr/datasets/batiments-publics-equipes-de-defibrillateurs-a-paris/",
+            u"Défibrillateurs à Paris",
+            CSV(Source(attribution = u"data.gouv.fr:Mairie de Paris", millesime = "23/04/2018",
+                    fileUrl = u"https://www.data.gouv.fr/fr/datasets/r/d2b94d3d-977d-4dcb-ab8c-77e2568de736"),
+                quote=u'"',
```
frodrigo: Not required, as it is the default quote.

analysers/analyser_merge_defibrillateurs_FR_basrhin.py
wadouk: I will get more that 10 analyzers, ok to go up to 100 ?

frodrigo: yes

analysers/analyser_merge_defibrillators_FR_paris.py
```diff
@@ -0,0 +1,47 @@
...
+                select = Select(
+                    types = ["nodes"],
+                    tags = {"emergency": "defibrillator"}),
+                conflationDistance = 50,
+                generate = Generate(
+                    static1 = {
+                        "emergency": "defibrillator",
+                    },
+                    static2 = {"source": self.source},
+                text = lambda tags, fields: {"en": u"%s, %s, %s" % (fields["Type"], fields["Nom"], fields["Addresse"])},)))
```
frodrigo: Please, filter the fields to avoid None in text: `<text lang="en" value="Stade, Stade Porte de la Muette, None" />`

Like this: https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_merge_tourism_FR_aquitaine_camp_caravan.py#L50

Maybe also do it for the others.

osmose_config.py
```diff
@@ -355,12 +357,14 @@ class gen(default_country):
...
     # Paris
     'merge_bicycle_parking_FR_paris',
+    'merge_defibrillators_FR_paris',
 ], exclude=[
     'merge_shop_FR',
 ])
 france_departement("ile_de_france/hauts_de_seine", 7449, "FR-92", include=include_ile_de_france + [
     # Hauts-de-Seine
     'merge_restriction_FR_92',
+    'merge_defibrillators_FR_issylesmoulineaux',
```
frodrigo: This analyser return no issues. Is that the desired behaviour?

wadouk:
For this one, it seems to be the point. This city seems well covered.

frodrigo:
#596

analysers/analyser_merge_defibrillators_FR_paris.py
wadouk: A noob question about python : would like to check syntax before bush, how to just "compile" without running ?

frodrigo: `python analysers/analyser_merge_defibrillators_FR_paris.py`

frodrigo:
Merged, thank!

frodrigo:
#596

PR #636, opened 2019-09-30
Add ref:vatin test V2

fansanelli:
New version with most of comments addressed

### DIFF
--- a/plugins/TagFix_Vatin.py
+++ b/plugins/TagFix_Vatin.py
@@ -60,6 +60,9 @@ class TagFix_Vatin(Plugin):
             if tags["ref:vatin"].startswith("IT"):
                 if self.it_vatin(tags["ref:vatin"][2:]) == False:
                     return {"class": 99999, "subclass": 1 }
+            else:
+                if len(tags["ref:vatin"]) < 3 or tags["ref:vatin"][0:2].isalpha() == False or tags["ref:vatin"].isupper() == False:
+                    return {"class": 99999, "subclass": 0 }
 
     def way(self, data, tags, nds):
         return self.node(data, tags)
@@ -81,6 +84,12 @@ class Test(TestPluginCommon):
         assert not a.way(None, {"ref:vatin": "IT11111111115"}, None)
         assert not a.relation(None, {"ref:vatin": "IT11111111115"}, None)
         assert not a.node(None, {"ref:vatin": "DE115055186"})
+        # country code should be uppercase
+        assert a.node(None, {"ref:vatin": "it11111111111"})
+        # check digit is wrong
         assert a.node(None, {"ref:vatin": "IT11111111111"})
+        # only digits allowed in Italy
         assert a.node(None, {"ref:vatin": "ITAAAAAAAAAAA"})
+        # missing country code
+        assert a.node(None, {"ref:vatin": "11111111115"})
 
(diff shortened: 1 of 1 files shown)


frodrigo:
Note, you can push on your branch to update the pull request.

plugins/TagFix_Vatin.py
```diff
@@ -0,0 +1,89 @@
...
+            y += 2 * (int(vatin[i:i + 1]))
+            if int(vatin[i:i + 1]) >= 5:
+                z = z + 1
+        t = (x + y + z) % 10
+        c = int(vatin[10:11])
+        return c == (10 - t) % 10
+
+    def node(self, data, tags):
+        if "ref:vatin" in tags:
+            x = re.split("^([A-Z]{2})(.+)$", tags["ref:vatin"])
```
frodrigo: Precompile the regex in the init. But here you does not need a regex `if tags["ref:vatin"].startswith('IT')`

frodrigo: and `tags["ref:vatin"]|2:]` for the last part.

fansanelli: Well, consider that any country code (IT, DE, FR,...) followed by a code (alphanum) is valid...
I am not a python expert, how should I pre-compile it?

frodrigo: Not sure what you want to achieve here. But avoid using regex if not required.
You can check the first two are country code like https://docs.python.org/fr/2/library/stdtypes.html?highlight=isalpha#str.isalpha

To compile regex https://docs.python.org/2/library/re.html#re.compile

fansanelli: The idea was to check that the pattern <CountryCode><VAT> is respected, but could achieved in a later revision of this file

fansanelli:
I slept on this. Please wait before merging I want to do restore the check inside an else branch

fansanelli:
I'm quite happy with it, even without the regex. Please let me know if we're done

frodrigo:
Merged. Thank you.
I will ping you when the code will be deployed.

fansanelli:
Thank you so much

frodrigo:
Results are here, please confirm it's OK.
http://osmose.openstreetmap.fr/fr/errors/?item=3260&useDevItem=all
This new item is still hidden.

frodrigo:
Humm, can you try to report more details? like bad length, bad checkesum, checksum is missing, country code is missing...

fansanelli:
Hello,
I've checked some values (first ten or so) and I can confirm all of them should be fixed...
Anyway my code has 2 sub errors... 0 is "invalid format" and 1 "bad Italian partita IVA"

frodrigo:
The display can be improved by using multiple class (self.errors[99999]) and/or a subtitle description (text): https://github.com/osm-fr/osmose-backend/blob/master/plugins/Source.py#L45

fansanelli:
Do you think I can add the texts above or should I create more subclasses?

frodrigo:
You can do both (class, no subclasses)

frodrigo:
@jocelynj please can you enable item 3260?

frodrigo:
Item publicly available now
http://osmose.openstreetmap.fr/en/map/#item=3260&zoom=6&lat=44.72&lon=11.96&level=1%2C2%2C3&tags=&fixable=

fansanelli:
Thanks for the help!

PR #637, opened 2019-10-01
remove dead code

stephankn:
in 563a71768 you changed the code to return on the first match.
This leaves dead code which this change removes to keep the source readable.


frodrigo:
Thank

PR #638, opened 2019-10-01
init base class in constructor

stephankn:
I think the base class should also be initialized from the constructor.  As Geocode_Addok_CSV is not accepting keyword arguments for base, calling init with default arguments.

### DIFF
--- a/analysers/Analyser_Merge_Geocode_Addok_CSV.py
+++ b/analysers/Analyser_Merge_Geocode_Addok_CSV.py
@@ -28,6 +28,7 @@ from .modules import downloader
 class Geocode_Addok_CSV(Source):
 
     def __init__(self, source, columns, logger, citycode = None, delimiter = ',', encoding = 'utf-8'):
+        Source.__init__(self)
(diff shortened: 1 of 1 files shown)


stephankn:
@frodrigo please check whether this makes sense.

frodrigo:
It does not make sense. Maybe this class is not a sub class of Source. Or Source and Geocode_Addok_CSV must have a common super class.

stephankn:
@frodrigo So syntactically it actually **is** a subclass of Source:

    class Geocode_Addok_CSV(Source):

What do you think it should be? In any case, the __init__ is not calling the base class init, which should be done.

frodrigo:
Have to check if it works with only `class Geocode_Addok_CSV:`

stephankn:
when you introduced the `Geocode_Addok_CSV(Source)` you used it to replace another Source: `class Power_Plant_FR_Source(Source):`

So at least from that perspective it sounded reasonable that the new class is also a `Source`.

PR #639, opened 2019-10-01
remove unreachable code

stephankn:
rmtree is never executed here, as either try-block returns in all possible code-paths or an exception is re-raised.
With the rmtree called always on entry of the function it sounds ok, not to call it.


frodrigo:
Look good thank.

PR #640, opened 2019-10-02
function overloading not possible in Python

stephankn:
It is not possible to have multiple functions sharing the same name and differing in the parameters.

Removed the not working one and provided an implementation using default parameters to provide the intended signature. Slicing with None as the end parameter works the same way as leaving it out.

Also replaces the test by checking for identity instead of equality to speed it up a bit and comply with PEP8.

### DIFF
--- a/mapcss/mapcss_lib.py
+++ b/mapcss/mapcss_lib.py
@@ -461,6 +461,8 @@ def regexp_match(regexp, string):
 #regexp_match(regexp, string, flags)
 #    Tries to match string against pattern regexp. Returns a list of capture groups in case of success. The first element (index 0) is the complete match (i.e. string). Further elements correspond to the bracketed parts of the regular expression. Flags is a string that may contain "i" (case insensitive), "m" (multiline) and "s" ("dot all") [since 5701] 
 
+#substring(str, idx)	
+#    return the substring of str, starting at index idx (0-indexed) [since 6534] 
 #substring(str, start, end)
 #    return the substring of str, starting at index start (inclusive) up to end (exclusive) (0-indexed) [since 6534] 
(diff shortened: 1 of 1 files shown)


mapcss/mapcss_lib.py
```diff
@@ -461,16 +461,10 @@ def regexp_match(regexp, string):
 #regexp_match(regexp, string, flags)
 #    Tries to match string against pattern regexp. Returns a list of capture groups in case of success. The first element (index 0) is the complete match (i.e. string). Further elements correspond to the bracketed parts of the regular expression. Flags is a string that may contain "i" (case insensitive), "m" (multiline) and "s" ("dot all") [since 5701] 
 
-#substring(str, idx)
-#    return the substring of str, starting at index idx (0-indexed) [since 6534]
```
frodrigo: Please keep this comment, as it notes we implement here this mapcss function.

stephankn: ok. I added that back

mapcss/mapcss_lib.py
```diff
@@ -461,16 +461,12 @@ def regexp_match(regexp, string):
 #regexp_match(regexp, string, flags)
 #    Tries to match string against pattern regexp. Returns a list of capture groups in case of success. The first element (index 0) is the complete match (i.e. string). Further elements correspond to the bracketed parts of the regular expression. Flags is a string that may contain "i" (case insensitive), "m" (multiline) and "s" ("dot all") [since 5701] 
 
-#substring(str, idx)
+#substring(str, idx)
```
frodrigo: Sorry, please remove the extra space.

frodrigo:
Mergerd, than you.

PR #643, opened 2019-10-03
osmosis need integer for maxInterval

tkas:
osmosis need integer for maxInterval for configuration.txt, with python3.6, the current code produces:

maxInterval=43200.0

for geofabrik with diffs, which ends osmosis import via error:
java.lang.NumberFormatException: For input string: "43200.0"

### DIFF
--- a/modules/OsmOsisManager.py
+++ b/modules/OsmOsisManager.py
@@ -297,7 +297,7 @@ class OsmOsisManager:
         elif line.startswith("maxInterval"):
             if "geofabrik" in conf.download["diff"]:
                 # on daily diffs provided by Geofabrik, we should apply only one diff at a time
-                sys.stdout.write("maxInterval=" + str(round(60*60*24/2))) # 1/2 day at most
+                sys.stdout.write("maxInterval=" + str(60*60*24//2)) # 1/2 day at most
             else:
                 sys.stdout.write("maxInterval=" + str(7*60*60*24)) # 7 day at most
         else:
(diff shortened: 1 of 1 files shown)


modules/OsmOsisManager.py
```diff
@@ -297,7 +297,7 @@ def init_diff(self, conf):
         elif line.startswith("maxInterval"):
             if "geofabrik" in conf.download["diff"]:
                 # on daily diffs provided by Geofabrik, we should apply only one diff at a time
-                sys.stdout.write("maxInterval=" + str(60*60*24/2)) # 1/2 day at most
+                sys.stdout.write("maxInterval=" + str(round(60*60*24/2))) # 1/2 day at most
             else:
                 sys.stdout.write("maxInterval=" + str(7*60*60*24)) # 7 day at most
```
stephankn: Better use **//**
as per https://www.python.org/dev/peps/pep-0238/

    $ docker run -ti python:3.7 bash
    root@ff315ac86108:~# echo 'print("maxInterval=" + str(60*60*24/2))' | python
    maxInterval=43200.0
    root@449e4ea1e59c:/# echo 'print("maxInterval=" + str(60*60*24//2))' | python
    maxInterval=43200

tkas: OK, I'm fine with this version too.

frodrigo: Please, change the commit to `//`

tkas: Done

frodrigo:
Thank you all.

PR #645, opened 2019-10-05
fix invalid format string

stephankn:


frodrigo:
Ok. But it is dead code.

PR #646, opened 2019-10-05
use integer division

stephankn:
The tags are multiples of two, so using integer division sounds
reasonable


frodrigo:
Ok. But it is also dead code.

PR #647, opened 2019-10-05
fix unexpected trunc to integer

stephankn:
worked only because division truncated to integer, so that
19/10 returned 1 instead of 1.9 which then resulted in 1.5 to
pass to round to return the ceiling.

Instead of writing the round with integer division // I use
the ceil function to express clearer what is meant here.


frodrigo:
19/10 -> 1.9 un py3 and 1 in py2

But your fix is right. It works well in all case.

Thank.

PR #651, opened 2019-10-06
remove unused imports to improve read-/maintainability

stephankn:


analysers/analyser_merge_power_plant_FR.py
```diff
@@ -20,12 +20,10 @@
 ##                                                                       ##
 ###########################################################################
 
-import re
 from io import open # In python3 only, this import is not required
 from .Analyser_Merge import Analyser_Merge, Source, CSV, Load, Mapping, Select, Generate
 from .Analyser_Merge_Geocode_Addok_CSV import Geocode_Addok_CSV
-from .modules import downloader
-from .modules import Stablehash
+from modules import Stablehash
```
frodrigo: Please, keep relative import.

frodrigo:
cc @jocelynj

analysers/analyser_merge_power_plant_FR.py
stephankn: This is inside analysers. Should then be `..modules` I think. Other locations always import from `modules` by absolute path. For example in Analyser_Merge_Mapillary or analyser_sax. So this sounded more common.

frodrigo: ok

jocelynj:
@stephankn Thanks for the cleaning.

Can you check that `./osmose_run.py` works correctly after these changes, with python2 and python3?

Tests are passing in python2 according to Travis, but it would be good to check with python3 too (you can modify `./tools/pynose.sh` to launch python3 on the last line)

analysers/analyser_merge_power_plant_FR.py
frodrigo: Again on this one. Please can you remove this change from this commit. As there is others `from .modules` but untouched by this commit. We may work on this in an other PR. Thank.

stephankn: ok, I keep it as .modules. I still think this is wrong and can currently only guess why Python manges to resolve it.

frodrigo:
Ok for me. I will merge it.

analysers/analyser_merge_power_plant_FR.py
stephankn: Just seen why python manages to resolve in the wrong location. There are symbolic links in place.

frodrigo:
Merged. Thank.

analysers/analyser_merge_power_plant_FR.py
frodrigo: I think, it is required for tests.

PR #652, opened 2019-10-06
test for identity, not equality

stephankn:
better readability and potentially a tiny performance benefit

### DIFF
--- a/analysers/Analyser.py
+++ b/analysers/Analyser.py
@@ -313,7 +313,7 @@ class TestAnalyser(unittest.TestCase):
         if a != b:
             s = TestAnalyser.compare_dict(a, b)
             print(s)
-            assert s == None, "results differ"
+            assert s is None, "results differ"
             self.assertEquals(a, b, "results differ")
 
 
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -368,7 +368,7 @@ class Source:
         if "%s" in self.attribution:
             return self.attribution % self.millesime
         else:
-            return " - ".join(filter(lambda x: x!= None, [self.attribution, self.millesime]))
+            return " - ".join(filter(lambda x: x is not None, [self.attribution, self.millesime]))
 
     def match_attribution(self, s):
         if "%s" not in self.attribution:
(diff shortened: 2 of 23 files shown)


frodrigo:
Readability is subjective, and I know about PEP8 and so on.
I will not take changes of `is` neither the changes about indentation.

stephankn:
Don't forget about the performance penalty for the extra call. Also it helps to make the code reports more readable. 10% of the warnings relate to this issue.

When you compare an object to None, use is rather than ==. None is a singleton object, comparing using == invokes the __eq__ method on the object in question, which may be slower than identity comparison.

jocelynj:
@stephankn What kind of tool are using which prefers "a is not None" instead of "a != None" ?

stephankn:
@jocelynj regarding the tools:

I tried to clean up Semmle LGTM
https://lgtm.com/projects/g/osm-fr/osmose-backend/

as they now belong to github and will be soon available in a much closer integration.
They also focus more on security, which sounds like a good idea to use alongside other security features.
https://github.com/features/security

Python guarantees the None to be a singleton, so no need to do a function call for equality.
https://docs.python.org/3/library/stdtypes.html#the-null-object

So roughly the same reason for not comparing strings in Java using ==.

PyCharm inspections warns about it:
https://www.jetbrains.com/help/pycharm/code-inspection.html

PEP8 does as E711:
https://pep8.readthedocs.io/en/release-1.7.x/_modules/pep8.html

Pylint does as C0121:
https://pylint.readthedocs.io/en/latest/technical_reference/features.html

Here someone benchmarked it and the identity operation is 73% faster:
http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

frodrigo:
So, ok for "is". Can you remove all indentation changes of this PR, please ?

Secondly, warning about "is" in mapcss_lib.py. Becasue we overide _eq_ https://github.com/osm-fr/osmose-backend/blob/master/mapcss/mapcss_lib.py#L128

stephankn:
I reverted the unwanted indentation changes. Regarding the mapcss __eq__ overloading:

It is implemented as:

    def __eq__(self, o):
        if self.none:
            return False
        elif o.__class__ in (int, long):
            return self.to_n() == o
        else:
            return super(str_value_, self).__eq__(o)

In we we do `t == None`, then `self.none` is expected to be false. Also the class of `None` is not int/long. So we would end in super.__eq__. With this being `unicode` I don't expect it to do any crazy things when comparing versus `None`, so it should be fine.

stephankn:
I did rebase this PR on 4a3f050 for you to easier merge it. Maybe a bad idea, as the previous merged change seems to break travis:  https://travis-ci.com/osm-fr/osmose-backend/builds/130932491

frodrigo:
Merged. Thank you.

jocelynj:
@stephankn Thanks for the informations about this tool. LGTM looks very useful, and applying its suggestions should make the code better.

PR #653, opened 2019-10-06
duplicate entry in dictionary

stephankn:
having text two times is certainly broken. I keep the recently added format.


stephankn:
@frodrigo Is it expected to have no `fix` here?

frodrigo:
No, bad format is not auto-fixable.

Thank.

PR #654, opened 2019-10-06
remove empty disabled plugin

stephankn:
plugin is already in disabled folder, so comment out dead code
to get rid of warning.


frodrigo:
I even prefer to delete the file than apply this inconstant patch.

stephankn:
ok, changed to a remove of this empty code blob

frodrigo:
cc @jocelynj

jocelynj:
@frodrigo I don't have any objection in removing disabled plugins, especially as this one was disabled a long time ago.

frodrigo:
Ok. Let's remove this one.

PR #661, opened 2019-10-11
replace unneeded finally by with statement

stephankn:
`finally` was only used to close the file, so replacing by `with` statement.

see https://lgtm.com/rules/6770081/


frodrigo:
Thank. Merged.

PR #662, opened 2019-10-12
encoding unused since bca68e9

stephankn:


analysers/analyser_merge_healthcare_FR_finess.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, config, error_file, logger, srid, is_in, categories, items, m
 class Source_Finess(Source):
     def open(self):
         # Cheat the parent open
-        encoding, self.encoding = self.encoding, 'UTF-8'
+        self.encoding = self.encoding, 'UTF-8'
```
frodrigo:
```suggestion
self.encoding = 'UTF-8'
```

stephankn:
Oh, right. I missed the comma on the right

PR #663, opened 2019-10-12
describe different files in the tools folder

stephankn:


stephankn:
please add details about other tools after merge

frodrigo:
cc @jocelynj

jocelynj:
That's a good idea. I think some scripts are no longer used, and could be removed.

frodrigo:
Merged as is.

PR #664, opened 2019-10-12
adjust grep call as pattern is an extended regex

stephankn:


frodrigo:
cc @jocelynj

stephankn:
standard grep will not match, script is currently broken. It does not filter out the two unwanted pattern. Alternative way of fixing would be calling `egrep` in case your operating system supports this way of calling grep which would be similar to `grep -E`.

frodrigo:
Merged, but it is unmaintained script and unused.

Issue #674, opened 2019-10-22
Do not report "Highway above ground" for man_made=pier


danfos:
See http://osmose.openstreetmap.fr/en/map/#item=4110&zoom=17&lat=51.914434&lon=4.250668&level=3, two the same problems:

source=4682
level=3
item=4110 incorrect layer
class=41105

A pier above the water, so layer=1 makes sense I think.

The likely fix is:

```
diff --git a/plugins/TagRemove_Layer.py b/plugins/TagRemove_Layer.py
index 41d0094e..a84208d6 100644
--- a/plugins/TagRemove_Layer.py
+++ b/plugins/TagRemove_Layer.py
@@ -47,7 +47,7 @@ class TagRemove_Layer(Plugin):
             elif tags.get(u"highway") and tags.get(u"highway") != "steps" and (not tags.get(u"indoor") or tags.get(u"indoor") == "no"):
                 if layer[0] == "-" and (not tags.get(u"tunnel") or tags.get(u"tunnel") == "no"):
                     return {"class": 41104 if len(nds) > 3 else 41103, "subclass": 0}
-                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no"):
+                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no" or tags.get(u"man_made") == "pier"):
                     if len(nds) > 3:
                         return {"class": 41106, "subclass": 0, "fix": {"-": ["layer"]}}
                     else:
```

Could not test it tough as I do not have Osmose running.

frodrigo:
Please make a Pull Request and let Travis CI check it. Optionally add a test ant the end of TagRemove_Layer.py. Thank.

frodrigo:
#675

danfos:
Thanks for reviewing the change and committing the updated code!

PR #675, opened 2019-10-22
Do not report "Highway above ground" for man_made=pier

danfos:
See #674, for the problem to solve, see http://osmose.openstreetmap.fr/en/map/#item=4110&zoom=17&lat=51.914434&lon=4.250668&level=3

### DIFF
--- a/plugins/TagRemove_Layer.py
+++ b/plugins/TagRemove_Layer.py
@@ -47,7 +47,7 @@ class TagRemove_Layer(Plugin):
             elif tags.get(u"highway") and tags.get(u"highway") != "steps" and (not tags.get(u"indoor") or tags.get(u"indoor") == "no"):
                 if layer[0] == "-" and (not tags.get(u"tunnel") or tags.get(u"tunnel") == "no"):
                     return {"class": 41104 if len(nds) > 3 else 41103, "subclass": 0}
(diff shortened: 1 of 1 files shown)


plugins/TagRemove_Layer.py
```diff
@@ -47,7 +47,7 @@ def way(self, data, tags, nds):
             elif tags.get(u"highway") and tags.get(u"highway") != "steps" and (not tags.get(u"indoor") or tags.get(u"indoor") == "no"):
                 if layer[0] == "-" and (not tags.get(u"tunnel") or tags.get(u"tunnel") == "no"):
                     return {"class": 41104 if len(nds) > 3 else 41103, "subclass": 0}
-                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no"):
+                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no" or tags.get(u"man_made") == "pier"):
```
frodrigo: I think it should be `!= 'pier'`. But to be more clear
```suggestion
                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no") and tags.get(u"man_made") != "pier":
```

danfos:
Yes, correct, took over your suggestion.

frodrigo:
Merged. Thank.

PR #676, opened 2019-10-23
Let FANTOIR B to W type also include islet

danfos:
I was seeing these three reported problems:

http://osmose.openstreetmap.fr/en/map/#item=2060&zoom=17&lat=45.703616&lon=-0.329429&level=3&tags=&fixable=

So "FANTOIR B to W type is for locality, hamlet, isolated_dwelling or neighbourhood"

While I do not understand the "FANTOIR B to W type" an islet is also an isolated_dwelling (https://wiki.openstreetmap.org/wiki/Tag:place%3Dislet) so it seems like a good idea to me to include islet also in the list.


frodrigo:
Good catch, thank you.

PR #680, opened 2019-10-26
Add check for bicycle=no|use_sidepath in combination with cycleway=

danfos:
Far from sure this is the correct way to add a new test and if the code is correct but I would like to add a check for bicycle=no|use_sidepath in combination with cycleway=, I think the code is reasonable clear on what is it is meant to do.

See https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Duse_sidepath on the why:

_Please do not use bicycle=use_sidepath in combination with cycleway=track on the main highway, if there is no separate cycleway drawn on the map. This is confusing for routing engines because bicycle=use_sidepath indicates not to route on the main highway (use the separate cycleway) and this conflicts with the tag cycleway=track, which indicates that bicycles are allowed to route on (bike lanes connected to) the main highway._

The allowed values for cycleway= are based on https://taginfo.openstreetmap.org/keys/cycleway#values

Questions I have:
- Good idea?
- Good to re-use error class 30328, I see that now reporting only "highway=cycleway with cycleway=track", http://osmose.openstreetmap.fr/en/errors/?item=3032&class=30328
- I see there is also a plugins/Bicycle.validator.mapcss file should that also be updated?

### DIFF
--- a/plugins/Bicycle.py
+++ b/plugins/Bicycle.py
@@ -15,10 +15,13 @@ class Bicycle(Plugin):
         self.errors[20302] = {'item': 2030, 'level': 1, 'tag': mapcss.list_(u'tag', u'highway') + mapcss.list_(u'cycleway', u'fix:survey'), 'desc': mapcss.tr(u'Opposite or opposite lane in the same way of the oneway')}
         self.errors[20805] = {'item': 2080, 'level': 3, 'tag': mapcss.list_(u'tag', u'highway') + mapcss.list_(u'footway', u'fix:chair'), 'desc': mapcss.tr(u'{0} without {1}', mapcss._tag_uncapture(capture_tags, u'{0.tag}'), u'highway=footway|construction')}
         self.errors[30328] = {'item': 3032, 'level': 2, 'tag': mapcss.list_(u'tag', u'highway') + mapcss.list_(u'cycleway', u'fix:chair'), 'desc': mapcss.tr(u'{0} with {1}', mapcss._tag_uncapture(capture_tags, u'{0.tag}'), mapcss._tag_uncapture(capture_tags, u'{1.tag}'))}
+        self.errors[30329] = {'item': 3032, 'level': 2, 'tag': mapcss.list_(u'tag', u'highway') + mapcss.list_(u'fix:chair'), 'desc': mapcss.tr(u'{0} with {1}', mapcss._tag_uncapture(capture_tags, u'{0.tag}'), mapcss._tag_uncapture(capture_tags, u'{1.tag}'))}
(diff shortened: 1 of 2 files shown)


frodrigo:
I think the idea is good.

The plugins/Bicycle.validator.mapcss is the source, where the test must be written. plugins/Bicycle.py is a generated file from the mapcss using the tools mapcss2osmose from the mapcss directory.

```css
        # way["bicycle"="no|use_sidepath"]["cycleway"]["cycleway"!="no|none|separate"];
        # way["bicycle"="no|use_sidepath"]["cycleway:left"]["cycleway:left"!="no|none|separate"];
        # way["bicycle"="no|use_sidepath"]["cycleway:right"]["cycleway:right"!="no|none|separate"];
```

Report this to the mapcss file, add some asserts. You should not use `=` but regex match.
Look at https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation for the mapcss language.

You can check the mapcss file in JOSM. If you don't have the ability to generate the python file I will do it.

I think all this part is completely doc less.

danfos:
Generating the python file proved to be easier than trying the code in JOSM so I did so.
I see Travis does see a problem with the code, let me have a look at it.

danfos:
Understood why there is a problem, should check if the key is present before checking if the key value != no|none|separate

Let me upload the updated code, not 100% sure the generated code is correct though

danfos:
It looks one step further, the code did pass the asserts I think but I re-useing error class 30328 was not a good idea I think, it looks like the original error class is overwitten and, as the parameters are a bit different, the higher layer code is not seeing what it is expecting.

Let me change it to use a new class, 30329.

danfos:
Also changing the code to use a new class, 30329 is still triggering an error.

Not 100% sure what to do, should the front-end also be update or should something else be done?

frodrigo:
Just ignore this Travis test error. It is not directly linked to your changes. The tests contains a small OSM extract, and your commit change the result on it.

frodrigo:
It looks good for me now.

danfos:
Thanks for checking, good to know the latest code is okay.

frodrigo:
Merged. Thank.

danfos:
Thanks for merging, I see the data is now available:

http://osmose.openstreetmap.fr/en/errors/?item=3032&class=30329

Not too much problems apart from [brazil_sao_paulo](http://osmose.openstreetmap.fr/en/errors/?source=37211&item=3032&class=30329)

frodrigo:
We just deploy yesterday late. It take 24h to update the full planet.

PR #681, opened 2019-10-26
catch matching exception instead of base class

stephankn:
as we are expecting an ImportError in py2, handle it.
fixes a warning of https://lgtm.com/rules/6780080/


frodrigo:
Ok. Thank.

PR #682, opened 2019-10-26
complete io already imported, simplify bool

stephankn:


frodrigo:
No.
```
from io import open # In python3 only, this import is not required
```
If you remove this, in the code `open()` should be `io.open()` to use the open from py3 in py2. But the goal is to remove all py2 specific code when the switch to py3 will be done.

PR #687, opened 2019-11-01
Stablehash64

frodrigo:
Avoid hash conflict on subclass.

Waiting for frontend subclass switch to 64bits data type.


jocelynj:
Is the difference with previous hash the size of the outputted hash, going from 32 bits to 64 bits?

frodrigo:
yes

PR #693, opened 2019-11-06
fix toulouse defibrillators

wadouk:
from https://github.com/osm-fr/osmose-backend/issues/596#issuecomment-549541804


frodrigo:
OK. thank you.

PR #709, opened 2019-12-15
Add support checking Dutch (NL) house numbers

danfos:
I was seeing a warning generated for valid Dutch (NL) house numbers so did look up the documentation and added code to support checking Dutch (NL) house numbers.

### DIFF
--- a/plugins/TagFix_Housenumber.py
+++ b/plugins/TagFix_Housenumber.py
@@ -53,6 +53,9 @@ class TagFix_Housenumber(Plugin):
         self.housenumberRegexByCountry["LU"] = re.compile("^[1-9][0-9]{0,3}([A-Z]){0,3}(-[1-9][0-9]{0,3}([A-Z]){0,3})?$")
         # Allow "snc" (Senza numero civico) in Italy
         self.housenumberRegexByCountry["IT"] = re.compile("(:?^[1-9])|(^snc$)")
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank. The community is agree on using a such strict format for addresses number?

danfos:
I did not ping the community but the in the Netherlands almost all house data is imported from BAG, see https://wiki.openstreetmap.org/wiki/NL:BAG which is government data.

The link, https://imbag.github.io/catalogus/hoofdstukken/attributen--relaties#734-huisnummertoevoeging does specify the rules for that data and this PR is the implementation of that.

frodrigo:
Ok. Thank.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?country=netherlands_*&item=2060&class=10

Lot of issues are not raised (50 000 and still processing). Please check and fix the regex if required.

danfos:
Yes, a lot of issues for 4 provinces while the other have almost none, including the one I live in which is the most populated.

Had a look at the problems and they seems to be all of the type "building=houseboot" or "landuse=static_caravan", but that data is also coming from BAG, so two possibilities, excluding these (my preference) or extending the rules so that also the number type of these objects is also allowed.

The Caribbean Islands probably have to be excluded as they seem to follow not the official rules. 

No more time for today, let me come back on it tomorrow-evening.

frodrigo:
@danfos new issue about this https://github.com/osm-fr/osmose-backend/issues/711

PR #714, opened 2019-12-19
Remove barrier=kerb from highway=crossing

ecatmur:
The barrier=kerb prevents OSRM router from routing vehicle traffic through the crossing, and is incorrect - the kerb blocks the non-vehicle highway, not the vehicle highway. Remove barrier=kerb from the crossing node, and add kerb=yes if there is no kerb tag already present.

### DIFF
--- a/plugins/TagFix_MultipleTag.py
+++ b/plugins/TagFix_MultipleTag.py
@@ -81,6 +81,10 @@ class TagFix_MultipleTag(Plugin):
         if tags.get("barrier") == "fence" and "fence_type" not in tags and "material" in tags:
             err.append({"class": 303210})
 
+        if tags.get("barrier") == "kerb" and tags.get("highway") == "crossing":
+            err.append({"class": 30320, "subclass": 1001, "text": T_(u"Use kerb=* to describe the kerb at a crossing, not barrier=kerb"),
+                       "fix": ({"-": {"barrier": "kerb"}} if "kerb" in tags else {"-": {"barrier": "kerb"}, "+": {"kerb": "yes"}})})
+
         return err
 
     def node(self, data, tags):


osm-spiregrain:
This would help identify instances that have the impacts described in the thread beginning here [in talk-gb](https://lists.openstreetmap.org/pipermail/talk-gb/2019-December/023920.html).

FloatingBoater:
+1

Note - the original end-user report of VERY indirect routing by OSRM in my talk-gb post now routes normally after I manually removed the barrier=kerb from the main highway nodes.

A kerb is not part of a vehicle highway. Where a kerb-like structure impedes a roadway, other tags (such as traffic_calming or pedestrian) may be a better choice.

Related tags include kerb=lowered.

frodrigo:
This change seem legitimate.
The rule is simple and can be good to be implemented in JOSM MapCSS validation.
This validations are used in Osmose.

cc @Klumbumbus

Klumbumbus:
I'll have a look at this.

frodrigo:
@Klumbumbus still plan to look at this?

Klumbumbus:
yes

Klumbumbus:
Code is ready, I'll commit it to JOSM next week (because we just released a stable version hotfix, in case one more is needed).
As not only the barrier=kerb tag could be wrong but instead the highway=crossing tag, I'd limit the test to "major roads":

````
way[highway =~ /^(bus_guideway|living_street|motorway|motorway_link|pedestrian|primary|primary_link|raceway|residential|road|secondary|secondary_link|service|tertiary|tertiary_link|track|trunk|trunk_link|unclassified)$/] > node { set .is_in_major_road }

node[highway=crossing][barrier=kerb].is_in_major_road {
  throwWarning: tr("{0} together with {1}", "{0.tag}", "{1.tag}");
  suggestAlternative: "kerb=*";
  group: tr("suspicious tag combination");
}
````

frodrigo:
Thank you.

Klumbumbus:
see https://josm.openstreetmap.de/ticket/18507

PR #733, opened 2020-01-11
enable phone numbers in Italy

fansanelli:


frodrigo:
Do you no more want check for IT phone?

fansanelli:
I want to do a change to the regex. It's fine for you?

frodrigo:
Yes. But you can push more commits here.

plugins/Phone.py
```diff
@@ -287,14 +287,15 @@ class father:
...
+            options = {"country": "IT", "phone_code": "39", "phone_len": [6, 11], "phone_len_short": [2, 3, 4], "phone_international": "00", "phone_format": r"^[+]%s[- ]*[03][0-9]+(?:[- ][0-9]+)?(?:(?:[- ][0-9]+)|$)$"}
         class father:
             config = _config()
         p.father = father()
         p.init(None)
 
         for (bad, good) in (
-            (u"800 123", u"+39 800 123"),
+            # Use phone:IT for toll-free numbers
+            # (u"800 123", u"+39 800 123"),
```
frodrigo: toll-free are callable from aboard ? It is not the case in France, so we keep it in local form.

fansanelli:
Toll free (starting with 8) and public utility (starting with 1) numbers are not callable from abroad... So international prefix (+39) shouldn't be there. I thought was fine to mention they should be moved in "phone:IT".
If saw you have a special case for "0" in France... Should I do the same for the "8"? And fix the regex to be "^(?:(?:%s)|8[0-9]+)...."?

frodrigo:
Not sure about `phone:IT` for non abroad number.

In France we have a "0" prefix for calling number in France, except for short number.

fansanelli:
> Not sure about `phone:IT` for non abroad number.
> 
> In France we have a "0" prefix for calling number in France, except for short number.

I see... I hope to have addressed the special case with a check on the country

frodrigo:
It is now good you? ready to merge?

fansanelli:
I hope so! 😃
I'll check results if you can share them after first pass

frodrigo:
Merged. But not yet deployed.

frodrigo:
Please, can you check at the results http://osmose.openstreetmap.fr/en/errors/?country=italy*&item=3092&level=1%2C2%2C3

fansanelli:
I am pretty happy with it... I started fixing!

PR #753, opened 2020-02-07
Update Deprecated check and subtitle string

SilentSpike:
- Wasn't sure about the reason it was previously reading from `&force_cache_20180805` before, but this is successfully extracting expected details from the current page
- Even though subtitle doesn't support markdown, using `` for user consistency, then if it ever does in future it's already set up (plus we can format it in iD)

### DIFF
--- a/plugins/TagFix_Deprecated.py
+++ b/plugins/TagFix_Deprecated.py
@@ -26,67 +26,62 @@ import re
 
 
 class TagFix_Deprecated(Plugin):
-
-    def cleanWiki(self, src):
-        if src is None:
-            return src
-
-        # Remove bold and newlines
-        src = src.replace("'''", "").replace("<br/>", " ").replace("<br />", " ")
-
-        # Remove excess whitespace
-        return " ".join(src.split())
-
     def deprecated_list(self):
         wikiRoot = 'https://wiki.openstreetmap.org/wiki'
-        data = urlread(f'{wikiRoot}/Template:Deprecated_features?action=raw', 1)
-        data = data.split(r'{{Deprecated features/item')
+        data = urlread(wikiRoot + '/Template:Deprecated_features?action=raw', 1)
 
-        dataMult = []
-        for feature in data[1:]:
-            # Eliminate any whitespace around pipe characters (this also eliminates newlines)
-            feature = re.sub(r'\s*\|\s*', '|', feature)
+        # Tidy data up for processing
+        # Eliminate wiki bold formatting
+        data = data.replace("'''", "")
 
-            # Eliminate templates to prevent unexpected pipe characters
-            feature = re.sub(r'{{{lang\|}}}', '', feature, flags=re.I)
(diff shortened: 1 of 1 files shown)


plugins/TagFix_Deprecated.py
```diff
@@ -30,37 +30,54 @@ class TagFix_Deprecated(Plugin):
...
-                dataMult.append((deprecated_key, deprecated_value, deprecated_suggestion))
+            # Eliminate any whitespace around pipe characters (this also eliminates newlines)
+            feature = re.sub(r'\s*\|\s*', '|', feature)
+
+            # Eliminate templates to prevent unexpected pipe characters
+            feature = re.sub(r'{{{lang\|}}}', '', feature, flags=re.I)
+            # Tag template can take one or two params, with trailing | possible
+            feature = re.sub(
+                r'{{Tag\|(.+?)\|?}}',
+                lambda x : f'`{ x[1].replace("|", "=") }`',
```
frodrigo: Why use a fstring, it's not required, is it?

frodrigo:
@jocelynj this PR introduce f-string from pyhton 3.6, are we ready to execute 3.6 code ?

jocelynj:
I think it is too early to switch to py3>3.6, as we still have machine with Debian "oldstable", which is using python 3.5.

SilentSpike:
👍 Cool, I'll update my PRs to not use f-strings

plugins/TagFix_Deprecated.py
```diff
@@ -26,50 +26,62 @@
...
+        # Eliminate any whitespace around pipe characters
+        # This makes reading the template parameters simple
+        data = re.sub(r'\s?\|\s?', '|', data)
+
+        # Eliminate templates to prevent unexpected pipe characters
+        data = re.sub(r'{{{\s?lang\s?\|?\s?}}}', '', data, flags=re.I)
+        # Tag template can take one or two params, with trailing | possible
+        data = re.sub(
+            r'{{Tag\s?\|(.+?)\|?\s?}}',
+            lambda x : '`{}`'.format(x[1].replace("|", "=")),
```
frodrigo: Got this here
```
2020-02-10 17:26:19         lambda x : '`{}`'.format(x[1].replace("|", "=")),
2020-02-10 17:26:19     TypeError: '_sre.SRE_Match' object is not subscriptable
```

SilentSpike: Ah apparently the `__getitem__` method was only added to match objects in 3.6, will update

frodrigo:
Got this on the result:
```
<text value="The tag `leisure=club` is deprecated in favour of {{key" lang="en" />
```

SilentSpike:
🤔 Hmm, can add a line to resolve the `key` template. May also be worth adding a failsafe to just skip over things if `{{` or `}}` is still found in them after all the processing

frodrigo:
Thank

PR #754, opened 2020-02-07
Update untagged and area issue strings

SilentSpike:


frodrigo:
Ok. But waiting for @jocelynj  about f-strings.

SilentSpike:
I did notice the `merge_doc` method doesn't seem to be adding the list of tags where already implemented. Example: https://osmose.openstreetmap.fr/en/error/4d24f718-81ab-fb83-b91f-68997b3e789d

frodrigo:
You are right, it is not working, at least for plugins. https://github.com/osm-fr/osmose-backend/issues/756

frodrigo:
Thank you.

PR #760, opened 2020-02-10
remove false positives for old railway which are now highways

ferdinand0101:
This is mostly relevant for cycleways which follow (or are planned to follow) old railway routes.
as

This change should remove these false positives but I'm unsure if I did it correctly so please check and correct me


frodrigo:
Thank you.

PR #761, opened 2020-02-12
Access mispelling

ferdinand0101:
fixed mispellings of access usually as acces which is probably due to all you french people

### DIFF
--- a/analysers/analyser_osmosis_relation_route_access.py
+++ b/analysers/analyser_osmosis_relation_route_access.py
@@ -74,20 +74,20 @@ class Analyser_Osmosis_Relation_Route_Access(Analyser_Osmosis):
         self.map = {
             'bicycle': {
                 'class': 1,
-                'acces_tag': 'bicycle',
-                'no_acces': "'no', 'false', 'private', 'discouraged'",
+                'access_tag': 'bicycle',
+                'no_access': "'no', 'false', 'private', 'discouraged'",
                 'highway_overide': "'footway', 'pedestrian', 'trunk', 'motorway', 'trunk_link', 'motorway_link'",
                 'highway_overide_access': "'yes', 'permissive', 'true', 'designated', 'shoulder', 'dismount'"},
             'foot': {
                 'class': 2,
-                'acces_tag': 'foot',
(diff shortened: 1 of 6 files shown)


frodrigo:
Good catch. But some are French column name (with only one "s"), should be reverted. `access_recharge_grouped` should be checked in the data source.

frodrigo:
Travis error

```
        for route_type, access in self.map.items():

>           self.run(sql10.format('', '', route_type, access['acces_tag'], access['no_acces'], access['highway_overide'], access['highway_overide_access']), self.callback10(access['class']))

E           KeyError: 'acces_tag'

```

analysers/analyser_osmosis_relation_route_access.py
```diff
@@ -104,5 +104,5 @@ def analyser_osmosis_full(self):
 
     def analyser_osmosis_diff(self):
         for route_type, access in self.map.items():
-            self.run(sql10.format('', 'touched_', route_type, access['acces_tag'], access['no_acces'], access['highway_overide'], access['highway_overide_access']), self.callback10(access['class']))
-            self.run(sql10.format('touched_', 'not_touched_', route_type, access['acces_tag'], access['no_acces'], access['highway_overide'], access['highway_overide_access']), self.callback10(access['class']))
+            self.run(sql10.format('', 'touched_', route_type, access['access_tag'], access['no_access'], access['highway_overide'], access['highway_overide_access']), self.callback10(access['class']))
+            self.run(sql10.format('touched_', 'not_touched_', route_type, access['access_tag'], access['no_access'], access['highway_overide'], access['highway_overide_access']), self.callback10(access['class']))
```
ferdinand0101: The issue should be in this line right ???
But in the error message the names don't match

frodrigo: https://github.com/osm-fr/osmose-backend/pull/761/files/6c2de4442a81670b76858f422cd7a2e73787c2b6#diff-b631ed634ba0b61cf2bcc5121104356fR103

frodrigo:
Thank you.

PR #765, opened 2020-02-15
Update some strings for bad tagging checks

SilentSpike:
Was originally just going to update all the subtitles for these to better show the user which tag they concern. Ended up updating/adding some docs strings while I was at it.

This also cuts down translation a bit by using the same subtitle string for bad tagging checks.

### DIFF
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -57,12 +57,12 @@ class Name_UpperCase(Plugin):
             # first check if the name *might* match
             if self.UpperTitleCase.match(tags[u"name"]) and not self.RomanNumber.match(tags[u"name"]):
                 if self.whitelist is None:
-                    err.append({"class": 803, "text":{"en":tags[u"name"]}})
+                    err.append({"class": 803, "text": T_f("Concerns tag: `{0}`", '='.join(['name', tags['name']])) })
                 else:
                     # Check if we match the whitelist and if so re-try
                     name = " ".join(i for i in tags[u"name"].split() if not i in self.whitelist)
                     if self.UpperTitleCase.match(name) and not self.RomanNumber.match(name):
-                        err.append({"class": 803, "text":{"en":tags[u"name"]}})
+                        err.append({"class": 803, "text": T_f("Concerns tag: `{0}`", '='.join(['name', tags['name']])) })
         return err
 
     def way(self, data, tags, nds):
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -36,7 +36,12 @@ class Name_UpperCase(Plugin):
         self.errors[803] = self.def_class(item = 5010, level = 1, tags = ['name', 'fix:chair'],
             title = T_('Name with uppercase'),
             detail = T_(
-'''Word in capital letters.'''))
+'''This feature is tagged with a name which contains a fully uppercase word (or words).
+ This is not expected for the majority of named features.'''),
+            trap = T_(
+'''While uncommon, it isn\'t impossible for a name to have uppercase words.
+ This is particularly the case for corporate/branded locations as well as acronyms.''')
+        )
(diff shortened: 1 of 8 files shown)


plugins/Name_UpperCase.py
```diff
@@ -36,7 +36,12 @@ def init(self, logger):
         self.errors[803] = self.def_class(item = 5010, level = 1, tags = ['name', 'fix:chair'],
             title = T_('Name with uppercase'),
             detail = T_(
-'''Word in capital letters.'''))
+'''This feature is tagged with a name which contains a fully uppercase word (or words).
+ This is not expected for the majority of named features.'''),
+            trap = T_(
+'''While uncommon, it isn\'t impossible for a name to have uppercase words.
```
frodrigo:
```suggestion
'''While uncommon, it is possible for a name to have uppercase words.
```

plugins/Number.py
```diff
@@ -53,12 +53,12 @@ def node(self, data, tags):
...
                 ):
-                    return {"class": 3091, "subclass": 1, "text": T_(u"Incorrect number \"%s\"", tags[i])}
+                    return {"class": 3091, "subclass": 1, "text": T_f("Concerns tag: `{0}`", '='.join([i, tags[i]])) }
                 elif m and i == "height" and float(m.group(1)) > 500:
-                    return {"class": 3091, "subclass": 2, "text": T_(u"height=%s is really tall, look at ele=*", m.group(1)),
+                    return {"class": 3091, "subclass": 2, "text": T_(u"`height=%s` is really tall, consider changing to `ele=*`", m.group(1)),
                              "fix": {"-": ["height"], "+": {"ele": tags["height"]}} }
                 elif m and i == "maxspeed" and float(m.group(1)) < 5 and not "waterway" in tags:
-                    return {"class": 3091, "subclass": 3, "text": T_(u"%s is really slow", m.group(1))}
+                    return {"class": 3091, "subclass": 3, "text": T_(u"`maxspeed=%s` is really slow", m.group(1))}
```
frodrigo:
```suggestion
                    return {"class": 3091, "subclass": 3, "text": T_f("`maxspeed={0}` is really slow", m.group(1))}
```

plugins/TagFix_BadValue.py
```diff
@@ -28,7 +28,15 @@ class TagFix_BadValue(Plugin):
...
         Plugin.init(self, logger)
         self.errors[3040] = self.def_class(item = 3040, level = 1, tags = ['value', 'fix:chair'],
-            title = T_('Bad tag value'))
+            title = T_('Bad tag value'),
+            detail = T_(
+'''This feature is tagged with a value which does not match the typical format used for tags with distinct values
+ (lowercase alphanumeric characters with no spaces). It's most likely not an expected tagging.'''),
+            fix = T_('Check the value of the tag in question and update the tagging to reflect what this feature is.'),
+            trap = T_(
+'''It\'s possible a mapper was trying to map a feature with no existing agreed upon tagging.
```
frodrigo:
```suggestion
'''It's possible a mapper was trying to map a feature with no existing agreed upon tagging.
```

plugins/Number.py
SilentSpike: I do wonder about these variants (low `maxpeed` and high `height`) as they don't really match with the other issue docs strings and would seem more suitable as a new class of issue ("Impossible values" or something).

frodrigo: Yes. It may be switched to a new or an other item.

But it have no link with this change suggestion.

SilentSpike: 👍

frodrigo:
Thank

jocelynj:
It looks like it is breaking a plugin test:

https://travis-ci.com/osm-fr/osmose-backend/jobs/288140623
```
>               err.append({"class": 3050, "subclass": stablehash64(k), "text": T_f("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
E               TypeError: sequence item 1: expected str instance, int found
```

frodrigo:
Test fixed with #769

PR #774, opened 2020-02-19
Fuel IT v3

fansanelli:
Hello there,

we had some false positive, due to some "stop word"... This fixes (hopefully) most of warnings and also add the "unknown" check from another Analyzer (as per Frédéric suggestion).
This patch is untested (except for the function part), but hopefully you can comment on the approach.
Please also mind that previous version didn't trim the strings and this generated some error here and there too...

Many thanks
Best
Francesco


frodrigo:
Thank. Not deployed, yet.

frodrigo:
Deployed, update in progress.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?country=italy*&item=8200,8201,8202

http://osmose.openstreetmap.fr/en/errors/graph.png?country=italy*&item=8200,8201,8202

Please can you review results.

I think we can swtich operator tag to secondary tags. Secondary tag does not raise issue on map when not the same: mapping2 in place of mapping1 https://github.com/fansanelli/osmose-backend/blob/master/analysers/analyser_merge_fuel_IT.py#L54

fansanelli:
I was checking them just before you write me... I spotted a little nit: "di" should became "Di" when it's the first word.
I'm not sure about hiding the issues... I'll also ask other mappers and let you know

PR #778, opened 2020-02-21
Differentiate extreme numerical value checks from invalid numerical value checks

SilentSpike:
This way the docs strings are more appropriate for these issue types (and I could see further issues of this nature being added to this class).

### DIFF
--- a/plugins/Number.py
+++ b/plugins/Number.py
@@ -29,10 +29,10 @@ class Number(Plugin):
         self.errors[3091] = self.def_class(item = 3091, level = 2, tags = ['value', 'fix:chair'],
             title = T_('Numerical value'),
             detail = T_(
-'''The tag expects a numeric value. Decimals are set with one point and
-not a comma.'''),
-            fix = T_(
-'''Change to numeric value and/or the comma to point.'''))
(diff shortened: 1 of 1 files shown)


frodrigo:
May be have tow class, to have spesific and more explicit doc?

frodrigo:
Note for myself, add item 3092.

SilentSpike:
Generic means less work for translators and potential to add more cases (maxspeed higher than 160 - this could even be varied by country if highest speed limits are known, street `incline` higher than the world record, etc.)

frodrigo:
While there is more doc, I think we lose some thing very simple and understandable : "is really slow".

frodrigo:
Pick only a part of the PR.

SilentSpike:
👍 Apologies, have been busy with other projects and didn't get a chance to update this.

One suggestion I might make for the subtitle strings is to always put key and value in using formatting (this way the strings are more generic and could be used elsewhere too).

So
```python
"`{0}` is really slow", 'maxspeed=' + m.group(1))
```

frodrigo:
You are right. Changed.

PR #781, opened 2020-02-24
Fuel IT v4

fansanelli:
Hello,

I investigated many more cases and spotted some other false positive.
Please let me know it's ok for you

Francesco


fansanelli:
please wait before merging.. I can avoid the extra upper()

PR #799, opened 2020-03-16
Street number in Rennes : remove space between number and suffix

PanierAvide:
Hello,

This fix is intended to remove space between street number and suffix in Rennes addresses merge analyser. It is intended to conform widely-used rule of not having space.

Regards.


frodrigo:
Thank.

PR #817, opened 2020-03-26
reg name is for regional name

ferdinand0101:
it gets suggested to be changed to ref name


frodrigo:
Usually the tag a re added by pair. Witch is the other one?

ferdinand0101:
ref_name

frodrigo:
Thank you.

PR #829, opened 2020-04-11
Pharmacy-IT and Parapharmacy-IT fix for start_date

fansanelli:
Hello,

start_date is not the date of "opening", but initial date for the source.
Hope is good for all.

Thanks
Francesco


frodrigo:
So, is it worth keeping?

Note, we also enable this
http://osmose.openstreetmap.fr/en/errors/?item=8212
with lot of http://osmose.openstreetmap.fr/fr/error/8a342e46-b7b4-f34e-7023-e07b04962253 where only create tags, witch is bad

|k|v||
-- | -- | --
source | Ministero della Salute |  
start_date | 2005-01-01 |  

Maybe also move source to mapping2 ?

fansanelli:
Yeah... that was a problem, can you hide them for now?
Btw source is already in static2... Isn't fine?

frodrigo:
You are right, the change on start_date will fix it.

frodrigo:
Deployed

PR #839, opened 2020-04-14
Change millesime of opendata to 03/2020

zorglubu:
I am really not sure on how to propose this change.
Any feedback is welcome :-)


frodrigo:
No. The date is correct. We stay on 2019 millesime, as new is invalid.

zorglubu:
ah OK, sorry.
That' true, I have removed some lines lines in order to work properly with it :-)

frodrigo:
But if you know how to fix it may be an interesting update.

zorglubu:
To have the correct file, I just downloaded from https://www.data.gouv.fr/fr/datasets/finess-extraction-du-fichier-des-etablissements/
the file etalab-cs1100502-stock-20200304-0419.csv

It's in ISO-8859-01. separator ";". Do you want quotes around strings ?
Is it OK to upload it here (33 Mo) ?

frodrigo:
No. The problem is there is match in location and other data.
cc @pyrog

zorglubu:
Do you mean that the location is incorrect in the file "établissements géolocalisés" 
(etalab-cs1100507-stock-20200304-0416.csv) ?

PR #849, opened 2020-04-22
Add tag ref:FR:SIRET

pyrog:


vinber:
3 tags / ref:FR:SIRET phone and fax
For phone and fax, i don't know if it's automatic but we need to add +33.

frodrigo:
phone and fax are already available.

For siret, we do not need +33.

vinber:
ok, but next line concern only phone with 10 numbers and a 0 ? 
In our case, in "finess" it seem to be only phone number with 9 numbers but i not sure to understand "code"

frodrigo:
this PR is not about phone. We already have phone implement.

vinber:
Sorry I misread, misunderstood what was written in the code.

pyrog:
> In our case, in "finess" it seem to be only phone number with 9 numbers but i not sure to understand "code"

Oups, j'ai fait un copier/coller trop rapide. Rien à voir effectivement avec les n° de téléphones.

frodrigo:
Online.

PR #847, opened 2020-04-19
Plugin to find wayside_crosses without material

morray:
Plugin to find wayside_crosses without material restricted to DE, AT, and CH for the start. Code inspired by plugins/Administrative_INSEE_Name.py

PR #853, opened 2020-04-27
removed unused doc, changed referenced class to 303242

morray:
handling #issuecomment-619800319 from #847 

removed doc variable: though this is a global definition needed as it was in all examples
changed the class to 300242


frodrigo:
Thank you.

Issue #752, opened 2020-02-07
Check for pipes in destination tags


Discostu36:
The following tags should be checked if the value contains a pipe (`|`):

- `destination`
- `destination:forward`
- `destination:backward`
- `destination:ref`
- `destination:ref:forward`
- `destination:ref:backward`
- `destination:symbol`
- `destination:symbol:forward`
- `destination:symbol:backward`
- `destination:ref:to`
- `destination:ref:to:forward`
- `destination:ref:to:backward`
- `destination:int_ref`
- `destination:int_ref:forward`
- `destination:int_ref:backward`
- `destination:country`
- `destination:country:forward`
- `destination:country:backward`
- `destination:street`
- `destination:street:forward`
- `destination:street:backward`

Pipes should only be used for destination lanes, not for destination.

**Example:**

Wrong:
`destination=Paris|Bordeaux`

Correct 1:
`destination:lanes=Paris|Bordeaux`

Correct 2:
`destination=Paris;Bordeaux`

klenje:
Fixed with https://github.com/osm-fr/osmose-backend/pull/855

klenje:
http://osmose.openstreetmap.fr/fr/errors/?source=24767&item=3160&class=316010
I see quite some false positives like:

destination:symbol:lanes
destination:ref:lanes
destination:colour:lanes

We should probably exclude any key containing lanes

frodrigo:
Full result at
http://osmose.openstreetmap.fr/en/errors/?item=3160&class=316010

@klenje, thank you, please make a new PR

klenje:
@frodrigo deployment is done automaticaly? how can I check the version? Because I don't see the changes yet

frodrigo:
No is manual. And it is not yet deployed.
cc @jocelynj

klenje:
All right, thanks! No pressure :)

frodrigo:
Deployed, update in progress.

klenje:
Looks better, but I think the proposed fix is wrong, I will do another PR later

frodrigo:
Deployed, update in progress.

klenje:
looks good to me. No false positives and correct fix suggested

frodrigo:
Thank you.

PR #855, opened 2020-05-01
Check for pipes in destination tags #752

klenje:
Hi, first try with osmose code, I tested it with Docker but I may have missed something :)
Any convention for the error ID? I picked a random unused value

### DIFF
--- a/plugins/TagFix_Destination.py
+++ b/plugins/TagFix_Destination.py
@@ -1,0 +1,94 @@
+#-*- coding: utf-8 -*-
+import modules.mapcss_lib as mapcss
+import regex as re # noqa
+
+from plugins.Plugin import with_options # noqa
+from plugins.PluginMapCSS import PluginMapCSS
+
+
+class TagFix_Destination(PluginMapCSS):
+
+
+
+    def init(self, logger):
+        super().init(logger)
+        tags = capture_tags = {} # noqa
+        self.errors[316010] = self.def_class(item = 3160, level = 3, tags = mapcss.list_(u'tag', u'highway', u'waterway'), title = mapcss.tr(u'Pipe characters should not be used in destination tag, only in destination:lanes'))
+
+        self.re_262d3d80 = re.compile(r'\|')
+        self.re_49b44b3d = re.compile(r'^destination:lanes')
+        self.re_53d7e349 = re.compile(r'^destination:')
+
+
+    def way(self, data, tags, nds):
+        capture_tags = {}
+        keys = tags.keys()
+        err = []
+
+
+        # way[highway][destination][destination*="|"]
+        # way[highway][/^destination:/][!/^destination:lanes/][/^destination:/=~/\|/]
+        # way[waterway][destination][destination*="|"]
+        # way[waterway][/^destination:/][!/^destination:lanes/][/^destination:/=~/\|/]
+        if (u'destination' in keys and u'highway' in keys) or (u'destination' in keys and u'waterway' in keys) or (u'highway' in keys) or (u'waterway' in keys):
+            match = False
+            if not match:
+                capture_tags = {}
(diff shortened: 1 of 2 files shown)


frodrigo:
Simple analyser like this should be done in plugin and better, using mapcss.

Like this eg https://github.com/osm-fr/osmose-backend/blob/master/plugins/Construction2.validator.mapcss

Mapcss compiler https://github.com/osm-fr/osmose-backend/tree/master/mapcss

klenje:
Thanks, will try that. Question, any hint on how to test such plugins since you don' t define a new analyzer?

frodrigo:
All plugins run under the `sax` analyser.
The mapcss code can also be tested with JOSM.

klenje:
Second try based on mapcss and starting from Didier's hints

plugins/Destination.validator.mapcss
```diff
@@ -0,0 +1,44 @@
+/*#########################################################################
+##                                                                       ##
+## Copyrights Frédéric Rodrigo 2018                                      ##
```
frodrigo, line 3: Our own name, here.

plugins/Destination.validator.mapcss
```diff
@@ -0,0 +1,44 @@
...
+/* ways with "|" in destination tag excluding destination:lanes:*  */
+way[highway][destination][destination*="|"],
+way[highway][/^destination:/][!/^destination:lanes/][/^destination:/=~/\|/],
+way[waterway][destination][destination*="|"],
+way[waterway][/^destination:/][!/^destination:lanes/][/^destination:/=~/\|/]
+{
+    -osmoseItemClassLevel: "3320/33200:0/3";
+    throwError: tr("{0} contains a pipe character", "{1.tag}");
+    group: tr("Pipe characters should not be used in destination tag, only in destination:lanes");
+    suggestAlternative: tr("In case of multiple values, use instead a semicolon to separate values");
```
frodrigo, line 43: Can you add some tests with `assertMatch:` and `assertNoMatch`.

You can also add `fixAdd` with `;`.

klenje: Thanks, adding those, only tested on a country extract.
I am adding a fixRemove since this is about a wrong character

plugins/Destination.validator.mapcss
klenje: yep

plugins/Destination.validator.mapcss
frodrigo: It is `fixChange`, since you want to suggest change on the value of a tag.

klenje:
For sax failures, I think I will need to update the tests/ folder, any hint on what is wrong with lint test failures instead? I only see 2 warnings from generated code but I can' t reproduce locally

frodrigo:
About the lint fail, rebase and regenerate the python. It may fix it. (We just add the lint check yesterday).

Fox sax fails also try the rebase first.

frodrigo:
Can you rename it TagFix_Destination, please.
Better pick the item 3160 with class id 316010.

plugins/Destination.validator.mapcss
klenje: All right, I wanted to use that but I didn't see it in the converter (https://github.com/osm-fr/osmose-backend/blob/47a461a2242efb8b80999ed21566c3d463d5a86a/mapcss/mapcss2osmose.py#L221) nor used in the existing files, I will give it a try

frodrigo: It is fixAdd, like this: https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_MisspelledWordByRegex_Lang_fa.validator.mapcss#L39

klenje: done, thanks

klenje:
> Can you rename it TagFix_Destination, please.
> Better pick the item 3160 with class id 316010.
Done
Lint test now ok after rebasing, sax still fails

frodrigo:
I check the result of the test. It is ok.

frodrigo:
Thank you.

klenje:
Thanks a lot for the help!

PR #874, opened 2020-05-18
Add a sax plugin to add missing brand tags

nlehuby:
This plugin uses the OSM frequent names extracted by the [name-suggestion-index](http://nsi.guide/) (nsi) to detect brand candidates in OSM objects.

If they correspond to known brands in the nsi, it then suggests to add some tags (mostly `brand` and `brand:wikidata`).

Screenshot_2020-05-16 Osmose - Carte

### DIFF
--- a/plugins/TagFix_Brand.py
+++ b/plugins/TagFix_Brand.py
@@ -1,0 +1,106 @@
+# -*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Noémie Lehuby 2020                                         ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from plugins.Plugin import TestPluginCommon
+from plugins.Plugin import Plugin
+from modules.downloader import urlread
+import json
+
+#TODO : handle country codes
+
+class TagFix_Brand(Plugin):
+
+    def init(self, logger):
+        Plugin.init(self, logger)
+        doc = dict(
+                detail = T_(
+'''This object has a very common name that is likely to be a brand.'''),
+                fix = T_(
+'''If this is indeed a brand, add `brand` and `brand:wikidata` tags.
+If not, report it to the [name-suggestion-index project](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md)
+which is used here to link frequent names to brands'''))
+
+        self.errors[31301] = self.def_class(item=3130, level=2, tags=['brand', 'fix:chair'],
+                                            title=T_(
+                                                "This name is very common, is it a brand ?"),
+                                            resource='https://nsi.guide/', **doc)
+
+        self.frequent_names_from_nsi = self._get_frequent_names()
+        self.brands_from_nsi = self._get_brands()
+
+    def _get_frequent_names(self):
+        nsi_url_for_names = "https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/names_keep.json"
+        json_str = urlread(nsi_url_for_names, 30)
+        results = json.loads(json_str)
+        return [elem.lower() for elem in results.keys()]
+
+    def _get_brands(self):
+        nsi_url_for_brands = "https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/brands.json"
+        json_str = urlread(nsi_url_for_brands, 30)
+        results = json.loads(json_str)
+        additional_brands = {}
+        for brand_nsi_name, brand in results["brands"].items():
+            if "matchTags" in brand:
+                for additional_tag in brand["matchTags"]:
+                    nsi_key = "{}|{}".format(additional_tag, brand_nsi_name.split("|")[1])
+                    additional_brands[nsi_key.lower()]=brand
+            if "matchNames" in brand:
+                for additional_name in brand["matchNames"]:
+                    nsi_key = "{}|{}".format(brand_nsi_name.split("|")[0], additional_name)
+                    additional_brands[nsi_key.lower()]=brand
+            additional_brands[brand_nsi_name.lower()]=brand
+        return additional_brands
+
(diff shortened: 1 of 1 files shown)


frodrigo:
Nice!

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,109 @@
...
+            detail=T_(
+                '''This object has a very common name that probably corresponds to a brand name.'''),
+            fix=T_(
+                '''If this is indeed a brand, add `brand` and `brand:wikidata` tags.
+If not, see if you can improve the [name-suggestion-index project](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) which is used to link frequent names to brands and their tags'''))
+
+        self.errors[31301] = self.def_class(item=3130, level=2, tags=['brand', 'fix:chair'],
+                                            title=T_(
+                                                "This name is very common, is it a brand ?"),
+                                            resource='https://nsi.guide/', **doc)
```
frodrigo, line 42: No need to factorize the doc. You can put it here directly.

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,109 @@
...
+        return [elem.lower() for elem in results.keys()]
+
+    def _get_brands(self):
+        nsi_url_for_brands = "https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/brands.json"
+        json_str = urlread(nsi_url_for_brands, 30)
+        results = json.loads(json_str)
+        additional_brands = {}
+        for brand_nsi_name, brand in results["brands"].items():
+            if "countryCodes" in brand and self.country_code.lower() not in brand["countryCodes"]:
+                #print("ignoring brand " + brand_nsi_name)
```
frodrigo, line 61:
```suggestion
```

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self.country_code = self.father.config.options.get("country").split("-")[0]
+        self.frequent_names_from_nsi = self._get_frequent_names()
+        self.brands_from_nsi = self._get_brands()
+
+    def _get_frequent_names(self):
+        nsi_url_for_names = "https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/names_keep.json"
+        json_str = urlread(nsi_url_for_names, 30)
+        results = json.loads(json_str)
+        return [elem.lower() for elem in results.keys()]
```
frodrigo, line 52: Looking over a large array maybe slow, Maybe a set?

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,109 @@
...
+            additional_brands[brand_nsi_name.lower()] = brand
+        return additional_brands
+
+    def node(self, data, tags):
+        if "name" in tags and not "brand" in tags:
+            for main_key in ["shop", "amenity"]:
+                if main_key in tags:
+                    nsi_key = "{}/{}|{}".format(main_key, tags[main_key], tags["name"]).lower()
+                    if nsi_key in self.frequent_names_from_nsi:
+                        if nsi_key in self.brands_from_nsi.keys():
```
frodrigo, line 81: look into the keys is the default.
```suggestion
                        if nsi_key in self.brands_from_nsi:
```

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,109 @@
...
+                            brands_tags = self.brands_from_nsi[nsi_key]["tags"]
+                            tags_to_add = {}
+                            for tag in brands_tags:
+                                if not tags.get(tag):
+                                    tags_to_add[tag] = brands_tags[tag]
+                            return {"class": 31301, "subclass": 0,
+                                    "fix": {"+": tags_to_add}}
```
frodrigo, line 88: No sure here. If the tag already exist you need to use '~' in place of '+'.

nlehuby: It does not already exist (`if not tags.get(tag)`) : it only add tags that are not present yet.

Do you want to change that and replace all existing keys and values by the ones in the nsi ?  It seems a bit excessive to me but this is arguable.

frodrigo: Yes no replace. Ok.

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,106 @@
...
+        return self.node(data, tags)
+
+
+###########################################################################
+
+
+class Test(TestPluginCommon):
+    def test(self):
+        a = TagFix_Brand(None)
+        a.init(None)
```
frodrigo, line 103: Define a country in the test, since you use it.

```suggestion
        class _config:                                                                                             
            options = {"language": "fr"}                                                                           
        class father:                                                                                              
            config = _config()                                                                                     
        a.father = father()   
        a.init(None)
```

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,111 @@
...
+        Plugin.init(self, logger)
+        self.errors[31301] = self.def_class(item=3130, level=2, tags=['brand', 'fix:chair'],
+                                            title=T_(
+                                                "This name is very common, is it a brand ?"),
+                                            resource="https://nsi.guide/",
+                                            detail=T_(
+                                                "This object has a very common name that probably corresponds to a brand name."),
+                                            fix=T_('''If this is indeed a brand, add `brand` and `brand:wikidata` tags.
+If not, see if you can improve the [name-suggestion-index project](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) which is used to link frequent names to brands and their tags''')
+                                            )
```
frodrigo, line 40:
```suggestion
            title=T_("This name is very common, is it a brand?"),
            detail=T_(
'''This object has a very common name that probably corresponds to a brand name.'''),
            fix=T_(
'''If this is indeed a brand, add `brand` and `brand:wikidata` tags.
If not, see if you can improve the [name-suggestion-index project](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) which is used to link frequent names to brands and their tags.'''),
            resource="https://nsi.guide/")
```

plugins/TagFix_Brand.py
```diff
@@ -0,0 +1,110 @@
...
+from plugins.Plugin import Plugin
+from modules.downloader import urlread
+import json
+
+
+class TagFix_Brand(Plugin):
+
+    def init(self, logger):
+        Plugin.init(self, logger)
+        self.errors[31301] = self.def_class(item=3130, level=2, tags=['brand', 'fix:chair'],
```
frodrigo, line 32:
```suggestion
        self.errors[31301] = self.def_class(item=3130, level=3, tags=['brand', 'fix:chair'],
```
Since is only improvement and not fixing the data, I think you could move it to level 3?

nlehuby: sure

frodrigo:
It is now ok for you? Ready to merge?

nlehuby:
yes, sounds good to me

frodrigo:
Thank you.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?item=3130&useDevItem=all

frodrigo:
> This name is very common, is it a brand?

I think the title is not clear. Maybe: "The name suggest it could be a branded shop". I think a more detailed `description` doc field should be added.

PR #878, opened 2020-05-21
FR, datatourisme - add postal code

frodrigo:


analysers/analyser_merge_tourism_offices_FR.py
```diff
@@ -26,7 +26,7 @@
 class Analyser_Merge_Datatourisme_tourism_office_FR(Analyser_Merge):
     def __init__(self, config, logger = None):
         Analyser_Merge.__init__(self, config, logger)
-        self.def_class_missing_official(item = 8420, id = 310, level = 3, tags = ['merge'],
+        self.missing_official = self.def_class(item = 8420, id = 310, level = 3, tags = ['merge'],
```
frodrigo, line 29: This should not be changed. Its the new way to define the class on merge analyser to inherit common doc.

frodrigo:
We should also add self.def_class_possible_merge

Issue #875, opened 2020-05-19
FR, merge datatourisme


frodrigo:
cc @nlehuby @vinber 
Issue to improve datatourisme merge

First results with tourism_office
http://osmose.openstreetmap.fr/en/errors/?item=8420&useDevItem=all

frodrigo:
We should add the postal address in the subtitle.

nlehuby:
I've started a branch: https://github.com/osm-fr/osmose-backend/compare/tourism_office

frodrigo:
From place from data.gouv, 202062 lines
Categories stats: https://pastebin.com/QtWqu7D0

All have PlaceOfInterest and PointOfInterest.

I think we can ignore all generic Accommodation and Lodging.

Low frequency contains interesting thing... but there have low frequency.

Camping or Church can be the next step as they can be found on imagery.

frodrigo:
I make the analyser generic.
https://github.com/osm-fr/osmose-backend/pull/883

nlehuby:
Can we have something like [this](https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_building_shapes.py#L75) and let the user choose the kind of amenity/shop when we have several ones in the source dataset ?

![image](https://user-images.githubusercontent.com/919962/82749260-3e174380-9da8-11ea-90dd-963ba63d1581.png)

frodrigo:
Multiple fixes are not implemented in merge analyser. Should require to have multiple `Generate` parts. On Sirene is just done like this https://github.com/osm-fr/osmose-backend/blob/master/merge_data/shop_FR.mapping.json#L17-L18

frodrigo:
Merge Restaurant, Camping and Church.
Remove pervious Restaurant for Aquitaine and Saône-et-Loire

frodrigo:
Also disable merge_tourism_FR_aquitaine_information.

We have analyser_merge_tourism_FR_aquitaine_camp_caravan.py but with:
- tourism": "caravan_site
- tourism": "camp_site

Can we have both from Datatourisme?

frodrigo:
Camping
http://osmose.openstreetmap.fr/fr/errors/?item=8140&class=330

Church, few, probably only touristic ones (hidden item)
http://osmose.openstreetmap.fr/fr/errors/?item=8101&class=320&useDevItem=all

Restaurant
http://osmose.openstreetmap.fr/fr/errors/?item=8240&class=340

[...]

PR #883, opened 2020-05-23
Switch analyser_merge_tourism_offices_FR to dynamic merge #875

frodrigo:


vinber:
if i understand, now the main goal is to work on tourism_FR.mapping.json

Dynamic merge it's when a mapping.json file is required ?

frodrigo:
> if i understand, now the main goal is to work on tourism_FR.mapping.json

Yes. But also require to export the data into the source query.

> Dynamic merge it's when a mapping.json file is required ?

The json mapping is just a way to solve it. Dynamic is when a data source should require different mappings / queries over OSM data.

PR #889, opened 2020-05-26
brand - improve doc

nlehuby:


frodrigo:
Not sure.
My issue is about "the shop is one of the brand or not?", not about the name.

I worrying about contributor answer the question,: "the name of the shop is a brand?"

nlehuby:
I don't understand: you want to ask "is that optician part of the Optic 2000 brand ?" to an optician with Optic 2000 name ?
if it has the main tag (amenity =* or shop=*) and a name that is a brand name, it is very likely to be part of the brand :thinking:

SilentSpike:
I don't think these should be tagged `fix:chair`, surely `fix:survey`?

frodrigo:
What about?

> The name look like a brand, is this shop part of the brand network?

frodrigo:
OK, let's merge this.

PR #898, opened 2020-06-07
France - Schools - Add tag when public school

Codain:
When a school is explicitely 'public' then fill operator:type=public
See #897


Codain:
Note: Pull request not tested.

frodrigo:
Look good.

PR #902, opened 2020-06-12
aeroway and indoor mapping

didier2020:
add aeroway for way (not nodes)
fix false positive for indoor mapping


frodrigo:
OK. Thank.

PR #904, opened 2020-06-12
Building indoor

didier2020:
fix false positive for indoor mapping


frodrigo:
min_height is more linked to 3D than indoor.

didier2020:
intersections are "not true" when tagging indoor or 3D

frodrigo:
The point is not about ignoring the indoor (I am pro), but the way you detect it.

didier2020:
building don't overlaps :
http://osmose.openstreetmap.fr/fr/error/cf6626a2-e347-8474-6fa2-586a64170192

frodrigo:
I am telling min_height is not directly liked to inddor.

didier2020:
it's ok.

didier2020:
analysis is ok, if there are errors it comes from the tags used.
so i close the branck

PR #911, opened 2020-06-19
geojson Point Linestring

didier2020:
the point is located at half the number of points on the line. it's not in the middle but on the linestring.
there will be an exception in the code if the geojson is multilinestring or ...

### DIFF
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -537,10 +537,9 @@ class GeoJSON(Parser):
                         values.append(row['geometry']['coordinates'][npt][0])
                         values.append(row['geometry']['coordinates'][npt][1])
                     else:
-                        nli = len(row['geometry']['coordinates'])//2
-                        npt = len(row['geometry']['coordinates'][npt])//2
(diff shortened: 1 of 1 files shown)


analysers/Analyser_Merge.py
```diff
@@ -522,16 +522,28 @@ def header(self):
...
+                if row['geometry']['type'] in ('Point', 'MultiPoint', 'LineString', 'MultiLineString'):
+                    if row['geometry']['type'] == 'Point':
+                        values.append(row['geometry']['coordinates'][0])
+                        values.append(row['geometry']['coordinates'][1])
+                    elif row['geometry']['type'] in ('MultiPoint', 'LineString'):
+                        npt = len(row['geometry']['coordinates'])//2
+                        values.append(row['geometry']['coordinates'][npt][0])
+                        values.append(row['geometry']['coordinates'][npt][1])
+                    else:
+                        nli = len(row['geometry']['coordinates'])//2
```
frodrigo, line 540: You can just take the first

frodrigo:
Thank

didier2020:
you're welcome

Issue #923, opened 2020-07-05
merge_public_equipment_FR_nantes_toilets fails


frodrigo:
Error on merge_public_equipment_FR_nantes_toilets
```
2020-07-05 08:36:42   error: error on analyse merge_public_equipment_FR_nantes_toilets...
2020-07-05 08:36:42     Traceback (most recent call last):
2020-07-05 08:36:42       File "./osmose_run.py", line 248, in execc
2020-07-05 08:36:42         analyser_obj.analyser()
2020-07-05 08:36:42       File "/data/project/osmose/backend/analysers/Analyser_Osmosis.py", line 193, in analyser
2020-07-05 08:36:42         self.analyser_osmosis_common()
2020-07-05 08:36:42       File "/data/project/osmose/backend/analysers/Analyser_Merge.py", line 963, in analyser_osmosis_common
2020-07-05 08:36:42         table = self.load.run(self, self.parser, self.mapping, self.config.db_user, self.__class__.__name__.lower()[15:], self.analyser_version())
2020-07-05 08:36:42       File "/data/project/osmose/backend/analysers/Analyser_Merge.py", line 670, in run
2020-07-05 08:36:42         parser.import_(table, self.srid, osmosis)
2020-07-05 08:36:42       File "/data/project/osmose/backend/analysers/Analyser_Merge.py", line 497, in import_
2020-07-05 08:36:42         list(map(removequotesjson, row.values())))
2020-07-05 08:36:42       File "/usr/lib/python3/dist-packages/psycopg2/extras.py", line 141, in execute
2020-07-05 08:36:42         return super(DictCursor, self).execute(query, vars)
2020-07-05 08:36:42     psycopg2.ProgrammingError: column "complement_type" of relation "public_equipment_fr_nantes_toilets" does not exist
2020-07-05 08:36:42     LINE 1: ...e_pmr", "adresse_2", "adresse_1", "motif", "nom", "complemen...
2020-07-05 08:36:42    
```

cc @PanierAvide

PR #924, opened 2020-07-06
Fix FR toilets Nantes analyser

PanierAvide:
Fix for #923 :
- Check all rows instead of first one for columns names (as it's JSON, structure can change through features)
- Update names of columns based on datasource changes


frodrigo:
Great, thank you.

PR #926, opened 2020-07-10
public transport - add tests about stop_position

### LINKED ISSUE #900: proposal analysis on public_transport=stop_position
Good morning, sir,

The creation of a public transit system is long and tedious. In my spare time, I have been slowly progressing on the Narbonne network for 3 years now.
I had put a lot of place where the bus stops as an isolated node when they must be placed on the way. I take care of the correction but maybe this kind of mistake is frequent elsewhere?

Osmose does raise an anomaly if highway=bus_stop is on an isolated node (nickel) but as soon as the tag public_transport=stop_position exists, there is no more error. Yet this should be specified. Passenger waiting places can be on an isolated node but not the location where the bus stops: https://wiki.openstreetmap.org/wiki/FR:Key:public_transport

For the stop "Cave Cooperative", the node https://www.openstreetmap.org/node/2466864152 is well on the way and it has the tag "highway=bus_stop". I was expecting to get the fix-edit from Osmose proposes the tag "public_transport=stop_position" instead of "public_transport=platform". 
Ideally osmosis should say "this belongs to the road,
so fix with stop_postion or disconnect from the road for a platform."

For the "La Ricarde" stop, the https://www.openstreetmap.org/node/4498238410 node is isolated but carries tags that need to be attached to the path (stop_position and bus_stop). I was expecting Osmose to ask for it to be reattached OR for the tags to be modified (platform) so that it corresponds to an isolated node. For the moment Osmose doesn't report any error on "public_transport=stop_position" on isolated nodes.
ideally osmose should say "attach the stop_position to the
if it is the location of the vehicle or correct in the platform.
if that's where the vehicle's waiting.

This is a summary of several exchanges on the openstreetmap talk-fr list

Sincerely

nlehuby:
* bus stop part of a way, without `public_transport` = `stop_position`
* `public_transport` = `stop_position` not part of a way
* `platform` role in a relation but part of a way
* `stop` role in a ptv2 relation, but not part of a way

Fix #900

### DIFF
--- a/analysers/analyser_osmosis_relation_public_transport.py
+++ b/analysers/analyser_osmosis_relation_public_transport.py
@@ -117,6 +117,7 @@ SELECT DISTINCT ON (relations.id, ways.id)
   relations.id,
   relation_members.member_type,
   ways.id AS mid,
+  relation_members.member_role as mrole,
   ST_Transform(ways.linestring, {0}) AS geom
 FROM
   relations
@@ -138,6 +139,7 @@ SELECT DISTINCT ON (relations.id, nodes.id)
   relations.id,
   relation_members.member_type,
   nodes.id AS mid,
+  relation_members.member_role as mrole,
   ST_Transform(nodes.geom, {0}) AS geom
 FROM
   relations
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank, but can you make the Travis CI lint green, please.

frodrigo:
Thank you

maro-21:
Please enable translations for it :)

frodrigo:
The translation is enable in the code. But the update of the translation loop is manual and update time to time. And it takes two tow manual updates to be available on the Osmose.

Issue #823, opened 2020-03-31
junction + area:highway


maro-21:
I would like to ask you if tagging `area:highway`+`junction` really incorrect is?
Because Wiki doesn't mention anything if it can be used or not with `area:highway`.
https://wiki.openstreetmap.org/wiki/Key:junction

I would like to know if it's wrong tagging and I should correct it somehow or is it false positive.

`junction=*` is used 702,777 times and in combination with `area:highway=*` 32,090 times.
https://taginfo.openstreetmap.org/keys/?key=junction#values

Item and class number:
- [x] 20800 "Tag highway missing on junction" and
- [x] 9001001 "junction without highway".

Edit:
- [ ] update MapCSS for 9001001

frodrigo:
This rule come from JOSM validator combinations.mapcss

cc @Klumbumbus

Klumbumbus:
>junction=* is used 702,777 times and in combination with area:highway=* 32,090 times.

The number of concerned objects is not so high, because junction=yes is already excluded from the test. These around 1100 ways are concerned: https://overpass-turbo.eu/s/S9t so aside from some single cases it only affects poland and one part in germany. Most of these objects are tagged junction=y_junction which is not even documented in the wiki.

However I think we can exclude area:highway=* from the test anyway.

Klumbumbus:
Fixed in JOSM in https://josm.openstreetmap.de/changeset/16233.

frodrigo:
Thank you.
Will be available in next MapCSS update.

maro-21:
I would like to point out that one feature with
```
area:highway = tertiary
junction = y_junction
```
generates two issues:
- `2080, 20800: Tag highway missing on junction` and it's on level=1
- and  `9001, 9001001: junction without highway` and it's on level=3

I assume it has been corrected only for 9001001.

maro-21:
Could you reopen this issue? Because it hasn't been resolved.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?item=2080&class=20800

maro-21:
Was there already "MapCSS update"? Because I still can see that error from JOSM 9001001.

frodrigo:
The last MapCSS was update on 2020-03-30 , so no yet.

But the Osmose plugin part of this issue is fixed.

frodrigo:
MapCSS update in progress

maro-21:
> MapCSS update in progress

[...]

PR #928, opened 2020-07-20
Fix boolean expression in 20800

gileri:
If only `highway=*` or `area:highway=*` is used (not both), the 20800 class error is triggered (all roundabouts for example).

This PR requires that both `highway=*` and `area:highway=*` are unset to trigger the error.

See also :
#823 

https://github.com/osm-fr/osmose-backend/commit/3126e1045d36206bcef9c2b13324622da40e35ab#diff-9e304dd3fdef0c5cbd496492c166a748R184


frodrigo:
Thank.
I like when other fix my mistake ;).

gileri:
Haha no problem, I was able to spot it quickly thanks to the code organisation and running tests was a breeze !

frodrigo:
@jocelynj please can we redeploy to have this fix?

PR #937, opened 2020-08-01
Do not warn for wrong layer for culverts. Fixes #934

### LINKED ISSUE #934: Culverts falsely marked
Hello Osmose developers,

Today I noticed a small issue in Osmose. I looked at a culvert with the tags tunnel=culvert, waterway=ditch, layer=-1. Osmose said that waterways are not supposed to have a layer tag, but according to [the wiki page on culverts](https://wiki.openstreetmap.org/wiki/Tag:tunnel%3Dculvert) a layer tag should be correct in this case. 

Kind regards,

Casper

P.s. My apologies if I've overlooked some information that could be useful to you. I'm new around here ;)

danfos:
I saw #934 and remembered seeing similar problems for natural=water + tunnel=culvert for which I had already a patch ready.

The PR combines a fix for #934 (tunnel=culvert, waterway=ditch, layer=-1) and natural=water + tunnel=culvert.

The latest version did still trigger [a failure](https://travis-ci.com/github/danfos/osmose-backend/jobs/367541646) but that seems to have nothing to do with this patch.

### DIFF
--- a/plugins/TagFix_MultipleTag.py
+++ b/plugins/TagFix_MultipleTag.py
@@ -190,7 +190,7 @@ For further detail, see [the wiki](https://wiki.openstreetmap.org/wiki/Key:acces
         if tags.get("highway") in ("motorway_link", "trunk_link", "primary", "primary_link", "secondary", "secondary_link") and not "maxheight" in tags and not "maxheight:physical" in tags and (("tunnel" in tags and tags["tunnel"] != "no") or tags.get("covered") not in (None, "no")):
             err.append({"class": 71301, "subclass": 0})
 
-        if "waterway" in tags and "level" in tags:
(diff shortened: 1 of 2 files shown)


frodrigo:
I think we just should allow to use level where there is a tunnel (or a birdge), culvert or not.

danfos:
Yes, that makes sense, to allow natural=water/waterway=* in combination with and bridge or tunnel. 

Let me make a new PR for that.

frodrigo:
But, you can just amend this PR.

danfos:
Could indeed amed the PR, but then it did no longer cover the title of the PR, so therefore did chose to make a new PR.

frodrigo:
As you like, but you can also change the PR title ;)

danfos:
That I did not know, let me give that a try.

PR #939, opened 2020-08-04
Have natural=water follow waterway=* on checking for layer.

danfos:
First try in #937 but could not update that PR anymore because I did rebase.

Did analyze #934 further and saw at least one problem was for waterway=* + tunnel=culvert + level=-1. The problem here is the key level, that should be layer instead. This patch adds details to clarify that.

The other problem is for natural=water + tunnel=culvert + level=-1, this patch makes natural=water follow waterway=* on checking layer as that code has already all the good things.

Did check [natural=*](https://wiki.openstreetmap.org/wiki/Key:natural) but only for natural-water layer=* makes sense to me.

### DIFF
--- a/plugins/TagFix_MultipleTag.py
+++ b/plugins/TagFix_MultipleTag.py
@@ -34,7 +34,9 @@ class TagFix_MultipleTag(Plugin):
         self.errors[30323] = self.def_class(item = 3032, level = 3, tags = ['tag', 'fix:chair'],
             title = T_('Watch multiple tags'))
         self.errors[30327] = self.def_class(item = 3032, level = 2, tags = ['tag', 'fix:chair'],
-            title = T_('Waterway with level'))
+            title = T_('Waterway with level'),
+            detail = T_('Level should be used for buildings, shops, amenities, etc.'),
+            trap = T_('Remove level and check if layer is needed instead'))
         self.errors[303210] = self.def_class(item = 3032, level = 1, tags = ['tag', 'highway', 'fix:chair'],
             title = T_('Fence with material tag, better use fence_type tag'))
         self.errors[20800] = self.def_class(item = 2080, level = 1, tags = ['tag', 'highway', 'roundabout', 'fix:chair'],
--- a/plugins/TagRemove_Layer.py
+++ b/plugins/TagRemove_Layer.py
@@ -39,20 +39,20 @@ class TagRemove_Layer(Plugin):
         self.errors[41106] = self.def_class(item = 4110, level = 2, tags = ['highway', 'fix:chair'],
             title = T_('Long Highway above ground and no bridge'))
         self.errors[41107] = self.def_class(item = 4110, level = 3, tags = ['highway', 'fix:chair'],
-            title = T_('Waterway underground and no tunnel'))
+            title = T_('Waterway/water underground and no tunnel'))
         self.errors[41108] = self.def_class(item = 4110, level = 2, tags = ['highway', 'fix:chair'],
-            title = T_('Long Waterway underground and no tunnel'))
+            title = T_('Long Waterway/water underground and no tunnel'))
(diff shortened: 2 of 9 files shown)


frodrigo:
Ok, for me.
But tests fails:
```
    def way(self, data, tags, nds):

        if tags.get(u"layer") and tags.get(u"layer") != "0":

            layer = tags.get(u"layer")

            if tags.get(u"landuse"):

                return {"class": 41101, "subclass": 0}

            elif tags.get(u"natural") and layer[0] == '-' and tags.get(u"natural") != "water":

                return {"class": 41102, "subclass": 0}

            elif tags.get(u"highway") and tags.get(u"highway") != "steps" and (not tags.get(u"indoor") or tags.get(u"indoor") == "no"):

                if layer[0] == "-" and (not tags.get(u"tunnel") or tags.get(u"tunnel") == "no"):

                    return {"class": 41104 if len(nds) > 3 else 41103, "subclass": 0}

                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no") and tags.get(u"man_made") != "pier":

                    if len(nds) > 3:

                        return {"class": 41106, "subclass": 0, "fix": {"-": ["layer"]}}

                    else:

                        return {"class": 41105, "subclass": 0, "fix": {"+": {"bridge": "yes"}}}

            elif tags.get(u"waterway") or ((tags.get(u"natural") and tags.get(u"natural") == "water")):

                if layer[0] == "-" and (not tags.get(u"tunnel") or tags.get(u"tunnel") == "no"):

                    if len(nds) > 3:

                        return {"class": 41108, "subclass": 0, "fix": {"-": ["layer"]}}

                    else:

                        return {"class": 41107, "subclass": 0}

                elif layer[0] != "-" and (not tags.get(u"bridge") or tags.get(u"bridge") == "no"):

>                   return {"class": 41100 if len(nds) > 3 else 41109, "subclass": 0}

E                   TypeError: object of type 'NoneType' has no len()

plugins/TagRemove_Layer.py:72: TypeError
```

danfos:
Yes, saw the error this morning but had to go to work. Let me see if I can fix it.

danfos:
Fixed one problem I think but there are many others that I likely can also fix but that will take time.
I am surprised that changing some text requires many updates in the test data.

frodrigo:
Forget about the test data and version error, we need to change this.
Just look at the error I pointed.

danfos:
Had the fix ready for the test data so pushed it.

I did not see the "TypeError: object of type 'NoneType' has no len()" in the Travis log, how was it triggered?

The error is for line 72:

return {"class": 41100 if len(nds) > 3 else 41109, "subclass": 0}

but that line was not touched by the commit.

The problem seems to be that the nds variable has the value None but that is an function argument, so it is the calling function doing that.

frodrigo:
Yes the test set no nodes https://github.com/osm-fr/osmose-backend/blob/999e0a672614b089cf755e7169d385f6534aded0/plugins/TagRemove_Layer.py#L84
But previously this part of code was not called by the test (it is bad, but it is not the point). But the change in the previous if change the flow of process to this part, and it is not want we want. It is due to the add of ` and tags.get(u"natural") != "water"`

danfos:
Thanks for the hint!

Took me some time to understand nds means nodes so normally this function is called with a list of nodes of the way.

Fixed the natural test and added waterway/water tests, let's see.

danfos:
Okay, looks to me the problems with the changes an sich are gone, but what now seems to fail is the checksum, as the file changed the checksum changed?!

Just a matter of updating the magic number in modules/SourceVersion.py?

frodrigo:
No, about source version it is ok, is liked to an other issue.

frodrigo:
Thank. Merged, but not deployed, yet.

danfos:
Thanks, again some issues down on the list of objects I have recently edited with Osmose warnings once deployed.

PR #940, opened 2020-08-05
Analyser for AEDMAP defibrillators

PanierAvide:
Analyser for checking defibrillators from this database published on data.gouv.

### DIFF
--- a/analysers/analyser_merge_defibrillators_FR.py
+++ b/analysers/analyser_merge_defibrillators_FR.py
@@ -31,7 +31,9 @@ class Analyser_merge_defibrillators_FR(Analyser_Merge):
             return None
         else:
             etg = unidecode.unidecode(etg.lower().replace("-", "").replace(".", "").replace(" ", ""))
-            if etg in ["rezdechaussee", "rezdejardin", "rezdechausse", "rez", "red-de-chaussee", "rdc", "rdj"]:
+            if etg == "0":
+                return None
+            elif etg in ["rezdechaussee", "rezdechausse", "reddechaussee", "rdc"]:
                 return "0"
             elif re.compile(r"^r\+\d$").match(etg):
                 return etg[2:]
@@ -55,7 +57,8 @@ class Analyser_merge_defibrillators_FR(Analyser_Merge):
     def __init__(self, config, logger = None):
         Analyser_Merge.__init__(self, config, logger)
         self.def_class_missing_official(item = 8370, id = 120, level = 3, tags = ["merge"],
-            title = T_("Defibrillator not integrated"))
+            title = T_("Defibrillator not integrated"),
+            trap = T_("Location of defibrillators from this dataset can be very approximative. Check carefully the position before adding to OSM."))
 
         self.init(
             u"https://geo.data.gouv.fr/fr/datasets/a701db3964e8fd81823c92afc029f138ffa207b3",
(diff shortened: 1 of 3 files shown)


frodrigo:
I have an issue for merging this using the name tag #936.

Maybe not for this PR, but all merging DEA should have better Osmose doc.

PanierAvide:
Should I start a discussion on name tag for defibrillators on talk-fr ?

I can work on documentation, however I'm not sure of the new display of doc is implemented.

frodrigo:
> Should I start a discussion on name tag for defibrillators on talk-fr ?

I think so.
 
> I can work on documentation, however I'm not sure of the new display of doc is implemented.

Look at the doc of the doc https://github.com/osm-fr/osmose-backend/blob/master/doc/0-Index.md#classes-of-issues

PanierAvide:
Discussion is started here : https://lists.openstreetmap.org/pipermail/talk-fr/2020-August/100720.html

PanierAvide:
And pushed documentation for all defibrillators analysers.

analysers/analyser_merge_defibrillators_FR.py
```diff
@@ -55,7 +55,10 @@ def normalizeHours(self, jours, heures):
     def __init__(self, config, logger = None):
         Analyser_Merge.__init__(self, config, logger)
         self.def_class_missing_official(item = 8370, id = 120, level = 3, tags = ["merge"],
-            title = T_("Defibrillator not integrated"))
+            title = T_("Defibrillator not integrated"),
+            detail = T_("This defibrillator is available through open data but seems to be missing in OpenStreetMap."),
+            fix = T_("Check with ground survey or local knowledge if it is present, and then add it to OSM."),
+            trap = T_("Precision of defibrillator from this database can be very approximative. Be careful when positioning it."))
```
frodrigo, line 61: Note, all opendata merge already inherit the common doc https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Merge.py#L857-L865
Only specific here.

PanierAvide: Hmm. Except trap for GeoDAE database, I'm not sure what can be defined as doc for other defibrillators, I mean it's quite similar to solve compared to other open datasets.

frodrigo: Maybe linked to the bad quality of some sources / geocoding, and the need of survey ?

PanierAvide: OK so no doc by default, and just add doc on dataset we know are of bad quality ? In this case it will be straightforward, as only know about quality of GeoDAE (quite poor), Montfort (good), and AEDMAP (quite good).

frodrigo: OK

PanierAvide: Updated docs accordingly.

PanierAvide:
So according to discussion on list, what seem to emerge is to put the information into `defibrillator:location`, could be a good solution if no more precise access description is given in datasource. So if that's OK for you, I can do following changes for GeoDAE + AEDMap :
- Remove use of name tag
- Use name from input datasource as defibrillator:location if no better information for this tag is available

frodrigo:
Ok, for me.

PanierAvide:
Upated accordingly, also changed for GeoDAE to not keep level value "0" (but keeping explicit textual values like "rdc", "rez-de-chaussée" and so on, which are translated to level=0).

frodrigo:
> but keeping explicit textual values like "rdc", "rez-de-chaussée" and so on, which are translated to level=0

Je n'avais pas fait attention à ça, mais je pense aussi que c'est de la sûr-interprétation pour rezdejardin, rdj, si ce n'est pas rdc qui est utilisé, c'est qu'il y a une différence avec le rdc et que ça peut ne pas correspondre à "0". Ici à Bordeaux des avenues entière ou le rdj est le -1.

On est sur des DAE, je pense que c'est important de ne introduire de fausse information quand on n'est pas sûr.

PanierAvide:
OK, so I keep "0" as "0", but remove rez-de-jardin as "0" ?

PanierAvide:
Updated accordingly

frodrigo:
> OK, so I keep "0" as "0", but remove rez-de-jardin as "0" ?

No, for me 0 should be excluded.

PanierAvide:
OK, that is what the code is doing now

frodrigo:
Good for me with the current version.

frodrigo:
Deployed

PanierAvide:
Great, thanks :smiley:

PR #959, opened 2020-08-29
Handle path/pedestrain like a footway/platform to check if a public transport platform is part of a way

danfos:
I was seeing some Issues saying the public transport platform is not part of a way and checking the code I found highway=path was not included in the check. Also added highway=pedestrian.


frodrigo:
Merged, not deployed, yet. Thank you.

frodrigo:
Deployed

PR #960, opened 2020-08-31
Add source to Analyser_Merge_Charging_station_FR

gileri:
Implement a source attribute for the Analyser_Merge_Charging_station_FR

The dataset used for this analyser is an aggregation of multiple other open datasets, that provides an acceptable but not perfect attribution for each data point : a link to the original dataset on data.gouv.fr.

I believe that is sufficient for Osmose to comply with the license.


frodrigo:
Thank.

PR #969, opened 2020-09-12
Add source tag to Analyser_Merge_Emergency_Points_LU

gileri:
Similar to #960, implement a source attribute for the Analyser_Merge_Emergency_Points_LU.

I checked and there should not be any other Merge Analyser with missing source tag.


frodrigo:
Great, thank you for looking at this.

PR #975, opened 2020-09-19
Fix analyser aed FR (remove ref as it's not stable yet)

PanierAvide:
This PR fixes the analyser_merge_defibrillator_FR : documentation and questions asked both said that column `gid` is stable over time in dataset, recent updates show that's it's not true yet. In order to avoid headaches, I removed ref from analyser and increased conflation distance, this will permit a better conflation for now.


frodrigo:
Deployed

Issue #953, opened 2020-08-19
National bank of canada errors shows up in Germany


thibaultmol:
Hi,

Someone on the osm Discord server pointed out an error that osmose generated (already marked as false positive)n
where a 'brand in name' error appeared on this node https://www.openstreetmap.org/node/289980683 
for the National bank of canada, yet the geojson file that's used for it's region on the Name Suggestion index seems correct: https://ideditor.github.io/location-conflation/?locationSet=%7B%22include%22%3A%5B%22quebec.geojson%22%5D%7D&referrer=nsi

page in NSI: https://nsi.guide/index.html?k=amenity&v=bank#Banque%20Nationale

Does the osmose analyze not work with .geojson files yet? or is there some other problem happening?

Cheers,

frodrigo:
https://github.com/osm-fr/osmose-backend/blob/4af11a1846c96a0ddd2e5fad28f6858c1bed3a0c/plugins/TagFix_Brand.py

cc @nlehuby 

We need to implement polygon support or exclude rules with polygon. Better if we can use country / area code than polygon.

frodrigo:
https://github.com/osmlab/name-suggestion-index/blob/main/CONTRIBUTING.md#locationset

Only code in  ISO 3166-1 of size 2 can match Osmose code implementation. Nevertheless it could a satisfiant partial implementation.

frodrigo:
@thibaultmol not sure what is the issue here. I see the code already check for the country and seem correct setting on NSI side also
https://nsi.guide/index.html?k=amenity&v=bank#nationalbank-d90ecd
https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagFix_Brand.py#L60

frodrigo:
Finally found two issues in the code #988.

frodrigo:
Should be ok, but reopen if not.

PR #988, opened 2020-10-03
Fix country include/exclude of nsi #953

frodrigo:


plugins/TagFix_Brand.py
```diff
@@ -85,9 +88,9 @@ def node(self, data, tags):
                                     tags_to_add[tag] = brands_tags[tag]
                             return {"class": 31301, "subclass": 0,
                                     "fix": {"+": tags_to_add}}
-                        else:
-                            return {"class": 31301, "subclass": 1,
-                                    "fix": {"+": {"brand": tags["name"]}}}
+#                        else:
+#                            return {"class": 31301, "subclass": 1,
+#                                    "fix": {"+": {"brand": tags["name"]}}}
```
frodrigo, line 93: @nlehuby not sure what this part is for. Why just create brand from name when the name is frequent. May be a false positive.

Note: without httis par can also remove the full `_get_frequent_names()`.

nlehuby: Note sure to understand :thinking:
This part is actually to create brand from name when name is frequent but not part of nsi.
how do you get frequent names if you remove `_get_frequent_names()`

frodrigo: But why do you want to create brand from frequent name, not using nsi? It the cause of #953

nlehuby: The purpose was to foster contribution to the nsi: if it's a false positive, a change may be made to remove generic name or add synonyms to existing brand according to the case. If not, the brand metadata may be missing in nsi (and adding a `brand` tag in OSM is still a good addition).

But if you think there are too much false positive, you can remove it and only use nsi and add brand tags to already known brands objects.

frodrigo: Ok thank you, Yes I prefer remove possible false positive.

nlehuby:
`countryCodes` were remplaced by `locationSet` on last June: https://github.com/osmlab/name-suggestion-index/commit/4323d115472a0c7f44e629c1a3ae54777bbdcc30
I wonder how we can keep track of the format changes in NSI to be more reactive about these...

PR #990, opened 2020-10-05
Remove duplicate value

Famlam:
Remove duplicate value mountain_pass (shown in warnings for `Untagged named object`)


frodrigo:
Thank you.

PR #1014, opened 2020-10-25
Update merge_service_public_FR

nlehuby:
according to https://wiki.openstreetmap.org/wiki/France/Services_Publics


analysers/analyser_merge_service_public_FR.py
```diff
@@ -130,20 +130,49 @@ def __init__(self, config, logger = None):
...
+class Analyser_Merge_ServicePublic_FR_CG(_Analyser_Merge_ServicePublic_Name_FR):
     def __init__(self, config, logger = None):
-        _Analyser_Merge_ServicePublic_Name_FR.__init__(self, config, logger, 8110, 9, 3, ["cg", "cr"], {"office": "government", "government": "parliament"}, {"office": "government", "government": "parliament"})
+        _Analyser_Merge_ServicePublic_Name_FR.__init__(self, config, logger, 8110, 9, 3, "cg",
+                        {"office": "government", "government": "parliament"},
+                        {"office": "government", "government": "parliament", "admin_level": "6"})
+
+class Analyser_Merge_ServicePublic_FR_CR(_Analyser_Merge_ServicePublic_Name_FR):
+    def __init__(self, config, logger = None):
+        _Analyser_Merge_ServicePublic_Name_FR.__init__(self, config, logger, 8110, 9, 3, "cr",
```
frodrigo, line 145: Cannot duplicate class=9

analysers/analyser_merge_service_public_FR.py
```diff
@@ -130,20 +130,48 @@ def __init__(self, config, logger = None):
...
+    def __init__(self, config, logger = None):
+        _Analyser_Merge_ServicePublic_Name_FR.__init__(self, config, logger, 8110, 14, 3, ["chambre_agriculture", "chambre_metier", "cci"],
+                        {"office": "government", "government": "chamber_of_commerce"},
+                        {"office": "government", "government": "chamber_of_commerce"})
+
+class Analyser_Merge_ServicePublic_FR_CR(_Analyser_Merge_ServicePublic_Name_FR):
+    def __init__(self, config, logger = None):
+        _Analyser_Merge_ServicePublic_Name_FR.__init__(self, config, logger, 8110, 15, 3, "cr",
+                        {"office": "government", "government": "parliament"},
+                        {"office": "government", "government": "parliament", "admin_level": "4"})
```
frodrigo, line 177:
```suggestion
                        {"office": "government", "government": "parliament", "admin_level": "4"})

```

frodrigo:
Thank you!

frodrigo:
Deployed

nlehuby:
That was quick ! thanks :partying_face: 
How often is the source opendata file updated ?

frodrigo:
Cache default 30d.
https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Merge.py#L312

nlehuby:
not really related to this PR, but there are a lot of false positive with this analysis (Can be seen quite easily by browsing the townhall class)
I thought it was due to outdated open data source, but 30d is ok, that's not the reason ... 

For now, it seems that we have two kinds of issues : some bad geoloc in the open dataset and lots of OSM element as multipolygon relation :thinking:

frodrigo:
We can also conflate with relations it is not an issue.

Concerning the data update I just see there is no. It use a static file. Last update from 2016. Using : https://github.com/osm-fr/osmose-backend/blob/master/merge_data/service_public_FR.rb
Must be redo in pyhton as specific data parser.

PR #1016, opened 2020-10-25
Fix untagged named obj when main tag is animal=*

fansanelli:
Hello,

I think "animal" should be considered a valid main tag, see:

animal = school
animal = swimming
animal = wellness
animal = cemetery
animal = crematorium

tha are suggested by wiki pages:

https://wiki.openstreetmap.org/wiki/Animals
https://wiki.openstreetmap.org/wiki/Tag:animal%3Dschool

Here an example of use:

http://osmose.openstreetmap.fr/it/error/520b6aa6-370e-1fdc-1f17-2720d4467c15

Many thanks
Cheers
Francesco

### DIFF
--- a/plugins/TagFix_MultipleTag.py
+++ b/plugins/TagFix_MultipleTag.py
@@ -28,7 +28,7 @@ class TagFix_MultipleTag(Plugin):
     def init(self, logger):
         Plugin.init(self, logger)
         self.country = self.father.config.options.get("country")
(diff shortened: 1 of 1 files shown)


frodrigo:
Not sure about this. The usage is low and in conflict with `amenity=animal_training`. Both very low usage but on the wiki.

Marc-marc-marc:
animale=* seems to be a mess, a mix about :
- access tag : =yes/no : is animal allowed in a poi (main usage)
- "content" description : what kind of animal in an area (for ex in a pasture or an area in a zoo) : horse, dog, cow, pig, birds, bear, lion ... 
- feature for animals : school (funny value for training, with a similar tag amenity=animal_training)
imho mixing meaning 2&3 with the main usage is a bad idea.

PR #1023, opened 2020-11-09
Add sax plugin for capacity and its subkeys

gileri:
This plugins check that :
* `capacity=*` and `capacity:*=*` values are positive integers
* `capacity:*=*` subvalues are inferior or equal to `capacity=*`

https://wiki.openstreetmap.org/wiki/Key:capacity

I picked 3500 as an item number, but I'm not sure how those should be allocated.

### DIFF
--- a/plugins/Capacity.py
+++ b/plugins/Capacity.py
@@ -83,7 +83,7 @@ class Capacity(Plugin):
                         "class": 30913,
                         "subclass": 1,
                         "text": T_(
-                            'Specific "{0}" value "{1}" is larger than total capacity {2}',
+                            'Specific "{0}" value "{1}" should be lower than total capacity {2}',
                             key,
                             value,
                             total_capacity,
--- a/plugins/Capacity.py
+++ b/plugins/Capacity.py
@@ -44,25 +44,26 @@ class Capacity(Plugin):
         )
 
     def node(self, data, tags):
+        errors = []
         if ("capacity" not in tags
                 # Ignore errors that should be reported by generic analysers
                 or tags["capacity"] == ""):
-            return
+            return []
         try:
             total_capacity = int(tags["capacity"])
+            if total_capacity < 0:
+                return [{
+                    "class": 30913,
+                    "subclass": 5,
+                    "text": T_('"{0}" value "{1}" is negative', "capacity", total_capacity),
+                }]
         except ValueError:
-            return {
+            errors.append({
                 "class": 30912,
                 "subclass": 4,
                 "text": T_('"{0}" value "{1}" is not an integer', "capacity", tags["capacity"]),
-            }
-
-        if total_capacity < 0:
-            return {
-                "class": 30913,
-                "subclass": 5,
-                "text": T_('"{0}" value "{1}" is negative', "capacity", total_capacity),
-            }
+            })
+            total_capacity = None
 
         for key, value in tags.items():
             if (
@@ -77,31 +78,33 @@ class Capacity(Plugin):
 
             try:
                 capacity_int = int(value)
+                if total_capacity is not None and capacity_int > total_capacity:
+                    errors.append({
+                        "class": 30913,
+                        "subclass": 1,
+                        "text": T_(
+                            'Specific "{0}" value "{1}" is larger than total capacity {2}',
+                            key,
+                            value,
+                            total_capacity,
+                        ),
+                    })
+
+                if capacity_int < 0:
+                    errors.append({
+                        "class": 30912,
+                        "subclass": 5,
+                        "text": T_('"{0}" value "{1}" is negative', key, value),
+                    })
             except ValueError:
-                return {
+                errors.append({
                     "class": 30912,
                     "subclass": 4,
                     "text": T_('"{0}" value "{1}" is not an integer', key, value),
-                }
+                })
+                continue
 
-            if capacity_int > total_capacity:
-                return {
-                    "class": 30913,
-                    "subclass": 1,
-                    "text": T_(
-                        'Specific "{0}" value {1} is larger than total capacity {2}',
-                        key,
-                        value,
-                        total_capacity,
-                    ),
-                }
-
-            if capacity_int < 0:
-                return {
-                    "class": 30912,
-                    "subclass": 5,
-                    "text": T_('"{0}" value "{1}" is negative', key, value),
-                }
+        return errors
 
     def way(self, data, tags, nds):
         return self.node(data, tags)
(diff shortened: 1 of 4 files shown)


plugins/Capacity.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def node(self, data, tags):
+        try:
+            if tags["capacity"] == "":
+                # Ignore errors that should be reported by generic prefix analysers
+                return
+            total_capacity = int(tags["capacity"])
+        except KeyError:
+            # No capacity tag
+            return
```
frodrigo, line 46: Could you just check for "capacity" in tags, rather than raisin an exception (to be the same as other plugins).

plugins/Capacity.py
```diff
@@ -0,0 +1,133 @@
...
+        except ValueError:
+            return {
+                "class": 35001,
+                "subclass": 0,
+                "text": T_('Invalid "{0}" value: {1}', "capacity", tags["capacity"]),
+            }
```
frodrigo, line 52: Maybe we can check this in Number plugin ?

frodrigo:
Maybe add it in item 3091.

The sum of sub-capacity should lower then total capacity?

plugins/Capacity.py
gileri: Changed in be3f47bc.

gileri:
> The sum of sub-capacity should lower then total capacity?

I don't think it should, usually the same spot/capacity allow for multiple user types, for example :

* bus, coach, hgv, trailer
* disabled and wheelchair

plugins/Capacity.py
gileri: I thought about adding valid integer and min/max validation, but it would introduce additional complexity to Number, which seems already not really [straightforward](https://github.com/osm-fr/osmose-backend/blob/8709c8e/plugins/Number.py#L60-L74). Also it currently allows having an height defined in kilograms ^^.

I tried defining a [declarative rule set](https://gitlab.com/-/snippets/2038252) for the Number plugin (or any plugin really), but while rules would be more explicit and precise, that would also make it tricky to handle edge-cases.

In conclusion I think it would be easier and clearer to handle each ~numerical "tag family" (speed, height, capacity etc.) in its own plugin/class to avoid convoluted case handling, even if that causes a bit of repetition.

What do you think ?

frodrigo: MapCSS may be more appropriate for "declarative" rule set. JOSM already have a MapCSS for this.

plugins/Capacity.py
```diff
@@ -0,0 +1,131 @@
...
+                or key == "capacity:"
+                or value == ""
+                or value in ("yes", "no")
+            ):
+                continue
+
+            try:
+                capacity_int = int(value)
+            except ValueError:
+                return {
```
frodrigo, line 72: You may return an array, you do not have to return on first issue.

plugins/Capacity.py
gileri: JOSM does have some numeric validation in MapCSS, but not for `capacity`, `capacity:*`. Can you expand on what you think should be done ? Thank you :)

frodrigo: The rule can (probably) also be written in MapCSS an shared with JOSM.

gileri: I agree that declarative (and MapCSS should be preferred when possible. This particular error could indeed be handled by a MapCSS rule. However :

* The plugin must do this check anyway to compare the value to sub-capacities later. Why not raise the error ?
* [MapCSS don't support wildcard/regex for keys](https://wiki.openstreetmap.org/wiki/MapCSS/0.2#Tests), so the same check for subcapacities must be implemented in this plugin anyway.
* That would split error handling on this tag between two plugins.

Do you think I should make a separate MapCSS and Plugin containing this check only ? I don't mind as it seems easy enough thanks to [mapcss2osmose.py](https://github.com/osm-fr/osmose-backend/blob/master/mapcss/mapcss2osmose.py).

frodrigo: So ok, we keep think like this.

plugins/Capacity.py
gileri: Thank you for the suggestion, I made it so in 3ddfae591179175b88d8e3156417625157bafd76.

gileri:
> Maybe add it in item 3091.

That look fitting indeed, I changed the error definitions in 9bb8e343385659a278a121d3a3c8470dd1090f3d.

plugins/Capacity.py
```diff
@@ -26,79 +26,90 @@ class Capacity(Plugin):
...
-            detail=T_("""A capacity tag value is incorrect."""),
+            detail=T_("""Capacity tags should be positive integers."""),
+            resource="https://wiki.openstreetmap.org/wiki/Key:capacity",
+        )
+        self.errors[30913] = self.def_class(
+            item=3091,
+            level=2,
+            tags=["tag"],
+            title=T_("Specific capacity is greater than total capacity"),
+            detail=T_("""A capacity:* value is greater than the total capacity."""),
```
frodrigo, line 42:
```suggestion
            detail=T_("""A capacity:* value should be lower than the total capacity."""),
```

plugins/Capacity.py
```diff
@@ -0,0 +1,147 @@
...
+            total_capacity = None
+
+        for key, value in tags.items():
+            if (
+                not key.startswith("capacity:")
+                # Ignore errors that should be reported by generic prefix analysers
+                or key == "capacity:"
+                or value == ""
+
+                or value in ("yes", "no")
```
frodrigo, line 75:
```suggestion
                or value in ("", "yes", "no")
```

plugins/Capacity.py
gileri: Changed in b939fc85dc648b50702dc4d8c2b1007fb39c3310

plugins/Capacity.py
gileri: I split those two checks as `""` is an invalid value that should be checked in other Analysers, while `yes|no` are valid values. I believe the performance penalty is really low for the clarity added by [the comment L71](https://github.com/osm-fr/osmose-backend/pull/1023/files#diff-d376487ccb6816192f500423b8279401ea86f138eb7be2bbb05c26567be043e9R71-R73).

frodrigo: Ok. So just remove the empty line, please.

gileri: I reordered the conditions to remove the empty line while still keeping the comment relevant, I hope that's okay.

plugins/Capacity.py
```diff
@@ -0,0 +1,147 @@
...
+            })
+            total_capacity = None
+
+        for key, value in tags.items():
+            if (
+                not key.startswith("capacity:")
+                # Ignore errors that should be reported by generic prefix analysers
+                or key == "capacity:"
+                or value == ""
+
```
frodrigo, line 74:
```suggestion
```

frodrigo:
Merged, thank you.

gileri:
Thank you @frodrigo for the review and merge !

frodrigo:
Deployed.

Famlam:
Added a minor fix for man_made=* keys, where it's commonly used for volumes. (And volumes of course can be decimal and/or with units) #1398

PR #1029, opened 2020-11-20
Avoid false positive detections of railway signals

Famlam:
Prevent false positives due to railway signals. Example: http://osmose.openstreetmap.fr/nl/error/fed5255c-48a9-0e25-56c7-4cc31bc92657
Mapillary flags a signal (correctly) as a traffic light. However, the traffic light applies to trains and is hence tagged as railway=signal


frodrigo:
Thank you.

PR #1034, opened 2020-11-30
Relax key characters for hazmat:*

### LINKED ISSUE #948: item=3050 class=30501 allow dot char '.' in tag name
The `.` char is sometimes used for the `ref` key to reference a website source. 

For instance French churches may use https://clochers.org website and use `ref:clochers.org=xxx` tag to link with it. 

It should be OK to allow that character or at least decrease the severity of such issues.

1ec5:
Exempt ~~`ref:*` and~~ `hazmat:*` from the requirement for alphanumeric subkeys in item 3050, class 30501.

/ref #948

### DIFF
--- a/plugins/TagFix_BadKey.py
+++ b/plugins/TagFix_BadKey.py
@@ -60,7 +60,6 @@ separator '_' or ':'. See
                                 "voltage-high", "voltage-low",
                                 "cityracks.housenum", "cityracks.installed", "cityracks.large", "cityracks.rackid", "cityracks.small", "cityracks.street", # NYC amenity=bicycle_parking
                                 "strassen-nrw", # DE import
-                                "ref",
                                 "hazmat",
                              ) )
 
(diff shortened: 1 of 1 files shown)


frodrigo:
Ok for `hazmat`. As it is like this on the wiki.
But not for ref, no reason to white list all.

1ec5:
I was just going off of the bug report in #948. Would it be more appropriate to allow just `ref:clochers.org` specifically?

frodrigo:
> I was just going off of the bug report in #948. Would it be more appropriate to allow just `ref:clochers.org` specifically?

No, `ref:clochers.org` should be reported. It contains a dot. But for some frequent char like dot, the level of the issue must be moved from 1 to 2. As I suggest here https://github.com/osm-fr/osmose-backend/issues/948#issuecomment-691710164g

1ec5:
Ah, I misunderstood that comment. 👍 At a glance, I’m unsure how to make the level depend on the character without creating a separate kind of issue, so for now I’ve removed the `ref:*` part of this PR and kept just the `hazmat:*` part.

frodrigo:
Merged, thank. But nor deployed, yet.

frodrigo:
> Ah, I misunderstood that comment. +1 At a glance, I’m unsure how to make the level depend on the character without creating a separate kind of issue, so for now I’ve removed the `ref:*` part of this PR and kept just the `hazmat:*` part.

Yes, it requires to create an other class of issue.

PR #1036, opened 2020-12-01
merge_poste_FR : add brand. don't set operator for postoffice not ope…

Marc-marc-marc:
add brand
don't set operator for postoffice not operated by LaPoste


frodrigo:
Deployed

PR #1040, opened 2020-12-02
fix nsi url. close #1039

### LINKED ISSUE #1039: NSI dist file in 404
Osmose is using https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/brands.json in  https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagFix_Brand.py#L48
but it return 404 due to the fact the master branch is deleted https://github.com/osmlab/name-suggestion-index/issues/4748
the https://raw.githubusercontent.com/osmlab/name-suggestion-index/main/dist/brands.json was deleted by https://github.com/osmlab/name-suggestion-index/commit/6757a8820520f41b2ba45858e4f78351700696e4
new file seem to be https://github.com/osmlab/name-suggestion-index/commit/6757a8820520f41b2ba45858e4f78351700696e4 semble dire que c'est https://github.com/osmlab/name-suggestion-index/blob/main/dist/index.min.json or https://github.com/osmlab/name-suggestion-index/blob/main/dist/index.json for not-minified version
it's correct ? witch one to use ?

Marc-marc-marc:


frodrigo:
Close, alternate fix.

PR #1044, opened 2020-12-06
revert to long form for mounts to allow changing

stephankn:
long form is not only more readable, but also allows to change the mounts by adding additional compose configs, specifically when replacing mounts by tmpfs.


frodrigo:
Thank you

PR #1051, opened 2020-12-13
Opening hours: port from oh_sanitizer to PyKOpeningHours

dfaure:
The KOpeningHours parser is much more complete, it implements
the full grammar plus tolerance for some invalid input.
See the tests at https://invent.kde.org/libraries/kopeninghours/-/blob/master/autotests/parsertest.cpp

The python wrapping using boost::python is in https://github.com/dfaure/PyKOpeningHours

This fixes issue 555.


dfaure:
I have tested the build of the C++ module in docker, that part works.
I'm having trouble testing TagFix_Opening_Hours.py though, due to the way it imports modules.

```
osmose@b075fa1d4f19:/opt/osmose-backend/plugins$ python TagFix_Opening_Hours.py 
  File "TagFix_Opening_Hours.py", line 22, in <module>
    from modules.OsmoseTranslation import T_
ModuleNotFoundError: No module named 'modules.OsmoseTranslation'
```
and yet `$PYTHONPATH/modules/OsmoseTranslation.py` exists.

dfaure:
Travis CI fails because it only offers Qt 5.9. This is weird because the Dockerfile pulled Qt 5.11 for me, from debian buster.
It seems the CI uses ubuntu bionic while the docker setup uses debian buster, how come?

frodrigo:
> I have tested the build of the C++ module in docker, that part works.
> I'm having trouble testing TagFix_Opening_Hours.py though, due to the way it imports modules.

Run the test from the root `python plugins/TagFix_Opening_Hours.py`

frodrigo:
> Travis CI fails because it only offers Qt 5.9. This is weird because the Dockerfile pulled Qt 5.11 for me, from debian buster.
> It seems the CI uses ubuntu bionic while the docker setup uses debian buster, how come?

It is a basic Travis CI, without using Docker.

docker/Dockerfile
```diff
@@ -19,8 +19,12 @@ RUN apt-get update && \
...
         python-pip \
         python-setuptools \
-        python-wheel && \
-    apt-get clean
+        python-wheel \
+        cmake \
+        qtbase5-dev \
+        flex \
+        bison \
+        && apt-get clean
```
frodrigo, line 27: Should be in alphabetical order

frodrigo:
Is Qt an heavy used dependency or can be removed ?

@jocelynj what do you think about adding Qt as dependency ?

dfaure-kdab:
I solved the Qt version issue already, it builds with Qt 5.9 now.

Removing the Qt dependency would mean a very invasive fork of KOpeningHours, I'm hoping we can avoid going there.
It builds and it works, what's the problem with a few more libs installed?

dfaure:
> Run the test from the root `python plugins/TagFix_Opening_Hours.py`

Same result:
```
osmose@b075fa1d4f19:/opt/osmose-backend$ python plugins/TagFix_Opening_Hours.py 
  File "plugins/TagFix_Opening_Hours.py", line 22, in <module>
    from modules.OsmoseTranslation import T_
ModuleNotFoundError: No module named 'modules.OsmoseTranslation'
```

jocelynj:
Can you point to which debian packages are needed to activate qt?

To run a test, it is better to use: "pytest-3 plugins/TagFix_Opening_Hours.py"

dfaure:
The debian package needed is `qtbase5-dev`, as indicated in this merge request.

Thanks for the hint about pytest, I'll fix what it says.

modules/build_c++.sh
```diff
@@ -0,0 +1,16 @@
...
+git clone https://github.com/dfaure/PyKOpeningHours.git
+cd PyKOpeningHours
+mkdir build
+cd build
+cmake ..
+make
+# Remove these directories, their python script don't pass CI, and we don't need them anymore, anyway
+rm -rf external/share/ECM
+rm -rf external/src
+mv PyKOpeningHours.so ../..
```
frodrigo, line 16: I think the right way of doing this is by adding a setup.py on your project and include it in osmose requirement.txt

Some think like this
```
git+git://github.com/dfaure/PyKOpeningHours.git@v1.33
```

frodrigo:
To pass the mypy linting you must add a pyi file. It declares type on the function.
Look at https://github.com/osm-fr/osmose-backend/blob/master/modules/osm_pbf_parser/osm_pbf_parser.pyi

modules/build_c++.sh
dfaure: Not sure I understand. That `setup.py` file is supposed to run `cmake`, `make` etc.?

frodrigo: Yes. But I never do it myself.
https://docs.python.org/3/extending/building.html
https://martinopilia.com/posts/2018/09/15/building-python-extension.html

If it is too complicated, we will do it in your way. But my point, is how we will update it on (non docker) sever.

dfaure:
> To pass the mypy linting you must add a pyi file. It declares type on the function.

Thanks for the hint. I generated one using `stubgen -m` then cleaned it up. https://github.com/dfaure/PyKOpeningHours/blob/master/PyKOpeningHours.pyi
(This is all new to me, I'm discovering python...)

Only one issue left with mypy though: it doesn't find modules/PyKOpeningHours/PyKOpeningHours.pyi, it wants it in modules/PyKOpeningHours.pyi for some reason? This is weird because the osm_pbf_parser.pyi in under osm_pbf_parser/. But well it's used a bit differently (`from .osm_pbf_parser import osm_pbf_parser` in a file that's under modules/, while in my case the user is plugins/TagFix_Opening_Hours.py...). Any recommendation?

modules/build_c++.sh
dfaure: Ok, I implemented a [setup.py](https://github.com/dfaure/PyKOpeningHours/blob/master/setup.py) using `skbuild` (cmake wrapper). Was a bit tricky (I even submitted [a MR for skbuild](https://github.com/scikit-build/scikit-build/pull/522), but I implemented a workaround so we don't need that fix to be merged). Anyhow, now `python setup.py build` works in PyKOpeningHours.

However adding that as a requirement in osmose-backend seems to hit an issue I don't really understand.
```
Step 10/17 : RUN pip install -r requirements.txt ${PIP_INSTALL}
 ---> Running in 489932b3afd1
Collecting git+git://github.com/dfaure/PyKOpeningHours.git@v1.1 (from -r requirements.txt (line 12))
  Cloning git://github.com/dfaure/PyKOpeningHours.git (to revision v1.1) to /tmp/pip-req-build-rabkz4i8
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /usr/local/bin/python /usr/local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpp8segvp5
       cwd: /tmp/pip-req-build-rabkz4i8
  Complete output (7 lines):
  running egg_info
  creating lib/PyKOpeningHours_package.egg-info
  writing lib/PyKOpeningHours_package.egg-info/PKG-INFO
  writing dependency_links to lib/PyKOpeningHours_package.egg-info/dependency_links.txt
  writing top-level names to lib/PyKOpeningHours_package.egg-info/top_level.txt
  writing manifest file 'lib/PyKOpeningHours_package.egg-info/SOURCES.txt'
  error: package directory 'lib/PyKOpeningHours' does not exist
  ----------------------------------------
ERROR: Command errored out with exit status 1: /usr/local/bin/python /usr/local/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpp8segvp5 Check the logs for full command output.
```

Ah, actually `pip install .` in PyKOpeningHours gives the same error. It doesn't even try to build anything, unlike `python3 setup.py build`. Is it supposed to work? Or should I ask the skbuild people?
This might indeed all be rather complicated if skbuild isn't made for this.

But I see your point about "how will we update it". I guess the alternative is to put a copy of the PyKOpeningHours code in osmose-backend, to mimick the osm_pbf_parser solution, but that's not great. Another solution would be a git submodule, if you're up for that.

frodrigo: Your project does not have the standard python module layout. Move PyKOpeningHours.pyi __init__.py pykopeninghours.cpp python_qt_wrappers.cpp python_qt_wrappers.h to dir (aka module) named `PyKOpeningHours` (or maby be `pykopeninghours` to be pytonish).

frodrigo: I successfully install your module using pip. That a good new ! Thank you.

frodrigo: - I think you should remove the `src` dir and have `PyKOpeningHours` on the root dir (can also remove `PyKOpeningHours/lib`). In `setup.py`, package_dir should be `package_dir={'': 'PyKOpeningHours'},`. It make more sens in python.
- It look like to me `PyKOpeningHours.so` should be in `PyKOpeningHours` in dir, and `setup.py` already tell so, but not working `cmake_install_dir='PyKOpeningHours`.
- By copping `PyKOpeningHours.so` to `PyKOpeningHours`. I am able to load the python module from `PyKOpeningHours` dir, but not from outside.

Do you want a PR on this?

frodrigo: - I found the last issue with the import. The empty file `__init__.py` is missing in `PyKOpeningHours` dir. Python need this to consider a dir as a module.

dfaure: Thanks for taking a look, very appreciated.

The `src` subdir (which I also don't like) seems to be necessary for skbuild to work. It looks for PyKOpeningHours (the package name) under the package_dir (src). At least it's what I thought (so before adding the `src` subdir I tried `package_dir={'': ''}` but that triggered the skbuild bug I mentioned (and fixed)). If it works for you that's great. I can make change myself or merge a MR from you, as you prefer.

I don't really understand your third point though. At but I see a weird install dir indeed, `~/lib/PyKOpeningHours/lib/PyKOpeningHours`, this makes little sense. CMake alone works. So maybe cmake_install_dir should be empty?

Does `python -c 'import PyKOpeningHours; help(PyKOpeningHours)'` work after your changes? It was obviously working in my initial MR, but since the switch to using pip, it doesn't work anymore.

dfaure: > I found the last issue with the import. The empty file __init__.py is missing in PyKOpeningHours dir. Python need this to consider a dir as a module.
 
Heh, my last commit was about removing it, thinking it would help [because help() showed `__init__.py` and nothing else]. I just reverted that commit, so it's back :)

dfaure: > * I think you should remove the `src` dir and have `PyKOpeningHours` on the root dir (can also remove `PyKOpeningHours/lib`). In `setup.py`, package_dir should be `package_dir={'': 'PyKOpeningHours'},`. It make more sens in python.

This leads to `pip install .` saying `error: package directory 'PyKOpeningHours/PyKOpeningHours' does not exist`

frodrigo: I make a more simple version and I'm able to compile and install it with pip.

`~/lib/PyKOpeningHours/lib/PyKOpeningHour`, make no sense for me. It should be in`lib/python3.7/site-packages/PyKOpeningHours`. By moving at this location it is better. But now I have an issue about missing `libKOpeningHours.so.1`

dfaure: Yeah lib/python3.x is missing. Ah you're saying I should do that by hand, I see.

The .so.1 issue is solved if you rebase, it's using a static lib now.

Sorry in advance for the merge conflicts...

plugins/TagFix_Opening_Hours.py
```diff
@@ -23,16 +23,16 @@
 from plugins.Plugin import Plugin
 
 try:
-    from modules.oh_sanitizer import sanitize_field, SanitizeError, InconsistentField
-    module_oh_sanitizer = True
+    from modules.PyKOpeningHours import OpeningHours, Error
```
frodrigo, line 26:
```suggestion
    from PyKOpeningHours.PyKOpeningHours import OpeningHours, Error
```

dfaure: Thanks!

dfaure:
Ah, mypy found an old ./PyKOpeningHours/PyKOpeningHours.pyi locally, damn, that's why it worked locally.

requirements.txt
```diff
@@ -4,12 +4,12 @@ polib
 psycopg2-binary >= 2.7
 shapely >= 1.2
 regex
-lark-parser >=  0.6, < 0.7
 requests
 transporthours
 pyproj >= 2.1.0
 Unidecode
 osmium
+git+git://github.com/dfaure/PyKOpeningHours.git@v1.3.3
```
frodrigo, line 12:
```suggestion
git+git://github.com/dfaure/PyKOpeningHours.git@v1.3.4
```

frodrigo:
@jocelynj do you know the min version of qt to be supported?

jocelynj:
All our backends handled by OSM-Fr have qt 5.11.3, coming from Debian Buster.

frodrigo:
By running it on real data I got a console output
```
unexpected character: � at 1:1
unexpected character: c at 1:1
unexpected character: C at 1:1
unexpected character: C at 1:1
unexpected character: C at 1:1
unexpected character: C at 1:1
unexpected character: W at 1:1
unexpected character: _ at 1:1
unexpected character: e at 1:1
unexpected character: l at 1:1
unexpected character: M at 1:1
unexpected character: l at 1:1
unexpected character: c at 1:1
unexpected character: O at 1:1
unexpected character: p at 1:1
```

Please can you remove this.

frodrigo:
Your new implementation is quite faster than the lark version, 15s vs 1m30 on Gironde.

But you raise lot more issues for now, 134 vs 644.

In particular false positive with space
```diff
-<tag k="opening_hours" v="Mo, We-Fr 09:00-18:00; Tu 10:00-18:00; Sa 09:00-12:00" />
+<tag k="opening_hours" v="Mo,We-Fr 09:00-18:00; Tu 10:00-18:00; Sa 09:00-12:00" />
```

dfaure:
> Your new implementation is quite faster than the lark version, 15s vs 1m30 on Gironde.

Cool. It also parses "Dec 24" correctly, which was the whole point :-)

> But you raise lot more issues for now, 134 vs 644.
> 
> In particular false positive with space
> 
> ```diff
> -<tag k="opening_hours" v="Mo, We-Fr 09:00-18:00; Tu 10:00-18:00; Sa 09:00-12:00" />
> +<tag k="opening_hours" v="Mo,We-Fr 09:00-18:00; Tu 10:00-18:00; Sa 09:00-12:00" />
> ```

This is actually not a false positive. With the space it means "Monday all day, and We-Fr 09:00-18:00".
https://openingh.openstreetmap.de/evaluation_tool/ doesn't show that difference because it applies the same autocorrect than KOpeningHours does.

But if we don't want to be too strict in order to reduce noise and preserve behavior of the old osmose plugin, I can re-instate the ignore-white-space code from the old plugin. It's just technically wrong :-)

frodrigo:
> This is actually not a false positive. With the space it means "Monday all day, and We-Fr 09:00-18:00".
> https://openingh.openstreetmap.de/evaluation_tool/ doesn't show that difference because it applies the same autocorrect than KOpeningHours does.
> 
> But if we don't want to be too strict in order to reduce noise and preserve behavior of the old osmose plugin, I can re-instate the ignore-white-space code from the old plugin. It's just technically wrong :-)

Ok this one. But what about
```diff
-Mo-Fr 08:30-21:30;Sa 08:30-21:00
+Mo-Fr 08:30-21:30; Sa 08:30-21:00
```

frodrigo:
Maybe we can also ignore case ?

jocelynj:
Are you sure about the space signification on `Mo, We-Fr 09:00-18:00;`?

I thought that you would need to replace `,` by `;` to mean that it is opened all day on Monday. So:

* `Mo, We-Fr 09:00-18:00;` -> opened monday, wednesday, thursday and friday from 9h to 18h.
* `Mo; We-Fr 09:00-18:00;` -> opened (monday 0h to 24h), and (wednesday, thursday and friday from 9h to 18h)

dfaure:
> By running it on real data I got a console output
> Please can you remove this.

Done in v1.3.5.

vkrause:
Hi, I've implemented the parser @dfaure is using for this, so he asked me to comment here :)

The basis for this is https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#Syntax_diagram, in particular `<normal_rule_separator>` and `<additioanl_rule_separator>` (and their footnotes).

`<normal_rule_separator>` is defined as `; <space>`. Since the semicolon appears nowhere else, having the space there or not has no semantic difference and is just a question on how precisely you want to follow the specification.

For the comma this is more complicated unfortunately: with a following space a comma is an `<additional_rule_separator>`, without one it's used to separate a set of time spans or weekdays. This is a relevant semantic difference, but syntactically the difference is so subtle that it's way too easy to get wrong. Therefore the existing parsers seem to auto-correct this to what was the most likely intended meaning, the footnotes in the specification describe this as well.

For illustration here's the example I sent to @dfaure earlier, referring to testing in https://openingh.openstreetmap.de/evaluation_tool/, which applies the same auto-correction to the comma/space issue, making the space not seem to matter:

> To see the difference compare "Mo; We-Fr 09:00-18:00" and "Mo,We-Fr
> 09:00-18:00" there. Ie. this means "(Mo 0-24) or (We-Fr 9-18)" rather than the
> (most likely) intended "(Mo,We-Fr) 9-18". That's due to the space, not the
> comma vs. semi-colon.
>
> The difference between semicolon and comma as rule separator you can see here:
> "Mo-Fr 12:00-14:00, We 15:00-17:00" vs. "Mo-Fr 12:00-14:00; We 15:00-17:00".
> Normal rules (semi-colon) override previous results if the day matches (We
> 15-17 is the only result on Wednesdays), additional rules (comma) adds to the
> result if the day matches that of previous rules (We 12:00-14:00 +
> 15:00-17:00). If the day doesn't match, both types are equivalent (and that's
> why the example above is valid to see the effect of the space).

Now, all that is the technical view on how this is specified and implemented of course. Whether you want to expose all of that to your users is an entirely different story, it might very well make sense to focus on issues with a bigger impact.

dfaure:
> Maybe we can also ignore case ?

`oh_sanitizer` did fix case issues, but indeed the osmose plugin used to ignore case differences.
I'm a bit confused about the purpose: isn't the idea to suggest fixes for incorrect entries? The spec mandates strict casing, so fixing case makes sense to me. Of course one might assume all parsers are case-insensitive, but.. where do we stop? If we assume all parsers are tolerant to all possible errors, then this plugin is completely pointless :-)

dfaure:
plugins/TagFix_Opening_Hours.py finally passes but plugins/TagFix_Brand.py seems to fail for unrelated reasons?

frodrigo:
> plugins/TagFix_Opening_Hours.py finally passes but plugins/TagFix_Brand.py seems to fail for unrelated reasons?

Yes, a remote file changed. Please, just rebase.

frodrigo:
> > plugins/TagFix_Opening_Hours.py finally passes but plugins/TagFix_Brand.py seems to fail for unrelated reasons?
> 
> Yes, a remote file changed. Please, just rebase.

In fact, the remote file have changed... again

dfaure:
> In fact, the remote file have changed... again

OK, let me know when to rebase again...

Apart from that I think this MR is ready, I have re-enabled the ignoring of whitespace and case issues as requested.
These type of fixes can indeed be done by a bot rather than manually.
My next project, I guess :)

frodrigo:
Thank you. The number of raised issues is now lower.

On Gironde:
- Lark
  - class 0: 27
  - class 1: 0
- PyKOpeningHours
  - class 0: 105
  - class 1: 107

Real raised issues
```
<tag k="opening_hours" v="Mo-Su 09:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Su 09:00-24:00" />
<tag k="opening_hours" v="summer mo-sa 14:00-16:00" />
<tag k="opening_hours" v="mo-tu:10:00-12:30 14:00-19:00; we:10:00-19:00; th-fr:10:00-12:30 14:00-19:00; sa:10:00-19:00" />
<tag action="modify" k="opening_hours" v="Mo-Tu 10:00-12:30,14:00-19:00; We 10:00-19:00; Th-Fr 10:00-12:30,14:00-19:00; Sa 10:00-19:00" />
<tag k="opening_hours" v="Mo-Fr 00:00-01:00,14:00-00:00; Sa; Su 00:00-02:00,14:00-00:00; PH 00:00-02:00,14:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Fr 00:00-01:00,14:00-24:00; Sa; Su 00:00-02:00,14:00-24:00; PH 00:00-02:00,14:00-24:00" />
<tag k="opening_hours" v="00:00-04:00,10:00-00:00" />
<tag action="modify" k="opening_hours" v="00:00-04:00,10:00-24:00" />
<tag k="opening_hours" v="Mo-Su 11:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Su 11:00-24:00" />
<tag k="opening_hours" v="Mo-Th 09:30-19:30; Fr-Sa 09:30-20:00; " />
<tag action="modify" k="opening_hours" v="Mo-Th 09:30-19:30; Fr-Sa 09:30-20:00" />
<tag k="opening_hours" v="Jul-Aug: Tu-Fr 14:00-18:00; Jul-Aug: Sa 10:00-17:00;Sep-Jun: Tu 10:00-12:00,14:00-18:00; Sep-Jun: We,Sa 10:00-18:00; Sep-Jun: Th,Fr 14:00-18:00" />
<tag action="modify" k="opening_hours" v="Jul-Aug Tu-Fr 14:00-18:00; Jul-Aug Sa 10:00-17:00; Sep-Jun Tu 10:00-12:00,14:00-18:00; Sep-Jun We,Sa 10:00-18:00; Sep-Jun Th,Fr 14:00-18:00" />
<tag k="opening_hours" v="Du lundi au vendredi de 8:30 à 12:30 et de 14:30 à 20:00. Le samedi de 8:30 à 12:30 et de 14:30 à 19:00." />
<tag k="opening_hours" v="Du mardi au vendredi : 9h à 19h Non Stop;Le samedi : de 8h30 à 18h Non Stop" />
<tag k="opening_hours" v="Su-Th 11:00-00:00; Fr-Sa 11:00-02:00" />
<tag action="modify" k="opening_hours" v="Su-Th 11:00-24:00; Fr-Sa 11:00-02:00" />
<tag k="opening_hours" v="Lundi-Jeudi 9h-12h30/14h30-19h30 // Vendredi-Samedi 9h-19h30" />
<tag k="opening_hours" v="Mo-Su 05:00-24:00+" />
<tag k="opening_hours" v="Mardi-Samedi 10h-19h30" />
<tag action="modify" k="opening_hours" v="Tu-Sa 10:00-19:30" />
<tag k="opening_hours" v="Lundi-Samedi 10h-19h30" />
<tag action="modify" k="opening_hours" v="Mo-Sa 10:00-19:30" />
<tag k="opening_hours" v="Lundi-Samedi 10h à 19h" />
<tag action="modify" k="opening_hours" v="Mo-Sa 10:00-19:00" />
<tag k="opening_hours" v="Tu-We-Fr 09:00-12:45,14:00-18:00; Th 09:00-12:45,15:45-18:00; Sa 09:00-12:45" />
<tag k="opening_hours" v="1er dimanches d'avril à septembre, de 14h30 à 18h" />
<tag k="opening_hours" v="Mo-Su 09:30-12:60; Mo-Su 14:00-19:00" />
<tag k="opening_hours" v="juillet,août" />
<tag k="opening_hours" v="Mo-Th 11:00-00:00; Fr 14:00-00:00; Sa 11:00-00:00; Su 12:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Th 11:00-24:00; Fr 14:00-24:00; Sa 11:00-24:00; Su 12:00-24:00" />
<tag k="opening_hours" v="en fonction des évènements" />
<tag k="opening_hours" v="Mo-Fr 20:00-00:00; Sa 12:00-00:00; Su 09:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Fr 20:00-24:00; Sa 12:00-24:00; Su 09:00-24:00" />
<tag k="opening_hours" v="sunrise-sunset" />
<tag k="opening_hours" v="9h-12h // 14h - 19h" />
<tag k="opening_hours" v="Sa 21:00-00:00" />
<tag action="modify" k="opening_hours" v="Sa 21:00-24:00" />
<tag k="opening_hours" v="Mo-Sa 09:00-12:30, 14:00-18:30; PH off; Sun 09:00-12:30" />
<tag action="modify" k="opening_hours" v="Mo-Sa 09:00-12:30,14:00-18:30; PH off; Su 09:00-12:30" />
<tag k="opening_hours" v="9h - 12 h // 13h30 - 17h" />
<tag k="opening_hours" v="L/J = 11H30-14H30  18H30-23H00  V/D= 18H30-23H" />
<tag k="opening_hours" v="Covid-19 : du lundi au samedi 8h30-13h et 14h-18h30" />
<tag k="opening_hours" v="Covid 2019 : permanance du lundi au vendredi 9h-12h30" />
<tag k="opening_hours" v="Su 10:00+" />
<tag k="opening_hours" v="We 07:30-18:30; SH Mo,Tu,Th,Fr 07:30-18:30" />
<tag k="opening_hours" v="Mo-Fr 08:30-19:00 Sa 9:00-12:30" />
<tag action="modify" k="opening_hours" v="Mo-Fr 08:30-19:00; Sa 09:00-12:30" />
<tag k="opening_hours" v="Mo-Su 11:00-15:00,17:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Su 11:00-15:00,17:00-24:00" />
<tag k="opening_hours" v="8/8h30 - 18h30/21h (dépend des mois dans l'année)" />
<tag k="opening_hours" v="24/7”" />
<tag k="opening_hours" v="Sunday 8-12" />
<tag k="opening_hours" v="Mo-Su 10:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Su 10:00-24:00" />
<tag k="opening_hours" v="Mo-Su 10:00-00:00" />
<tag action="modify" k="opening_hours" v="Mo-Su 10:00-24:00" />
<tag k="opening_hours" v="Mo,We 12:00-14:30,18:00-20:00; Tu,Th 12:00-14:30,17:00-20:00; Fr 12:00-14:30,18:00-22:00; Sa 10:30-17:00; Su 09:00-13:00; SH Mo 12:00-20:00; SH Tu-Th 10:00-20:00; SH Fr 10:00-22:00; Jul-Aug off" />
```

dfaure:
Looks good to me. But the 00:00 vs 24:00 for an end time is only a matter of normalization preference, I guess both are equally correct, so it's not great to bother people with this. It makes me wonder if we want to either make KOpeningHours preserve the input, or to add an ignore rule in the plugin here.

frodrigo:
> Looks good to me. But the 00:00 vs 24:00 for an end time is only a matter of normalization preference, I guess both are equally correct, so it's not great to bother people with this. It makes me wonder if we want to either make KOpeningHours preserve the input, or to add an ignore rule in the plugin here.

As you wish. But it is important to avoid false positive.

plugins/TagFix_Opening_Hours.py
```diff
@@ -42,14 +42,14 @@ def sanitize_tags(self, tags):
...
-            sanitized_field = sanitize_field(tags['opening_hours'])
-            if sanitized_field.replace(' ', '').replace('0', '').lower() != tags['opening_hours'].replace(' ', '').replace('0', '').lower(): # Ignore sapce and 0 changes
-                return {"class": 32501, "subclass": 0, 'fix': {'opening_hours': sanitized_field}}
-        except InconsistentField as e:
-            return {"class": 32501, "subclass": 1, 'text': {'en': str(e)}}
-        except SanitizeError:
-            return {"class": 32501, "subclass": 2}
+        parser = OpeningHours()
+        parser.setExpression(tags['opening_hours'])
+        if parser.error() != Error.NoError:
```
vkrause, line 47: I think this should check for Error.SyntaxError || Error.IncompatibleMode rather than NoError. This will fix false-positives e.g. for Error.UnsupportedFeature which only matters for evaluation not for validation (see e.g. the reports for "12:00+" or "SH" selectors, both perfectly valid).

dfaure: Thanks, done.

frodrigo:
> > In fact, the remote file have changed... again

> OK, let me know when to rebase again...

Fixed

plugins/TagFix_Opening_Hours.py
```diff
@@ -42,14 +42,15 @@ def sanitize_tags(self, tags):
...
-            return {"class": 32501, "subclass": 1, 'text': {'en': str(e)}}
-        except SanitizeError:
-            return {"class": 32501, "subclass": 2}
+        parser = OpeningHours()
+        parser.setExpression(tags['opening_hours'])
+        if parser.error() == Error.SyntaxError or parser.error() == Error.IncompatibleMode:
+            return {"class": 32501, "subclass": 1, 'text': {'en': 'The opening_hours value is invalid and could not be parsed'}}
+        sanitized_field = parser.normalizedExpression()
+        # Ignore trivial changes that can be fixed by bots rather than humans like spaces, case errors, etc
+        simplify = lambda s : s.replace(' ', '').replace('0', '').replace('24:00', '00:00').lower()
```
frodrigo, line 51:
```suggestion
        simplify = lambda s: s.replace(' ', '').replace('24:00', '00:00').replace('0', '').lower()
```

plugins/TagFix_Opening_Hours.py
```diff
@@ -69,9 +70,14 @@ def test(self):
...
+        self.check_err(a.node(None, {'opening_hours': 'Monday to Friday 8:00AM to 4:30PM'}), expected={"class": 32501, "subclass": 0, 'fix': {'opening_hours': 'Mo-Fr 08:00-16:30'}})
 
+        # These return a parse error
+        self.check_err(a.node(None, {'opening_hours': 'mo-fr 10h - 19h00"'}), expected={"class": 32501, "subclass": 1, 'text': {'en': 'The opening_hours value is invalid and could not be parsed'}})
+        self.check_err(a.node(None, {'opening_hours': '09:00-21:00 TEL/072(360)3200'}), expected={"class": 32501, "subclass": 1, 'text': {'en': 'The opening_hours value is invalid and could not be parsed'}})
+
+        # These are OK, no suggestion
         assert not a.node(None, {'opening_hours': 'Mo-Fr 10:00-19:00'})
-        assert not a.node(None, {'opening_hours': 'Mo-Fr   10:00 -19:00'})
+        assert not a.node(None, {'opening_hours': 'Mo-Tu,Th-Fr 09:30-12:00; We 15:00-17:00; 2020 Dec 24,2020 Dec 31 off; Sa,Su off; PH off'})
```
frodrigo, line 83:
```suggestion
        assert not a.node(None, {'opening_hours': 'Mo-Tu,Th-Fr 09:30-12:00; We 15:00-17:00; 2020 Dec 24,2020 Dec 31 off; Sa,Su off; PH off'})

        # Check 00:00 and 24:00 are both OK                                                                         
        assert not a.node(None, {'opening_hours': 'Mo-Su 09:00-00:00'})                                             
        assert not a.node(None, {'opening_hours': 'Mo-Su 09:00-24:00'}) 
```

frodrigo (approved):

frodrigo:
This PR is now ok for me.

@jocelynj is it ok for you ?

I think we should deploy this in the same time of switching to osmium (when ready)

dfaure:
Thanks for your reviews, for your help and for your fixes.

How long of a delay would it mean, to "wait until switching to osmium"?

As long as the current plugin is up, people are making invalid changes to the entries set by my script (https://wiki.openstreetmap.org/wiki/Import/FrenchPostOfficeOpeningHours) because oh_sanitizer makes invalid suggestions when encountering "Dec 24 08:00-12:00". Apparently people apply the change blindly even if it's incorrect, and then I'll have to manually revert all those "fixes" if I want those entries to keep being updated by my script.
This is because of the logic "only update entries if they still have the hours set by the script in the past", which was implemented in order to give priority to real people looking at real buildings, in the unlikely case that the imported data was wrong. But people clicking OK in osmose are not in that category...

This is the overall motivation for my work on this, to stop people from damaging the data I'm importing :)
The parser bug is very old, but I suspect that single-day exceptions weren't as common before my import, which is why this has become much more urgent.
Thanks for considering a solution that doesn't incur too much delay before deployment.

frodrigo:
The switch from osmosis to osmium is ready and optional at runtime. But like your changes, we have to ask to external backend owner to pay attention to this update and install required dependencies.

jocelynj:
I'm installing PyKOpeningHours version 1.3.6 on all backends handled by OSM-Fr. In my opinion, we can merge this PR to master, and inform to install the new package separately than the update to osmium (which I think need a bit more checks on OSM-Fr backends)

jocelynj (approved):

frodrigo:
Merged. Thank you all for this.

frodrigo:
Deployed on most backend server. The others need manual setup by the owners, but will not use the old parser neither.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?item=3250&class=32501
http://osmose.openstreetmap.fr/en/errors/graph.png?item=3250&class=32501
http://osmose.openstreetmap.fr/en/errors/graph.png?item=3250&class=32501&start_date=2020-10

HubMiner:
Greeting!
@frodrigo, is the spike in the graph due false positive reports being reset or improved detection in the new parser?

For comparison, non Osmose graph: https://openingh.openstreetmap.de/stats/

frodrigo:
On osmose, it is the graph of detected issues. Yes the last skipe is due to the parser changing.

dfaure:
Some of those are false positives, https://github.com/osm-fr/osmose-backend/pull/1066 fixes that.

PR #1066, opened 2020-12-25
PyKOpeningHours was integrated into KOpeningHours

dfaure:
at the request/suggestion of the KOpeningHours maintainer.

This brings in some bugfixes as well based on the first
set of results from osmose.

This brings in extra-cmake-modules as a direct dependency
(PyKOpeningHours was previously fetching that on its own,
but this makes less sense for KOpeningHours itself).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ transporthours
 pyproj >= 2.1.0
 Unidecode
 osmium
-git+git://github.com/dfaure/PyKOpeningHours.git@v1.3.6
+git+https://invent.kde.org/libraries/kopeninghours.git@83f1435
 libarchive
 
 # Tests
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,9 +26,11 @@ jobs:
           protobuf-compiler \
           osmosis \
           cmake \
+          extra-cmake-modules \
           qtbase5-dev \
           flex \
           bison
+
     - uses: actions/checkout@v2
 
     - name: Setup env
@@ -38,6 +40,7 @@ jobs:
         chown $USER /data/work/$USER /data/work/$USER/cache
         # compile osm pbf parser
         cd modules/osm_pbf_parser && make && cd ../../
+
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
@@ -90,9 +93,11 @@ jobs:
           protobuf-compiler \
           osmosis \
           cmake \
+          extra-cmake-modules \
           qtbase5-dev \
           flex \
           bison
+
     - uses: actions/checkout@v2
 
     - name: Setup env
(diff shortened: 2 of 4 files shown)


dfaure:
NameError: name 'OpeningHours' is not defined

@frodrigo Could I ask for your help one more time?
It works locally in docker so I don't understand the CI failure. Or maybe I'm not using docker properly...
I don't really understand enough about import statements in python to find out if
`from PyKOpeningHours.PyKOpeningHours import OpeningHours` is correct or not (it seems weird...).

frodrigo:
I see you try to merge into KOpeningHours. I'm not sure it is a good idea. But, keep the same layout as the python module, into the root dir, or just copy the whole module as directory into KOpeningHours.
It look like it can be installed from a directory https://stackoverflow.com/questions/13566200/how-can-i-install-from-a-git-subdirectory-with-pip

dfaure-kdab:
I did keep the same layout as the python module, and `pip install .` works fine in KOpeningHours.

frodrigo:
Also work for me, but I does not see what go wrong on travis. I test it on the alternative gihub actions CI and work fine (#1058).

dfaure:
(updated kopeninghours to include a 4th fix for issues detected in osmose)

frodrigo:
Please can you rebase. I merge the new Github Action CI.
If ok with it I will merge this PR.

dfaure-kdab:
Can you install `extra-cmake-modules` on the CI?

requirements.txt
```diff
@@ -9,7 +9,8 @@ transporthours
 pyproj >= 2.1.0
 Unidecode
 osmium
-git+git://github.com/dfaure/PyKOpeningHours.git@v1.3.6
+extra-cmake-modules
```
frodrigo, line 12:
```suggestion
extra-cmake-modules
```

frodrigo: extra-cmake-modules not a pyhton module

frodrigo, line 12:
```suggestion
```

dfaure-kdab:
Right. So the question is where to specify non-python dependencies :-)

frodrigo:
Pick this commit :
https://github.com/frodrigo/osmose-backend/commit/d4a5a122518743aa3a7c31bc500352ae22228115
https://github.com/frodrigo/osmose-backend/actions/runs/450574528

dfaure-kdab:
That's what I had initially (adding the line to requirements.txt was a bad attempt at fixing that).
So we're back to the initial error from this morning: extra-cmake-modules not found.

frodrigo:
Merged. Thank.
We will update the servers in the next days.

frodrigo:
Deployed for few days

http://osmose.openstreetmap.fr/en/errors/graph.png?item=3250&class=32501&start_date=2020-10
http://osmose.openstreetmap.fr/en/errors/?item=3250&class=32501

dfaure-kdab:
OK, thanks. The trend going down is partly the KOpeningHours fixes, but also my work fixing opening_hours warnings (with https://github.com/dfaure/OpeningHoursFixer because the Osmose GUI is too slow to my taste for such large-volume changes).

PR #1073, opened 2020-12-30
filter out some acceptable combinations

matkoniecz:
and `natural=bird_nest` appears due to stork nests such as https://commons.wikimedia.org/wiki/File:Ciconia_ciconia_nest_Wola_Batorska_Poland-00-2014-06-26_08-54-03.jpg


matkoniecz:
Not sure what I am doing wrong, but I again failed to run tests ( https://github.com/osm-fr/osmose-backend/issues/1074 )

frodrigo:
The test is already broken on master. We need to fix this.

Thank you.

PR #1101, opened 2021-01-13
adjust Indian territories after merge

stephankn:
Indian territories had been merged. See https://en.wikipedia.org/wiki/Dadra_and_Nagar_Haveli_and_Daman_and_Diu

OSM relation 1952530 contains new boundaries. Adjusting config.


stephankn:
@jocelynj Does this also need some changes on front-end to cater for the change in territory name and to remove the obsolete one?

frodrigo:
We never face this issue. I already see this issue, but do not fix it yet.
We must produce a new dedicated extract first.
We cannot merge territories on the front. We have to declare a new one.

stephankn:
In OSM the territories had been merged already (since long). relation 1952530 existed before with a different name and contains now the additional areas which had been "Daman and Diu" before. The obsolete relation was removed from OSM and fails now the analyzers.

IMHO task is to remove "Daman and Diu" from configuration and rename "Dadra and Nagar Haveli" into "Dadra and Nagar Haveli and Daman and Diu".

If the extract generation doesn't use life relation boundaries from OSM, then these need to be updated. The ID of the relation is unchanged, but got larger.

jocelynj:
Do we need to switch to the new states?

We could keep the old one, at least if the old boundaries are still available on OSM. And I would expect to get the old boundaries, at least for historical reasons.

If we want to switch to the new state, we need to:
  - generate the new extract and diffs on download.openstreetmap.fr
  - update the frontend with the new state (will be done "automatically" when this change will be merged)
  - update backend list of countries to run
  - remove results from the previous states on the frontend

frodrigo:
No. One old boundary is missing. I already write you a mail about this.

stephankn:
@jocelynj I run analyzers for India and investigated why a specific one was always failing. This was due to the removal of the no longer existing administrative relation in OSM. I can adjust the analyzer task within minutes.

Extract generation might be extra work. I do not know how you get the bounding polygons. As this change just makes an already existing administrative reagion larger, it is hopefully simply done with updating the bounding polygon.

frodrigo:
@jocelynj extracts was re-generated today, but not include the new Indian one.

jocelynj:
The extract and diff are now available: http://download.openstreetmap.fr/extracts/asia/india/

stephankn:
[20210120_093433-india_dadra_and_nagar_haveli_and_daman_and_diu.log](https://github.com/osm-fr/osmose-backend/files/5841915/20210120_093433-india_dadra_and_nagar_haveli_and_daman_and_diu.log)

Analysis seems to work fine, but Frontend answers for everything with status 404. See attached log.

Something needed to configure on front-end in addition?

stephankn:
I did a second run which no longer gets 404 from front-end. Maybe it was the first run ever which checked last update date? Seems to be all fine now

frodrigo:
Yes. First run cannot get last update from frontend.

PR #1108, opened 2021-01-19
Feature/parking lane checks

### LINKED ISSUE #1089: parking:lane:<value>=separate is not recognized
According to the osm wiki - `parking:lane:<side>=separate` is used on a highway to indicate that the parking information is mapped separately (see https://wiki.openstreetmap.org/wiki/Tag:parking%3Dstreet_side). This is not recognized by Osmose.

Item: 3161
Class: 31615
Message: Bad parking:lane:[side] value

jdhoek:
The first commit is fairly trivial and fixes #1089.

The second commit expands the `parking:lane` checks with two errors for cases where `parking:condition:<side>` is set with an incompatible `parking:lane:<side>` value. I.e., the values of `no`, `no_stopping`, `no_parking`, `fire_lane`, and `separate` should not have any condition set because no parking is allowed, or (in the case of `separate`) parking is mapped separately and conditions should be mapped there.

The test runs.

### DIFF
--- a/plugins/Highway_Parking_Lane.py
+++ b/plugins/Highway_Parking_Lane.py
@@ -72,7 +72,7 @@ sides.'''))
             err.append({"class": 31614})
 
         for side in ("parking:lane:right", "parking:lane:left", "parking:lane:both"):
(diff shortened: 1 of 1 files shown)


frodrigo:
Looks good. Thank you.

PR #1110, opened 2021-01-19
Treat building:part as top-level tag

### LINKED ISSUE #1109: `building:part` is not treated as a top-level tag
I'm seeing **Untagged named object** (item 2110) being flagged for ways tagged with `name` and `building:part`.

This doesn't seem right. `building:part` looks like a top-level tag, and building parts being named is not uncommon (e.g., a wing of a hospital).

jdhoek:
Fixes #1109.


frodrigo:
Thank you.

PR #1114, opened 2021-01-21
Analyser_Merge: Replace access=public with access=yes

gileri:
[access=yes is recommended over acces=yes](https://wiki.openstreetmap.org/wiki/Key:access#Possible_Tagging_Mistakes)

[It's usage is also declining](https://taginfo.openstreetmap.org/tags/access=public#chronology)


frodrigo:
cc @PanierAvide

PanierAvide:
That makes sense, thanks for offering the PR :smiley:

frodrigo:
Thank you all

gileri:
Thank you for the quick merge !

PR #1122, opened 2021-01-24
Fix JSON object

soulei-seg:
Set flashing_lights' property to null instead of "None".


frodrigo:
Note, you can repush on the same branch to fix the PR.

Issue #1123, opened 2021-01-24
Add detection of oneways


Famlam:
Mapillary contains about 200000 cases of the "oneway, no entry from this side" (`regulatory--no-entry--g1`) sign in NL, of which a large amount is already mapped.
Unlike the arrow-like oneway signs, like [this one](https://traffic-icons.mapillary.com/signs/regulatory--one-way-right--g1.svg) (which contain huge amounts of false positives) the ['no entry from this side' sign](https://traffic-icons.mapillary.com/signs/regulatory--no-entry--g1.svg) is quite well detected. 

Hence I would propose to add the following rule:
```
  {
    "item": 8300,
    "class": XXXXXXXXXXXX,
    "level": 3,
    "otype": [
      "ways"
    ],
    "conflation": 200,
    "title": "one-directional roads",
    "object": [
      "regulatory--no-entry--g1"
    ],
    "select_tags": [
      {
        "highway": null,
        "oneway": null
      },
      {
        "highway": null,
        "oneway:motor_vehicle": null
      },
      {
        "highway": null,
        "oneway:vehicle": null
      },
      {
        "highway": null,
        "oneway:motorcar": null
      },
      {
        "highway": null,
        "oneway:hgv": null
      },
      {
        "highway": null,
        "oneway:bus": null
      },
      {
        "highway": null,
        "oneway:foot": null
      },
      {
        "highway": null,
        "oneway:psv": null
      },
      {
        "highway": null,
        "oneway:bicycle": null
      },
      {
        "highway": null,
        "oneway:moped": null
      },
      {
        "highway": null,
        "vehicle:backward": null
      },
      {
        "highway": null,
        "motor_vehicle:backward": null
      },
      {
        "highway": null,
        "hgv:backward": null
      },
      {
        "highway": null,
        "hazmat:backward": null
      },
      {
        "highway": null,
        "bicycle:backward": null
      },
      {
        "highway": null,
        "access:backward": null
      }
    ],
    "generate_tags": {
      "oneway": "yes"
    }
  },
```

I don't know how well oneway ways are mapped in the remaining part of the world, so in case it is causing excessive messages, perhaps the following should be added - similar to the stop sign detection in Canada:
```
    "only_for": [
      "NL"
    ],
```

**Justification of the tags I picked (with XXXXX = a vehicle type) :**
[...]

PR #1124, opened 2021-01-24
Update mapillary-traffic-signs.mapping.json

Famlam:
Updates issue #1123 
Adds Mapillary-data based oneway detection (for now just for NL as trial)


frodrigo:
Merged. Not deployed yet.

Issue #1125, opened 2021-01-25
Assume that barrier=toll_booth includes street light


matkoniecz:
Yes, toll_booth has built-in traffic lights. No, it is not needed or useful to map it separately.

Marking false positives on every Mapillary image gathering going through any toll booth is waste of time, it should be automatically hidden.

https://www.openstreetmap.org/node/5978083463

http://osmose.openstreetmap.fr/en/map/#zoom=18&lat=50.082958&lon=19.792808&item=xxxx&level=1&tags=&fixable=&issue_uuid=4a0acba8-ef63-2fff-001b-acc0f03de73d

http://osmose.openstreetmap.fr/en/error/4a0acba8-ef63-2fff-001b-acc0f03de73d

https://www.mapillary.com/app/?focus=photo&pKey=XLUPbXGgAGPlnjjDFe5cLQ&signs=true&points=true&lat=50.082981108108115&lng=19.79130964864865&z=17

![screen10](https://user-images.githubusercontent.com/899988/105695248-2b8a7680-5f02-11eb-8aab-bc63eb597355.png)

![screen09](https://user-images.githubusercontent.com/899988/105694934-c767b280-5f01-11eb-80df-2ba1f10de8f2.png)

frodrigo:
cc @soulei-seg, if you interested to fix it, same thing.

soulei-seg:
I will be working on the issue soon

soulei-seg:
Not sure but I would suppose it's like this way :
![image](https://user-images.githubusercontent.com/31833254/105742089-5ba23b80-5f3b-11eb-946e-55bcf982b6c5.png)

frodrigo:
@soulei-seg only `barrier=toll_booth` https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dtoll_booth
Please, post code as markdown, not image.

soulei-seg:
Sure, I'll remember that for the next time.

frodrigo:
Deployed.

PR #1126, opened 2021-01-25
Fix toll_booth

soulei-seg:
Added new **barrier** tag into mapillary-street-objects.mapping.json to fix the issue #1125


frodrigo:
Thank you.

PR #1131, opened 2021-01-27
Update postgis to 3.1

stephankn:
Verified updated version with Bhutan analysis.
Expectation is to benefit from performance improvements, see
http://blog.cleverelephant.ca/2020/12/waiting-postgis-31-1.html

I grepped the source and none of the deprecated/changed methods are used.
https://git.osgeo.org/gitea/postgis/postgis/raw/tag/3.1.0/NEWS


frodrigo:
Ok. Thank you.

PR #1137, opened 2021-01-27
ignore analysers/disabled in mypy

stephankn:
all recent PR have failed validators due to mypy failing over things in disabled.
let's try this option, based on https://mypy.readthedocs.io/en/latest/config_file.html


stephankn:
no, doesn't work :(

    analysers/disabled/analyser_merge_wikipedia_FR.py:23: error: No parent module -- cannot perform relative import

not certain what mypy needs. maybe it needs an `__init__.py` in disabled?

stephankn:
ah, or does it maybe not use the mypy.ini from this commit during the lint run? I wanted to avoid installing mypy locally... Maybe have to . These false reports of linter annoy me...

stephankn:
fix #1138 unfortunately this exposed a larger amount of problems previously not analyzed.

stephankn:
I think this can be merged now. It improved the mypy lint situation significantly. The remaining issues need to be checked. They could be real issues. I assume that imports are not harmed by the new `__init__.py`. Not certain how to verify. Automated tests at least still pass.

frodrigo:
Ok. Thank you a lot for working on this.

PR #1151, opened 2021-02-01
allow markdown parsing

matkoniecz:
right now markdown is displayed as raw code, this should fix it (if frontend renders markdown and there are no other syntax issues)

screen01


frodrigo:
Well spoted, thank you.

PR #1164, opened 2021-02-06
Opendata source

frodrigo:
Use more Source with millesime from API

cc @gileri


gileri:
I meant to review this PR once it exited draft status, oh well :sweat_smile: 

Thank you for finding about the `data.public.lu` compatibility with `SourceDataGouv` !

frodrigo:
Sorry for the quick move. Not to late to review. Still not deployed.

osmose_config.py
```diff
@@ -284,11 +284,9 @@ class gen(default_country):
     # Bordeaux
     'merge_recycling_FR_bm',
     'merge_parking_FR_bm',
-    'merge_bicycle_parking_FR_bordeaux',
```
gileri, line 287: Even if the OpenData page is not available anymore, the source file [still is](http://opendatabdx.cloudapp.net/DataBrowser/DownloadCsv?container=databordeaux&entitySet=sigstavelo&filter=NOFILTER). Wouldn't it be better to keep it until the source file is removed ?

frodrigo: You are right. The are data on Bordeaux Metropole.
https://opendata.bordeaux-metropole.fr/explore/dataset/st_arceau_p/information/
But I misinterpret the role of merge_parking_FR_bm with is only for car.
We need to add a new analyser for Bordeaux Metropole in place.

frodrigo: #1170

Issue #1169, opened 2021-02-08
Enhance 3130 (Plugin TagFix_Brand.py) : suggest the brand:wikidata tag when only the brand tag is on the node


nuxper:
I'll use https://www.openstreetmap.org/node/6974650360 as example
This node represents a Promovacances travel_agency in Lyon.

Current tags in osm:
```
brand = Promovacances
name = Promovacances
shop = travel_agency
```

Promovacances is part of the name-suggestion-index : promovacances-9ff76a with 
```
    {
      "displayName": "Promovacances",
      "id": "promovacances-9ff76a",
      "locationSet": {"include": ["fr"]},
      "tags": {
        "brand": "Promovacances",
        "brand:wikidata": "Q3407444",
        "brand:wikipedia": "fr:Promovacances",
        "name": "Promovacances",
        "shop": "travel_agency"
      }
    }
```

Code of the plugin : https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagFix_Brand.py

### Current behavior

The plugin doesn't suggest adding brand:wikidata=Q3407444 because it only selects nodes where the tag "brand" is absent

### Suggested behavior

I suggest to select nodes where the tag "brand" is absent or the tag "brand:wikidata" is absent

I actually see two options:
1 select nodes where the tag "brand" is absent or the tag "brand:wikidata" is absent:
- Pros: less nodes to be processed
- Cons : won't suggest brand:wikipedia or additional secondary tags (clothes= , shoes=, ...)

2 select nodes with tag name and with tag amenity or shop without checking for brand or brand:wikidata

**I can work on the change myself then submit a pull request but would like your feedback before I spend time on it**

frodrigo:
We should not override the existing brand. Maybe look for `brand:wikidata` from `brand`.
Can also look how NSI is used elsewhere.

> I can work on the change myself then submit a pull request but would like your feedback before I spend time on it

Please do.

nuxper:
From what I understand of the code, the plugin only suggests adding tags, never override.

OK I'll work on this.

frodrigo:
cc @nlehuby @Marc-marc-marc

nlehuby:
Yes, that would be a good addition :+1: 
Your first option is my preference to limit false positives.

frodrigo:
It is deployed, but your changes lower the number of raised issues by 100 000.
http://osmose.openstreetmap.fr/en/errors/?item=3130
[...]

PR #1171, opened 2021-02-10
solves #1169 : suggest the brand:wikidata tag when only the brand tag is on the node

nuxper:
I checked it on Lyon and it works well
I updated the test cases to match the new logic


frodrigo:
Look fine, thank you.
Merged, but not deployed yet.

PR #1173, opened 2021-02-10
Stop removing last char of stop_id

alkino:
It seems to come from an old time, cannot find why in history. Sources have changed a lot of time.


alkino:
https://wiki.openstreetmap.org/wiki/Key:uic_ref

`The UIC station reference always consists of 7 digits`

Issue #1174, opened 2021-02-11
Add integration of operator:wikidata from name-suggestion-index


nuxper:
name-suggestion-index added a list of operators and the corresponding operator:wikidata
This will be integrated in future version of id but the list already exists.

I suggest to add the detection of operator matching an entry in name-suggestion-index

FOREACH node
   IF node has the "operator" tag but no "operator:wikidata" tag 
        IF "operator" is in name-suggestion-index
            suggest adding the operator:wikidata from name-suggestion-index
      
I can work on this but would like your feedback first:
* Do you agree with the idea ?
* Should I create a new plugin or complete the TagFix_Brand.py or create a TagFix_Operator.py ?
* If I should create a new plugin, what item class id should I use ?

frodrigo:
I am ok with it.
I see no need to create a new plugin. Just a new class in it under the same item.

cc @nlehuby

nuxper:
ok for me
so item 3130 class 31302 ? I'm not completely clear on the item / class / subclass notions ;)
do I need to register the class to the frontend somehow ?

frodrigo:
Yes this class is free 
http://osmose.openstreetmap.fr/en/errors/?item=3130&class=31302

FYI https://github.com/osm-fr/osmose-backend/blob/master/doc/0-Index.md#concepts-of-osmose-qa-issue

frodrigo:
It is the goal to have all this tags? That a lot!

![imagen](https://user-images.githubusercontent.com/1785486/108760093-80df9500-754d-11eb-87f1-9b1671790c1d.png)

nuxper:
Yes I'm a bit hesitant on the subject too.
I see the same problem with post_boxes.
The problem is more with the name-suggestion-index than the plugin I believe. I think atms and post_boxes shouldn't include that many tags.
My point of view: the information is actually accurate. Those tag values are correct and legitimate. But does the community want to have all those tags on all the atms and post_boxes ? not sure.
I'll open an issue on the name-suggestion-index on the matter to discuss and link it here. Worst case I could limit the plugin to only add operator:wikidata but I think it would be fixing the problem in the wrong place.

frodrigo:
I got this stange suggestion

http://osmose.openstreetmap.fr/en/map/#zoom=18&lat=44.81259&lon=-0.590057&item=xxxx&level=1%2C2%2C3&tags=&fixable=josm&issue_uuid=c70da878-6848-8afe-ed37-099cebead1b7

[...]

PR #1183, opened 2021-02-17
Added integration of operator:wikidata from name-suggestion-index. Solves #1174

nuxper:
Note : I didn't change the name of the whole class and file as I'm not sure of the consequences. I could change them to TagFix_Nsi.py if you prefer but I'm not sure of the consequences ?

Note : I tested this on the Rhône and it seems ok.


frodrigo:
Look good to me.

frodrigo:
Thank you.

Issue #1182, opened 2021-02-17
Generalize fix:* tags in analysers


PanierAvide:
I see that in http://osmose.openstreetmap.fr/api/0.3/tags we have available three `fix:*` tags : chair, imagery, survey. To make issues more easily reusable by third party tooling, would it make sense to :
- Introduce a `fix:picture` tag for issues that can be fixed with street-level pictures ?
- Generalize the presence of these tags in all analysers to have a precise idea for every analyser on how it can be fixed ?
- Document these tags so they can be used by developers working on analysers more easily ?
- (on frontend side) Display all these fix tags values in UI ?

I would like to spend some time on this task if there isn't any particular problem with it, and if you don't see any technical issue that would require major architecture change.

frodrigo:
* `fix:picture` does not have almost the same meaning as `fix:survey`?
* Generalize the presence of these tags. -> I agree, but lot of works.
* Display all these fix tags values in UI: already show

Yes, you can work on this. About the frontend UI, note, switching in progress to Vue js.

cc @Marc-marc-marc

PanierAvide:
Maybe it's a bit arbitrary, but I think it might make sense to distinguish picture/survey, examples :
- Street numbers on houses : they are rarely correctly visible from street pictures, needs survey
- Fire hydrants : can be seen correctly on street pictures and by survey
- Destination signs on motorways : quite dangerous or difficult to survey without pictures

By reading that all tags are already shown, I just realized that the "Avec correction" select field isn't about `fix:*` tags, but issues list of fixes... So UI already handles this, it's just not really clear.

frodrigo:
Deployed

PR #1185, opened 2021-02-18
Add fix:* tags on various analysers (see #1182)

PanierAvide:
Added fix:* tags which seemed logical according to the kind of analyser.

### DIFF
--- a/analysers/analyser_merge_street_objects.py
+++ b/analysers/analyser_merge_street_objects.py
@@ -43,7 +43,7 @@ class Analyser_Merge_Street_Objects(Analyser_Merge_Dynamic):
 class SubAnalyser_Merge_Street_Objects(SubAnalyser_Merge_Dynamic):
     def __init__(self, config, error_file, logger, classs, level, otype, conflation, title, object, selectTags, generateTags, mapping, source, layer):
         SubAnalyser_Merge_Dynamic.__init__(self, config, error_file, logger)
-        self.def_class_missing_official(item = 8360, id = classs, level = level, tags = ['merge', 'leisure'],
+        self.def_class_missing_official(item = 8360, id = classs, level = level, tags = ['merge', 'leisure', 'fix:survey', 'fix:picture'],
             title = T_('Unmapped {0}', T_(title)),
             detail = T_('Street object ({1}) detected by Mapillary, but no nearby "{0}" tagging.',
                 ', '.join(map(lambda kv: '{0}={1}'.format(kv[0], kv[1] if kv[1] else '*'), generateTags.items())), T_(title)),
--- a/analysers/analyser_merge_traffic_signs.py
+++ b/analysers/analyser_merge_traffic_signs.py
@@ -79,7 +79,7 @@ class SubAnalyser_Merge_Traffic_Signs(SubAnalyser_Merge_Dynamic):
                 ['`{}={}`'.format(kv[0], kv[1] if kv[1] else '*') for kv in selection.items()]
             ))
 
-        self.def_class_missing_official(item = item, id = classs, level = level, tags = ['merge'] + tags,
+        self.def_class_missing_official(item = item, id = classs, level = level, tags = ['merge', 'highway', 'fix:picture', 'fix:survey'] + tags,
             title = T_('Unmapped {0}', T_(title)),
             detail = T_('Traffic sign ({1}) detected by Mapillary, but no nearby tagging of any:{0}', '\n\n- ' + '\n- '.join(missing_tags), T_(title)),
             fix = T_('Add the appropriate highway tagging if the imagery is up-to-date and sign detection is correct.'))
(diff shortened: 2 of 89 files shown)


frodrigo:
Thank you, it is a lot of work.
Please can you clarify the meaning of `fix:picture`?

PanierAvide:
`fix;picture` is for when this can be solved using street-level pictures (mapillary, openstreetcam)

analysers/analyser_merge_traffic_signs.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, config, error_file, logger, item, classs, level, tags, otype,
                 ['`{}={}`'.format(kv[0], kv[1] if kv[1] else '*') for kv in selection.items()]
             ))
 
-        self.def_class_missing_official(item = item, id = classs, level = level, tags = ['merge'] + tags,
+        self.def_class_missing_official(item = item, id = classs, level = level, tags = ['merge', 'highway', 'fix:picture'] + tags,
```
Marc-marc-marc, line 82: also fix:survey

analysers/analyser_merge_water_FR.py
```diff
@@ -119,7 +119,7 @@ def __init__(self, config, logger = None):
 class SubAnalyser_Merge_Water_FR(SubAnalyser_Merge_Dynamic):
     def __init__(self, config, error_file, logger, item, id, level, title, select, tags, generate_tags):
         SubAnalyser_Merge_Dynamic.__init__(self, config, error_file, logger)
-        self.def_class_missing_official(item = item, id = id, level = level, tags = ['merge'], title = title)
+        self.def_class_missing_official(item = item, id = id, level = level, tags = ['merge', 'water', 'fix:survey', 'fix:imagery'], title = title)
```
Marc-marc-marc, line 122: fountain also with fix:chair

analysers/analyser_merge_street_objects.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, config, logger = None):
 class SubAnalyser_Merge_Street_Objects(SubAnalyser_Merge_Dynamic):
     def __init__(self, config, error_file, logger, classs, level, otype, conflation, title, object, selectTags, generateTags, mapping, source, layer):
         SubAnalyser_Merge_Dynamic.__init__(self, config, error_file, logger)
-        self.def_class_missing_official(item = 8360, id = classs, level = level, tags = ['merge', 'leisure'],
+        self.def_class_missing_official(item = 8360, id = classs, level = level, tags = ['merge', 'leisure', 'fix:survey', 'fix:picture'],
```
Marc-marc-marc, line 46: osmose produce a merge, so mapillary found it on a street level imagery. so imho fix:chair is also doable

Marc-marc-marc (approved):
I also don't see the difference between fix:picture and fix:survey (depending on the pictures, one picture is or isn't enough to find the information. this makes it impossible for me to say "a survey is always needed and not picture").
but I have nothing against this addition, the whole PR is anyway a big improvement!but it doesn't bother me to see it added, the whole thing is a great improvement over the current situation anyway.

I have posted a few comments, but which do not deserve to block the adoption of this PR

PanierAvide:
Thanks for your changes suggestions, I will integrate them soon.

Please read here [my comment here](https://github.com/osm-fr/osmose-backend/issues/1182#issuecomment-780429801) about how I see `fix:picture` and `fix:survey` differently. There is no 1-1 match between street-level pictures and aerial imagery, and ground survey might not always be possible (think of motorways, where "manual" survey is quite impossible whereas taking pictures is easy). As a data reuser and developer, I see this as an interesting tag to put on issues. As it doesn't replace any existing feature, it's just another information. Everyone would be free to use or not use it.

PanierAvide:
Let's take another example : if we had an analyser about tree diameters (centimeter precision), it would be fixable by survey. But no way to check that info using a picture or imagery. All `fix:picture` will be associated to `fix:survey`, but not all `fix:survey` could be solved using pictures.

analysers/analyser_merge_water_FR.py
PanierAvide: I'm not sure if `fix:chair` makes sense here, you will either need an imagery or to go on ground to integrate these features. Isn't `fix:chair` about things you can fix by just checking Osmose information (refs, topology issues, knowledge from your brain...) ?

analysers/analyser_merge_street_objects.py
PanierAvide: Same as above

analysers/analyser_merge_traffic_signs.py
PanierAvide: Added, thanks :+1:

frodrigo:
Thank you a lot for all this work.

Issue #1165, opened 2021-02-06
merge_public_equipment_FR_lyon_toilets source nor more available


frodrigo:
The data source for city of Lyon is no more available. But a new one the the Métropole is.
https://data.grandlyon.com/jeux-de-donnees/toilettes-publiques-metropole-lyon-v2-0-0/telechargements
analyser_merge_public_equipment_FR_lyon_toilets.py

@PanierAvide interested in fixing this?

frodrigo:
Deployed
http://osmose.openstreetmap.fr/fr/errors/?item=8180,8181,8182&class=33,34,35&useDevItem=all

@jocelynj 
Item 8181,8182 to be activated

frodrigo:
Enabled.

PR #1193, opened 2021-02-24
Update merge_defibrillators_FR

PanierAvide:
Various improvements on GeoDAE defibrillators merge analyser :
- Use GeoJSON as source format (CSV was causing errors)
- Re-introduce `ref:FR:GeoDAE` as ID was stabilized on GeoDAE dataset (and confirmed by a check I've made at 2 weeks interval)
- Moved annex tags in mapping2 (by the way, [doc could be clarified](https://github.com/osm-fr/osmose-backend/blob/master/doc/4-Merge.md#mapping) on the true reason of separating mapping1 and 2, I'm not sure myself what it changes : mapping2 tags changes are not considered for updates selection ?)

### DIFF
--- a/analysers/analyser_merge_defibrillators_FR.py
+++ b/analysers/analyser_merge_defibrillators_FR.py
@@ -21,7 +21,7 @@
 ###########################################################################
 
 from modules.OsmoseTranslation import T_
-from .Analyser_Merge import Analyser_Merge, Source, CSV, Load, Conflate, Select, Mapping
+from .Analyser_Merge import Analyser_Merge, Source, GeoJSON, Load, Conflate, Select, Mapping
 import unidecode
 import re
 from modules import reaccentue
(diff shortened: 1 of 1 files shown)


frodrigo:
>     * Moved annex tags in mapping2 (by the way, [doc could be clarified](https://github.com/osm-fr/osmose-backend/blob/master/doc/4-Merge.md#mapping) on the true reason of separating mapping1 and 2, I'm not sure myself what it changes : mapping2 tags changes are not considered for updates selection ?)

Yes, that the only point, change in mapping2 tags are not trigger an update suggestion. While there are still included in creation and update.

frodrigo:
I see it, maybe not on all yet. We deploy yesterday.
http://osmose.openstreetmap.fr/fr/error/e2ab3b70-c023-412a-c017-0c82ac289d10

PanierAvide:
Oh OK, I thought this was deployed before yesterday so that's why I was worried for it not being present. And what about items 8371 / 8372 ? Issues will also arrive in the following days ?

frodrigo:
Still to be enabled https://github.com/osm-fr/osmose-backend/issues/1165#issuecomment-776919236

PR #1195, opened 2021-02-24
Do not display unnecessary error message

archie83:
When osmose_run.py is run with the "--skip-upload", it shouldn't display an error message about missing password for upload.

Note: This is not a big fix for Osmose, but this is my first pull request on GitHub. I need to understand the development flow and get familiar with it.


frodrigo:
Thank you.

PR #1197, opened 2021-02-26
Dynamic millesime for merge analysers using IGN BDTOPO

archie83:
Currently, the millesime of all the merge analysers using teh IGN BDTOPO files is not up-to-date (2020-09 instead of 2020-12). So I have implemented the extraction of the millesime from the MetaData of the GPKG files.

It is not well-optimized, since the GPKG file is read twice, firstly to extract the millesime, secondly to import the data into the database. Is this acceptable?

### DIFF
--- a/analysers/analyser_merge_cemetery_FR.py
+++ b/analysers/analyser_merge_cemetery_FR.py
@@ -21,7 +21,7 @@
 ###########################################################################
 
 from modules.OsmoseTranslation import T_
-from .Analyser_Merge import Analyser_Merge, Source, GPKG, LoadGeomCentroid, Conflate, Select, Mapping
+from .Analyser_Merge import Analyser_Merge, SourceIGN, GPKG, LoadGeomCentroid, Conflate, Select, Mapping
 
 
 class Analyser_Merge_Public_Cemetery_FR(Analyser_Merge):
@@ -33,7 +33,7 @@ class Analyser_Merge_Public_Cemetery_FR(Analyser_Merge):
         self.init(
             "https://ign.fr",
             "IGN-Cimetière",
-            GPKG(Source(attribution = "IGN", millesime = "09/2020", gzip = True,
+            GPKG(SourceIGN(attribution = "IGN", gzip = True,
                     fileUrl = "http://files.opendatarchives.fr/professionnels.ign.fr/bdtopo/latest/geopackage/cimetiere.gpkg.gz")),
             LoadGeomCentroid(
                 select = {"etat_de_l_objet": "En service"}),
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -487,6 +487,12 @@ class SourceHttpLastModified(Source):
             downloader.HTTP_DATE_FMT,
         )
 
+class SourceIGN(Source):
+    """Get millesime from IGN BDTOPO MetaData"""
+    def get_millesime(self) -> datetime.datetime:
+        response = downloader.get("https://files.opendatarchives.fr/professionnels.ign.fr/bdtopo/latest/geopackage/meta.json")
+        response.raise_for_status()
+        return datetime.datetime.fromisoformat(response.json()["millesime"])
 
 class Parser:
     def header(self):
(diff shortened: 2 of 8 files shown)


frodrigo:
Using `vsigzip` org can read directly.
But I unsure about the global idea. Cannot just use `SourceHttpLastModified` in this case?

archie83:
OK, I'll have a look.

archie83:
I don't think we can use  `SourceHttpLastModified`:

```
$ curl -s -v -X HEAD http://files.opendatarchives.fr/professionnels.ign.fr/bdtopo/latest/geopackage/construction_ponctuelle.gpkg.gz
*   Trying 212.129.55.232:80...
* Connected to files.opendatarchives.fr (212.129.55.232) port 80 (#0)
> HEAD /professionnels.ign.fr/bdtopo/latest/geopackage/construction_ponctuelle.gpkg.gz HTTP/1.1
> Host: files.opendatarchives.fr
> User-Agent: curl/7.74.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: nginx
< Date: Fri, 26 Feb 2021 20:33:34 GMT
< Content-Type: application/octet-stream
< Content-Length: 16768208
< Connection: keep-alive
< Last-Modified: Sun, 17 Jan 2021 23:13:29 GMT
< ETag: "6004c499-ffdcd0"
< Accept-Ranges: bytes
< Strict-Transport-Security: max-age=15552000; includeSubdomains; preload

```
The date given by `SourceHttpLastModified` (17/01/2021) is not the same as the BDTOPO date (15/12/2020).

frodrigo:
I have asked to @cquest to add a meta info file in the data repository.

archie83:
Did you get an answer from Christian?
I don't understand why you don't want to use the metadata that are in the GPKG files. Adding separate files for the metadata would be redundant and error prone.

frodrigo:
The GPKG format is not the original one and come from a conversion. I does not know how the information is set.
In you implementation it require an extra unzip (at least it require using vsigzip)

archie83:
I agree, but since here is a cache mechanism, I suppose the unzipping will be done only when there is a new release of the GPKG file? So it is not a big overhead.

frodrigo:
No. here only the download will be cached, the milesime read will be done at each run.

frodrigo:
The is now here http://files.opendatarchives.fr/professionnels.ign.fr/bdtopo/latest/geopackage/tree.json

archie83:
The dates contained in _tree.json_ are the dates of the GPKG files (1/5/2021) and not the dates of the BDTOPO (15/3/2021). Therefore it is not possible to use them for the dynamic millesime.

archie83:
I have updated the patch to include analyser_merge_natural_FR.

cquest:
I've added meta.json file which is much simpler to use:

https://files.opendatarchives.fr/professionnels.ign.fr/bdtopo/latest/geopackage/meta.json

archie83:
Thank you. I'll update my patch accordingly.

archie83:
New patch updated, it runs much more faster than the first one, thanks to the meta.json file added by Christian.

analysers/analyser_merge_cemetery_FR.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, config, logger = None):
         self.init(
             "https://ign.fr",
             "IGN-Cimetière",
-            GPKG(Source(attribution = "IGN", millesime = "09/2020", gzip = True,
+            GPKG(SourceIGN(attribution = "IGN", gzip = True,
```
frodrigo, line 36: It is required for insert the millesime into the attribution string
https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Merge.py#L424

```
attribution = "IGN, {0}",
```

archie83: I tried your proposal, the effect is that it changes the format of the resulting XML file: 
```
-<tag action="create" k="source" v="IGN - 2021-06" />
+<tag action="create" k="source" v="IGN, 2021-06" />
```

So, what is the preferred version?

frodrigo: Hum, may be I'm wrong, and no change is required.

archie83: Anything else needed from my side?

frodrigo:
Thank you all.
Merged. But not deployed, yet.

frodrigo:
Deployed

Issue #1205, opened 2021-03-02
Osmose raises an error when tag housenumber is used in the member of an associatedstreet relation but also member of another relation


nuxper:
Item 2060 class 1
http://osmose.openstreetmap.fr/fr/error/4e1daae5-8324-8ef2-b0e3-f34278110a79

osmose reports an issue for a housenumber node member of:
* associatedstreet relation
* provides_feature relation

2060 reports that addr:housenumber or addr:housename without addr:street, addr:district, addr:neighbourhood, addr:quarter, addr:suburb, addr:place or addr:hamlet must be in a associatedStreet relation. It shouldn't report this issue as the node is member of an associatedStreet relation.

I believe I found the problem in the code. I'll submit a PR

frodrigo:
To be tested. but probably something like this
```sql
WHERE
     (relation_members.member_id IS NULL OR relations.id IS NOT NULL) AND
     ...
```
https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_relation_associatedStreet.py#L80

nuxper:
Hi @frodrigo ,

Thanks for your feedback.

I checked your suggestion but do not reach the expected result. Can I ask your help ?:

### Taking a step back and focusing on nodes only first:

If I understand correctly, sql01 should return the nodes which seem to be address nodes (with a addr:housenumber or addr:housename tag)

The intention is also to get the information
- whether or note the addr:street (or similar) exists
- whether the node is a member of an associatedStreet (or similar) relation

In order to do that, sql01 would left outer join the table of (nodes with a addr:housenumber) to the table of (members of an associated street relation)

### my understanding...
...is that members of an associated street relation should be an inner join:

```
relation_members
INNER JOIN relations ON
	relation_members.relation_id = relations.id AND
	relations.tags?'type' AND
	relations.tags->'type' IN ('associatedStreet', 'street')
```
	
I tested this solution which was in the PR and as far as I could tell, the fix worked as intended. I checked the differences (before->after fix) at least in Lyon.

### If I try to use a LEFT JOIN...
... between relation_members and relations then restrict it in the where clause, 

```
WHERE
     (relation_members.member_id IS NULL OR relations.id IS NOT NULL) 
```
     
then the nodes that have a addr:housenumber tag, are NOT in an associatedstreet but ARE in another type of relation are no longer in the result (example : 6440629970)

### Can you help please ? ;) :
Have I understood your suggestion correctly ?
[...]

PR #1206, opened 2021-03-02
#1205 : do not raise error when tag housenumber is used in the member of an associatedstreet relation which is also member of another relation

nuxper:
See https://github.com/osm-fr/osmose-backend/issues/1205


frodrigo:
No the right thing to do

PR #1208, opened 2021-03-05
fix typo, mention case of wrong riverbank

matkoniecz:
Is there sane method to update `po/` files? Or is it necessary to do it manually?

### DIFF
--- a/analysers/analyser_osmosis_waterway.py
+++ b/analysers/analyser_osmosis_waterway.py
@@ -149,8 +149,8 @@ class Analyser_Osmosis_Waterway(Analyser_Osmosis):
 '''There is one `waterway=riverbank` but there is no `waterway=river`
 inside it.'''),
             fix = T_(
-'''After checking, create a "river" line inside te "riverbank"
-polygon.'''))
+'''After checking, create a "river" line inside the "riverbank"
+polygon or eliminate "riverbank" polygon.'''))
         self.classs_change[2] = self.def_class(item = 1220, level = 3, tags = ['waterway', 'fix:imagery'],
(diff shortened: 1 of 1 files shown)


frodrigo:
The english version has to be fixed in the code.

matkoniecz:
> in the code.

Which one? Is my PR sufficient or should I edit also something else?

analysers/analyser_osmosis_waterway.py
```diff
@@ -149,8 +149,8 @@ def __init__(self, config, logger = None):
 '''There is one `waterway=riverbank` but there is no `waterway=river`
 inside it.'''),
             fix = T_(
-'''After checking, create a "river" line inside te "riverbank"
-polygon.'''))
+'''After checking, create a "river" line inside the "riverbank"
+polygon or eliminate "riverbank" polygon.'''))
```
frodrigo, line 153:
```suggestion
polygon or eliminate then "riverbank" polygon.'''))
```
```suggestion
polygon or eliminate "riverbank" polygon.'''))
```

frodrigo:
> > in the code.
> 
> Which one? Is my PR sufficient or should I edit also something else?

Yes, sorry I did not see at first time it is a PR and not an issue.

analysers/analyser_osmosis_waterway.py
matkoniecz: I am unsure what is being proposed here. Adding word "then" and removing it?

frodrigo: Ho, it is a typo. I was suggesting adding "the".

PR #1212, opened 2021-03-07
Merge_Fuel_IT: enable pompe bianche

fansanelli:
"Pompe Bianche" are fuel station without a brand and have been skipped since the analyzer got created... This leads to confusion due to warning about unknown or missing ref:mise, let's allow them now.

Thanks
Francesco

### DIFF
--- a/analysers/analyser_merge_fuel_IT.py
+++ b/analysers/analyser_merge_fuel_IT.py
@@ -58,7 +58,7 @@ class Analyser_Merge_Fuel_IT(Analyser_Merge):
             CSV(Source_Fuel(Source(attribution = 'MISE - Ministero Sviluppo Economico', fileUrl = 'https://www.mise.gov.it/images/exportCSV/anagrafica_impianti_attivi.csv'),
                     fileUrl = 'https://www.mise.gov.it/images/exportCSV/prezzo_alle_8.csv')),
             Load('Longitudine', 'Latitudine',
-                where = lambda row: row['Bandiera'] != 'Pompe Bianche' and row['Longitudine'] != 'NULL' and row['Latitudine'] != 'NULL'),
+                where = lambda row: row['Longitudine'] != 'NULL' and row['Latitudine'] != 'NULL'),
             Conflate(
                 select = Select(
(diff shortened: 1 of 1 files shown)


analysers/analyser_merge_fuel_IT.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, config, logger = None):
                     mapping1 = {
                         'ref:mise': 'idImpianto',
                         'operator': lambda res: italian_strings.normalize_common(res['Gestore']),
-                        'brand': 'Bandiera',
+                        'brand': lambda res: res['Bandiera'] if res['Bandiera'] != 'Pompe Bianche' else Mapping.delete_tag,
```
frodrigo, line 74: Since you exclude the data at load, you does not need this change.

fansanelli: I need it in case of rebrand

frodrigo: No. `res` is only OpenData. In this code you does not have access to OSM data.

fansanelli: I've just tested it and it adds the row to remove the brand in the XML as expected for pompe bianche.. what do you mean?

frodrigo: At the load step you exclude row with 'Pompe Bianche'. So in the folowing step on data process you have no more this data, and so the `if res['Bandiera'] != 'Pompe Bianche' ` is useless.

It you really want to suggest delete of OSM tags brand=Pompe Bianche, keep the Mapping.delete_tag, but you have to include the the row with ''Pompe Bianche' on load step.

fansanelli: Check row 61

frodrigo: Ho, sorry, I read the diff in opposite way!
So, it s good for me.

fansanelli: Nevermind and thanks

PR #1232, opened 2021-03-24
Allow quick fix on post_box_FR analyser

PanierAvide:
As this analyser is the only one not offering tag suggestions for creating new post boxes, I submit this PR in order to make tools able to get directly post box ref.


frodrigo:
Deployed.

PR #1233, opened 2021-03-24
Add analyser merge_natural_FR

archie83:
Add a new analyser to merge data from BD ORTHO IGN relative to relief (orography).

### DIFF
--- a/analysers/analyser_merge_natural_FR.py
+++ b/analysers/analyser_merge_natural_FR.py
@@ -32,45 +32,45 @@ class Analyser_Merge_Natural_FR(Analyser_Merge_Dynamic):
 
         maping = [
         #   40806 Vallée
-        dict(
-            #   40802
-            #       4 Ancien glacier
-            item = 8530,
-            id = 100,
-            level = 3,
-            title = T_('Valley not integrated'),
-            select = {'nature': 'Vallée'},
-            tags = {'natural': 'valley'},
-        ),
+#       dict(
+#           #   40802
+#           #       4 Ancien glacier
+#           item = 8530,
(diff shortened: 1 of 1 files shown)


frodrigo:
Have you make it run?

Usualy we try to avoid low frequency items, think the query will be run each day forever on a server.

archie83:
Yes, I tested it for the Alpes-Maritimes. Shouldn't it possible to run merge analysers less often, for example once a week?

frodrigo:
> Yes, I tested it for the Alpes-Maritimes. Shouldn't it possible to run merge analysers less often, for example once a week?

We can, but we have no plan to do this. And here frequent and less frequent are in the same analyser so it cannot be  done in this away anyway.

archie83:
I must admit that some elements, such as stones, are not very useful. Would it be acceptable if we keep only the most important ones, such as (by order of preference):
- Peaks
- Saddles
- Cave entrances and pits
- Capes
- Beaches
?

frodrigo:
Yes. Good candidates are frequent and not prone to false positives.

frodrigo:
Look good to me. Thank you.
Merged, but not deployed, yet.

archie83:
Changes in the frontend are also required, aren't they? Item #8530 is not known in the frontend.

frodrigo:
Yes, do not brother about that.

frodrigo:
Deployed but but not run.
Miss declaration in configuration FR + oversee, can you make a PR please?

archie83:
Is it enough to add two lines to osmose_config.py ?

frodrigo:
Yes

PR #1242, opened 2021-04-03
Analyser_Merge_Poste_FR: remove postcode

gileri:
This tag was used in this integration since at least 2013, when I believe postcodes were not as well mapped, so having a "fallback" from the post office for Nominatim and other geocoders was a good idea.

As far as I know postcodes are [pretty well mapped as areas today in France](http://taginfo.openstreetmap.fr/keys/addr%3Apostcode). So the benefit of this tag are slim, while they :

- May confuse geocoders (prefering such nodes instead of more detailed zones for some reason)
- Incite contributors to add this tag to other post offices or other objects while it's not necessary.

What do you think ?


frodrigo:
Ok. Yes we does not really need post code on this kind of object.

frodrigo:
Deployed.

gileri:
Thank you for the follow-ups @frodrigo !

PR #1250, opened 2021-04-17
fix Gas Stations Operator Problem in South Tyrol

### LINKED ISSUE #1247: Gas Stations Operator Problem in South Tyrol
Greetings,

I'm a mapper from the bilingual province of South Tyrol (Alto Adige) from Northern Italy. I noticed that there are quite some wrong markers in my province where Osmose wants to update the operator of a gas station, for example this one:

osmose.openstreetmap.fr/en/error/6e7c424a-2872-42a5-a6fb-f11cc75dbd97

I noticed that the input dataset correctly contains both the German and Italian Operator name and uses ' % ' to differentiate between them.

Could this be converted to ' - ', the way we split our bilingual names here? The marker should not show up as an issue when the operator already matches, or when it's flipped (`German Operator - Italian Operator` instead of `Italian Operator - German Operator`)

fansanelli:
Fix #1247

As suggested replaced " % " with " - " and moved operator to mapping2 to make it optional.

Cheers
Francesco

### DIFF
--- a/analysers/analyser_merge_fuel_IT.py
+++ b/analysers/analyser_merge_fuel_IT.py
@@ -70,7 +70,6 @@ class Analyser_Merge_Fuel_IT(Analyser_Merge):
                     static2 = {'source': self.source},
                     mapping1 = {
                         'ref:mise': 'idImpianto',
-                        'operator': lambda res: italian_strings.normalize_common(res['Gestore']),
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank you. Not deployed yet.

fansanelli:
Thanks @frodrigo

PR #1254, opened 2021-04-25
New public transport analysis

nlehuby:
new public transport analysis that checks 
- the order of the stops 
- the side of the street

It mostly works, but with some rough angles ... help welcomed to fine tune ;)

First, it computes the shape of each route as a linestring when possible : 
it takes the ways in the relation in their order and try to make a linestring with it.
The only magic is with the un-cut roundabout : as they are closed way and don't fit in the linestring, it skips them (which is not great, see below). 

_For the order of the stops analysis:_
Each platform is projected on the route geom (to create a kind of OSM `stop_position`). Then we have an error if the order of the platform in the route relation is not the same as the order of the computed stop_position on the route linestring.

Mostly tested on buses.

_For the side of street analysis_ (only for street served transport):
It makes a buffer on the right side of the linestring geom (because as we drive on the right, the bus stop are always on the right side of the road)

image
> (the buffer is actually bigger than on the picture, to stay coherent with the analysis "The stop or platform is too far from the track of this route")

Then we have an error if the platform is not inside the buffer.
> this one is ok:
image

> this one is an issue:
image

I've spotted two types of false positives:
The edges of the route relation: this seems quite frequent to have the start of the linestring not fully aligned with the stop.
image

I've chosen to ignore the first and last stop from the analysis because I don't know how to handle this in a better way. **Ideas welcomed** ;)

The stops on un-cut roundabout: as I simplify the roundabout when they are closed way, if there is a stop on the roundabout, it is very likely to not fit in the computed buffer:
image

> example of false positive:
image

I guess the solution here would be to cut the roundabout when making the route linestring, but I didn't manage to dot it. **Help welcomed** ;)
**EDIT**: I finally managed to cut the roundabout :tada: no more false positives here!

It needs to be reversed in countries that drives on the other side (take a left buffer if the bus stops are on the left side of the street). I guess we already have this kind of info as country config in Osmose, but I didn't find out how to do it : **help needed here too** !
**EDIT**: done :heavy_check_mark:

### DIFF
--- a/analysers/analyser_osmosis_relation_public_transport.py
+++ b/analysers/analyser_osmosis_relation_public_transport.py
@@ -475,7 +475,12 @@ SELECT route_linestring.id AS route_id,
        ROW_NUMBER () OVER (PARTITION BY route_linestring.id
                      ORDER BY stop_platform.morder) AS stop_order,
        ROW_NUMBER () OVER (PARTITION BY route_linestring.id
-                     ORDER BY ST_LineLocatePoint(ST_OffsetCurve(route_linestring.geom, -10), stop_platform.geom) DESC) AS projected_stop_order
+                     ORDER BY ST_LineLocatePoint(ST_OffsetCurve(route_linestring.geom, -10), stop_platform.geom) DESC) AS projected_stop_order,
+       (CASE
+          WHEN LEAD(stop_platform.morder, 1) OVER (PARTITION BY route_linestring.id
+                     ORDER BY stop_platform.morder) IS NULL THEN 1
+          ELSE 0
+       END) AS is_last_platform
 FROM stop_platform
 JOIN route_linestring ON route_linestring.id = stop_platform.id
 WHERE stop_platform.mrole IN ('platform',
@@ -506,17 +511,10 @@ SELECT
   ST_AsText(ST_Transform(platform_that_can_project.stop, 4326))
 FROM platform_that_can_project
 JOIN route_linestring ON route_linestring.id = platform_that_can_project.route_id
-JOIN (
-  SELECT
-    platform_that_can_project.route_id AS route_id,
-    MAX(platform_that_can_project.stop_order) AS last_platform
-  FROM platform_that_can_project
-  GROUP BY platform_that_can_project.route_id
-) AS y ON platform_that_can_project.route_id = y.route_id
 WHERE ST_DWithin(route_linestring.geom, platform_that_can_project.stop, 50) AND
   NOT ST_Intersects(ST_Buffer(route_linestring.geom, 50, 'side={}'), platform_that_can_project.stop) AND
   route_linestring.public_transport_mode IN ('bus', 'trolleybus', 'coach', 'share_taxi', 'school_bus') AND
-  platform_that_can_project.stop_order <> last_platform AND
+  platform_that_can_project.is_last_platform <> 1 AND
   platform_that_can_project.stop_order <> 1
 """
 
--- a/analysers/analyser_osmosis_relation_public_transport.py
+++ b/analysers/analyser_osmosis_relation_public_transport.py
@@ -427,6 +427,7 @@ $$;
 sql101 = """
 CREATE TEMP TABLE route_linestring AS
 SELECT id,
+       relations.tags->'route' AS public_transport_mode,
        ST_Transform(generate_linestring_geom(id), {0}) AS geom
 FROM relations
 WHERE relations.tags->'type' = 'route'
@@ -500,6 +501,7 @@ JOIN (
 ) AS y ON platform_that_can_project.route_id = y.route_id 
 WHERE ST_DWithin(route_linestring.geom, platform_that_can_project.stop, 50) AND
   NOT ST_Intersects(ST_Buffer(route_linestring.geom,50,'side=right'), platform_that_can_project.stop) AND
+  route_linestring.public_transport_mode IN ('bus', 'trolleybus', 'coach', 'share_taxi', 'school_bus') AND
   platform_that_can_project.stop_order <> last_platform AND
   platform_that_can_project.stop_order <> 1
 """
--- a/analysers/analyser_osmosis_relation_public_transport.py
+++ b/analysers/analyser_osmosis_relation_public_transport.py
@@ -459,6 +459,7 @@ WHERE relations.tags->'type' = 'route'
                              'school_bus')
   AND (NOT relations.tags?(relations.tags->'route')
        OR relations.tags->(relations.tags->'route') != 'on_demand')
+  AND generate_linestring_geom(id) IS NOT NULL
 """
 
 sql101b = """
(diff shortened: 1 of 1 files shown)


frodrigo:
> I guess the solution here would be to cut the roundabout when making the route linestring, but I didn't manage to dot it. **Help welcomed** ;)

Maybe just include completely the roundabout, so it avoid false positive.

> It needs to be reversed in countries that drives on the other side (take a left buffer if the bus stops are on the left side of the street). I guess we already have this kind of info as country config in Osmose, but I didn't find out how to do it : **help needed here too** !

Access drive side like this https://github.com/osm-fr/osmose-backend/blob/c6c0e8a4598495c3b21c308081dbda92be0d1717/analysers/analyser_osmosis_roundabout_reverse.py#L64

nlehuby:
Thanks for the drive side, I will investigate that :+1: 


> > I guess the solution here would be to cut the roundabout when making the route linestring, but I didn't manage to dot it. **Help welcomed** ;)
> 
> Maybe just include completely the roundabout, so it avoid false positive.

It won't be a linestring anymore if we add a closed way in the middle of it

frodrigo:
Could you use uppercase for all SQL key word, please ?

analysers/analyser_osmosis_relation_public_transport.py
```diff
@@ -366,6 +368,144 @@
...
+CREATE OR REPLACE FUNCTION generate_linestring_geom(route_id bigint) RETURNS geometry(MultiLineString, 4326) LANGUAGE PLPGSQL AS $$
+declare
+   f record;
+   full_way geometry(LineString, 4326);
+   last_roundabout geometry(LineString, 4326);
+begin
+    for f in
+        select ways.id, ways.linestring AS geom from relation_members
+        join relations on
+            relations.id = relation_members.relation_id
```
frodrigo, line 381: Also filter on relation member type `'w'` before join on way.

analysers/analyser_osmosis_relation_public_transport.py
```diff
@@ -423,4 +569,10 @@ def analyser_osmosis_common(self):
         self.run(sql70, self.callback70)
         self.run(sql80, self.callback80)
         self.run(sql90, self.callback90)
-        self.run(sqlA0.format(self.config.options.get("proj")), self.callbackA0)
```
frodrigo, line 426: You really want to delete this class ?

analysers/analyser_osmosis_relation_public_transport.py
```diff
@@ -366,6 +368,144 @@
...
+       ROW_NUMBER () OVER (PARTITION BY route_linestring.id
+                     ORDER BY ST_LineLocatePoint(ST_OffsetCurve(route_linestring.geom, -10), stop_platform.geom) DESC) AS projected_stop_order
+FROM stop_platform
+JOIN route_linestring ON route_linestring.id = stop_platform.id
+WHERE stop_platform.mrole in ('platform',
+                              'platform_exit_only',
+                              'platform_entry_only')
+  AND stop_platform.member_type='N'
+  AND GeometryType(ST_OffsetCurve(route_linestring.geom, -10)) = 'LINESTRING'
+  AND ST_LineLocatePoint(route_linestring.geom, ST_ClosestPoint(route_linestring.geom, stop_platform.geom)) <> 1
```
frodrigo, line 472: You disallow the last point, but allow the first (0) ?

analysers/analyser_osmosis_relation_public_transport.py
```diff
@@ -366,6 +368,144 @@
...
+  platform_that_can_project.stop_id,
+  ST_AsText(ST_Transform(platform_that_can_project.stop, 4326))
+FROM platform_that_can_project
+JOIN route_linestring ON route_linestring.id = platform_that_can_project.route_id
+JOIN (
+  SELECT
+    platform_that_can_project.route_id AS route_id,
+    MAX(platform_that_can_project.stop_order) AS last_platform
+  FROM platform_that_can_project
+  GROUP BY platform_that_can_project.route_id
```
frodrigo, line 500: Maybe you can precompute the last platform directly in `platform_that_can_project` when window function `lead()` is null.

analysers/analyser_osmosis_relation_public_transport.py
```diff
@@ -366,6 +368,144 @@
...
+                             'aerialway',
+                             'ferry',
+                             'coach',
+                             'funicular',
+                             'share_taxi',
+                             'light_rail',
+                             'school_bus')
+  AND (NOT relations.tags?(relations.tags->'route')
+       OR relations.tags->(relations.tags->'route') != 'on_demand')
+"""
```
frodrigo, line 449:
````
+    generate_linestring_geom(id) IS NOT NULL
````

File: analysers/analyser_osmosis_relation_public_transport.py (hunk 2 above)
Comment (nlehuby) [reply analysers/analyser_osmosis_relation_public_transport.py:426]: my bad, it was a mistake, thanks !

**Comment by nlehuby** [comment]
I think it's ready :rocket:

File: analysers/analyser_osmosis_relation_public_transport.py [hunk 6]
```diff
@@ -406,7 +562,13 @@ def __init__(self, config, logger = None):
         self.callback80 = lambda res: {"class":8, "data":[self.relation_full, self.any_full, self.positionAsText]}
         self.callback90 = lambda res: {"class":9, "data":[self.relation_full, self.any_full, self.positionAsText]}
         self.callbackA0 = lambda res: {"class":10, "data":[self.node_full, self.positionAsText]}
+        self.callback100 = lambda res: {"class":11, "data":[self.relation_full, self.positionAsText]}
+        self.callback110 = lambda res: {"class":12, "data":[self.relation_full, self.any_full, self.positionAsText]}
 
+        if self.config.options.get("driving_side") == "left":
+            self.buffer_driving_side = "left"
+        else:
+            self.buffer_driving_side = "right"
```
frodrigo, line 571: Unused variable `buffer_driving_side`.

nlehuby: pff, I really need vacations ... Should be ok now

frodrigo:
Look good. Thank you. Not deployed yet.

frodrigo:
Deployed yesterday. But cause lot of trouble. Take too many time on may countries.. up to timeout.

Lot of exceptions: https://buildbot.osmose.openstreetmap.fr/#/failing_builds

https://buildbot.osmose.openstreetmap.fr/#/builders/195/builds/262/steps/1/logs/stdio

frodrigo:
Fixed with this change https://github.com/osm-fr/osmose-backend/commit/5e489887b6f85b9497ccfa7d8c8580903a8d60c4

osm-ToniE:
I think, we have a false-positive here:

https://ptna.openstreetmap.de/relation.php?id=9832010

Here, “stop_position” of stop 11 must be ignored for the bus being on the way to stops 9 and 10, but must be considered afterwards. 

Ignore stop_position where the platform is on the "wrong" side, even if it is a stop_position of this route?

osm-ToniE:
same here:

https://ptna.openstreetmap.de/relation.php?id=9832269

and this has been reported

Issue #1256, opened 2021-05-02
detect (ARABIC DAMMA, 0x064f) as invalid character in "alt_name" key


hubaishan:
in 5070 invalid character
osmose detected "ُ" (ARABIC DAMMA, 0x064f) as invalid character in alt_name key while it a valid Arabic diacritic mark .
https://osmose.openstreetmap.fr/en/map/#item=5070&zoom=17&lat=30.556565&lon=41.146459&level=2&issue_uuid=a626bfdd-ae08-87c3-a8a7-7ee72cc29206

frodrigo:
Is not only a script setting, but also by language (/country).
The rules are here https://github.com/osm-fr/osmose-backend/blob/master/modules/languages.py

Since I am not able to read Arabic, what change is required in this file?

hubaishan:
Here is PR https://github.com/osm-fr/osmose-backend/pull/1259

PR #1259, opened 2021-05-03
fix issues/1256 to include all Arabic tashkeel

hubaishan:
Adding all Arabic Tashkeel marks
as in https://unicode-table.com/en/blocks/arabic/ 
and https://www.unicode.org/charts/PDF/U0600.pdf

### DIFF
--- a/modules/languages.py
+++ b/modules/languages.py
@@ -22,7 +22,7 @@
 # http://www.regular-expressions.info/unicode.html#script
 # https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
 language2scripts = {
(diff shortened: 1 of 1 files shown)


frodrigo:
@iriman this changes make sense to you, especially on Farsi?

hubaishan:
To avoid https://github.com/osm-fr/osmose-backend/issues/1256

iriman:
@frodrigo Yes, these diacritics are correct for Farsi, too. But, as @hubaishan mentioned, `064F` was missing (compare with character table on page 2 of U0600.pdf)

frodrigo:
Thank you all.

PR #1272, opened 2021-05-25
Fix analyser merge power substation minor following Enedis opendata update

flacombe:
Enedis had updated its opendata a few weeks ago regarding power distribution substations it operates.
In the past, item 8280 only had opendata for `minor_distribution` substations from Enedis.

Now, Enedis mixes both `distribution` and `minor_distribution` in the same dataset without attribute to distinguish them.
Item 8280, class 11 could be improved a bit as Enedis won't provide additional data that could enable to make it accurate.
Records corresponding to osm's `substation=distribution` currently cause a lot of false positive that couldn't be fixed.

This PR is intended to make this issue http://osmose.openstreetmap.fr/fr/issue/a0d4dee9-fd41-60c8-4e3c-35914c281cdb disappear in regard of this osm object: https://www.openstreetmap.org/node/6512238382

voltage=20000 is now enough instead of voltage=20000;400

Proposed changes would make substation=minor_distribution and substation=distribution selected for conflation with opendata.
New objects to OSM should always be created with `substation=minor_distribution` but existing objects with `substation=distribution` should remain as this and don't move to `substation=minor_distribution`.
Should I move `"substation": "minor_distribution"` from static1 to static2 in Mapping or is it good in static1?

Additionally, what is the purpose of `fix:chair` in item 8281, class 13 at line 32 of this analyser please?

### DIFF
--- a/analysers/analyser_merge_power_substation_minor_FR.py
+++ b/analysers/analyser_merge_power_substation_minor_FR.py
@@ -45,20 +45,16 @@ class Analyser_Merge_Power_Substation_minor_FR(Analyser_Merge):
                 select = Select(
                     types = ["nodes", "ways"],
                     tags = [
-                        {"power": "substation", "substation": "minor_distribution", "operator": False},
-                        {"power": "substation", "substation": "minor_distribution", "operator": "EDF"},
-                        {"power": "substation", "substation": "minor_distribution", "operator": "ERDF"},
-                        {"power": "substation", "substation": "minor_distribution", "operator": "Enedis"},
-                        {"power": None, "transformer": "distribution", "operator": False},
-                        {"power": None, "transformer": "distribution", "operator": "EDF"},
-                        {"power": None, "transformer": "distribution", "operator": "ERDF"},
-                        {"power": None, "transformer": "distribution", "operator": "Enedis"}]),
+                        {"power": "substation", "substation": "minor_distribution", "operator": [False, "EDF", "ERDF", "Enedis"]},
+                        {"power": None, "transformer": "distribution", "operator": [False, "EDF", "ERDF", "Enedis"]},
+                        {"power": "substation", "substation": "distribution", "operator": [False, "EDF", "ERDF", "Enedis"]}]),
                 conflationDistance = 50,
                 mapping = Mapping(
                     static1 = {
                         "power": "substation",
-                        "substation": "minor_distribution",
-                        "voltage": "20000;400",
+                        "voltage": "20000",
                         "operator": "Enedis"},
-                    static2 = {"source": self.source},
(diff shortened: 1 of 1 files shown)


frodrigo:
> Proposed changes would make substation=minor_distribution and substation=distribution selected for conflation with opendata.
> New objects to OSM should always be created with `substation=minor_distribution` but existing objects with `substation=distribution` should remain as this and don't move to `substation=minor_distribution`.
> Should I move `"substation": "minor_distribution"` from static1 to static2 in Mapping or is it good in static1?

Yes should be moved to `static2`.

> Additionally, what is the purpose of `fix:chair` in item 8281, class 13 at line 32 of this analyser please?

It for item filter, set the issue could be fixed on chair mapping, could be changed to any value: http://osmose.openstreetmap.fr/api/0.3/tags

```python
                            {"power": None, "transformer": "distribution", "operator": False},
                            {"power": None, "transformer": "distribution", "operator": "EDF"},
                            {"power": None, "transformer": "distribution", "operator": "ERDF"},
                            {"power": None, "transformer": "distribution", "operator": "Enedis"}]),
                            {"power": None, "transformer": "distribution", "operator": "Enedis"},
                            {"power": "substation", "substation": "distribution", "operator": "Enedis"},

```

I think the list could be refactored using array of values to lower the number of SQL query and execution duration.
https://github.com/osm-fr/osmose-backend/blob/master/doc/4-Merge.md#select

flacombe:
Thank you @frodrigo 

The two comments are solved

frodrigo:
@frodrigo Should be fixed before merge to accept `False` in list.
https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Merge.py#L934

flacombe:
Is it an action required on my side?

frodrigo:
> Is it an action required on my side?

No, unless you wat to fix it in the osmose backend core.

flacombe:
I can try to do it.

According to documentation False means `no value or the tag doesn't exist`.
At lines 947/948 I see :
```
if None in v:
    cond = "(" + cond + " OR " + k_not_exists + ")"
```
As I don't know what's inside `k_not_exists` actually, I only guess there is a condition filtering on k doesn't exist, but why is it relevant for `None` and not for `False`?
I would had done the same for `False` and have a the opposite `k_exists` for `None`.

frodrigo:
f679eea27251bb2ee7ddd1e2a5e385b30ba720e3
I fixed the core, please can you rebase and test gain ?

flacombe:
Thank you @frodrigo 

Rebase ok and tests sounds ok too

frodrigo:
Thank you, merged, not deployed yet.

PR #1274, opened 2021-05-28
fix order in public transport analysis

### LINKED ISSUE #1262: The stops may not be in the right order 1260/11
This new analyser raise may false positive

EG https://www.openstreetmap.org/relation/2422779
Only two stops, the first is good, the last raise an issue
http://osmose.openstreetmap.fr/fr/error/65d681fc-ac08-4609-b04f-eb9b11f9a3d6

nlehuby:
put back the offset (to make sure the projection on the the ways is ok even if the same way is used multiple times in the relation)

Fix #1262

### DIFF
--- a/analysers/analyser_osmosis_relation_public_transport.py
+++ b/analysers/analyser_osmosis_relation_public_transport.py
@@ -511,7 +511,7 @@ SELECT
     stop_platform.member_type || stop_platform.mid AS stop_id,
     stop_platform.geom AS stop,
     ROW_NUMBER () OVER (PARTITION BY route_linestring.id ORDER BY stop_platform.morder) AS stop_order,
-    ROW_NUMBER () OVER (PARTITION BY route_linestring.id ORDER BY ST_LineLocatePoint(route_linestring.geom, stop_platform.geom)) AS projected_stop_order,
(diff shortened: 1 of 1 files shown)


frodrigo:
Cannot be back as is. Was removed because of massive performances issues.

Eg try on http://download.openstreetmap.fr/extracts/south-america/brazil/southeast/rio-de-janeiro.osm.pbf

nlehuby:
`time docker-compose --project-name brazil_rio_de_janeiro run --rm backend ./osmose_run.py --country=brazil_rio_de_janeiro --skip-download --no-clean --analyser osmosis_relation_public_transport`
gives me
```
real	7m54,332s
user	0m0,710s
sys	0m0,142s
```

which seems reasonable :thinking:

frodrigo:
It is probably linked to postgis/postgres or under lib version.

Look at this run 260/261 was killed after hours of timeout: https://buildbot.osmose.openstreetmap.fr/#/builders/371

I was able to reproduce the issue on osm110 debian with postgres 11.10, postgis 2.5.

frodrigo:
On today osm110 hardward it takes 1m38 to run the current analyzer for brazil_rio_de_janeiro. And 3m9 with this PR.

I try various changes and I am not able to reduce the time.

I have only one idea left, I have not tested it. Only apply ST_OffsetCurve when a segment is duplicated in the relation directly into generate_linestring_geom. @nlehuby or @esteinmann I let you try if you want.

PR #1276, opened 2021-05-31
Merge public transport FR - no network tag on bus_stop

### LINKED ISSUE #1268: network added on stops by STAR analyser and removed by JOSM QA
On analysers/analyser_merge_public_transport_FR_star.py, network is added on bus stops.
But another analyser asks to but it only on relations.

key | value
-- | --
source | 407661
item | 8042
class | 84

Incompatible with:

key | value
-- | --
source | 407718
item | 9014
class | 9014007

[Example of network addition proposal](http://osmose.openstreetmap.fr/en/issue/ec8e77ee-1784-a7a1-cb17-9debca20b920)

[Example of network addition proposal](http://osmose.openstreetmap.fr/en/issue/adec5c1f-2f84-67b7-cb81-f4b67e6978ee)))

One, I guess STAR analyser should be updated.

nlehuby:
Fix #1268

### DIFF
--- a/analysers/analyser_merge_public_transport_FR_star.py
+++ b/analysers/analyser_merge_public_transport_FR_star.py
@@ -54,8 +54,7 @@ class Analyser_Merge_Public_Transport_FR_Star(Analyser_Merge):
                     static1 = {
                         "highway": "bus_stop",
                         "public_transport": "platform",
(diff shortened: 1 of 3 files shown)


frodrigo:
Thank you.

PR #1279, opened 2021-06-06
Add warning for roundabouts without right of way

Famlam:
Roundabouts without right of way (or without one-directional traffic) should be tagged junction=circular instead.
Added this as a likely trap to the docs.


frodrigo:
Thank you.

PR #1291, opened 2021-07-04
Enable warning for `access=yes` on `highway=busway`

Famlam:
highway=busway was recently approved and is slowly being rolled out:
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbusway

Shouldn't contain access=yes or similar (as this'd make it a normal highway=unclassified road or similar, rather than the distinct meaning of highway=busway)


frodrigo:
Thank

PR #1296, opened 2021-07-10
Update mapillary-traffic-signs.mapping.json

BalooUriza:
Suggesting stop sign additions seems to be working OK in OK, extending to the rest of the US.


frodrigo:
Usage of "Stop" was restricted to avoid overflow of issues.
While it could be ok in OK because it is well mapped, it may not be the case across the country.

frodrigo:
Merged. But not deployed yet.

Issue #1293, opened 2021-07-06
`man_made=piste:halfpipe` is criticized as error even though it's mentioned on the OSM Wiki.


habi:
I've been on a small ski-piste tagging spree (I know, it summer, but nonetheless).
I've tagged a half-pipe with `man_made=piste:halfpipe`, as suggested by the [OSM Wiki here](https://wiki.openstreetmap.org/wiki/Key:piste:type#See_also).

Osmose complained about this tag and linked me to https://github.com/osm-fr/osmose-backend/blob/da3d341478cef978cb8a6c5961050aa76df389d5/plugins/TagFix_BadValue.py#L40 where the error springs up.

The 'details' from Osmose are [here](http://osmose.openstreetmap.fr/en/issue/5be06a40-6074-2dc3-995c-f6686994ca07), the way in question is [here](https://www.openstreetmap.org/way/961560443).

Am I missing some additional  `piste`-tags on the half-pipe?

frodrigo:
Should be added in the exception list here https://github.com/osm-fr/osmose-backend/blob/da3d341478cef978cb8a6c5961050aa76df389d5/plugins/TagFix_BadValue.py#L84

habi:
I tried to add this to a PR, let me know if this needs something more!

PR #1297, opened 2021-07-12
Add man_made=piste:halfpipe to exceptions

habi:
See https://github.com/osm-fr/osmose-backend/issues/1293, the `piste` is mentioned on the OSM Wiki as the way to tag: https://wiki.openstreetmap.org/wiki/Key:piste:type#See_also


frodrigo:
Thank, merged, but not deployed, yet.

frodrigo:
Deployed

Issue #873, opened 2020-05-17
Bad turn lanes order


maro-21:
Osmose points out that `turn:lanes=reverse;left|left|through|through|through;right` is "Bad turn lanes order". Why?

frodrigo:
Do you have a link?
Issue item/class id, please ?

maro-21:
item: 3160
class: 31607

frodrigo:
It is form here
https://github.com/osm-fr/osmose-backend/blob/master/plugins/Highway_Lanes.py#L131

I re-read quickly the code and see no issue, yet.

maro-21:
Maybe `reverse`?
Where is the order in the code? Here?
https://github.com/osm-fr/osmose-backend/blob/c3cbc59cc61c76b9c96398a63ff7930d0139611e/plugins/Highway_Lanes.py#L112-L118
If so, `reverse` is in the middle.

frodrigo:
No this part is only translation.

Then reverse is ignored.
https://github.com/osm-fr/osmose-backend/blob/master/plugins/Highway_Lanes.py#L120

Order check is here
https://github.com/osm-fr/osmose-backend/blob/master/plugins/Highway_Lanes.py#L121-L131

maro-21:
Another examples:
- [`turn:lanes=none|left|left;right`](http://osmose.openstreetmap.fr/pl/error/faa38f29-fd19-756c-a806-18abd7ba08e3)
Here we have `turn:lanes=none|left|left;right` which looks ok.
- [`turn:lanes=left;through|right|right|none`](http://osmose.openstreetmap.fr/pl/error/17de631d-a6c2-f8ca-1aa5-20fca58b57ff)

frodrigo:
Both seem wrong to me. The none turn cannot be cross by other.

maro-21:
> Both seem wrong to me. The none turn cannot be cross by other.

What do you mean? None is a correct value.

frodrigo:
`left|none` is ok, but not `none|left`

maro-21:
> `left|none` is ok, but not `none|left`

On Wiki you have an example with `turn:lanes=none|none|none|merge_to_left`
https://wiki.openstreetmap.org/wiki/Key:turn#Motorway_with_links_and_destinations

frodrigo:
merge_to_left is not turn left.

dex2000:
> Both seem wrong to me. The none turn cannot be cross by other.

[...]

PR #1305, opened 2021-07-24
Fix some of the cases mentioned in #873

Famlam:
1. Fix the replacing order of 'left' and 'right' (converting 'merge_to_left' to 'merge_to_l', for instance, causing it to become a ' ')
2. Allow a single "none" of ""-lane as the outer lane (i.e. `turn:lanes=left|through|right|`), as this could very well be a bus lane (mentioned a few times in #873)

and test case for both that previously failed

To be fixed (might try that another time) and also part of #873:
- `;`-separated individual values (such as in `reverse;left|left`) are replaced (in the lambda function) by ` ` == through, resulting in ` l` == through|left for this example.
A way to fix this is probably to sort the `;`-separated values in this check according to (reverse -> (slight/sharp)left -> through ->(slight/sharp)right), then treat `;` as `|` for this check, rather than replace the full `reverse;left` by ` ` == through.
That would also immediately add warnings for i.e. having `turn:lanes=right;through|through|right` or `turn:lanes=through|right;left|through`

### DIFF
--- a/plugins/Highway_Lanes.py
+++ b/plugins/Highway_Lanes.py
@@ -111,14 +111,17 @@ side* and `the merge_to_left` on the *right side*.'''))
                     if not unknown:
                         # merge_to_left is a on the right and vice versa
                         t = tags_lanes[tl] \
-                            .replace("left", "l").replace("slight_left", "l").replace("sharp_left", "l") \
+                            .replace("slight_left", "l").replace("sharp_left", "l") \
                             .replace("through", " ") \
-                            .replace("right", "r").replace("slight_right", "r").replace("sharp_right", "r") \
+                            .replace("slight_right", "r").replace("sharp_right", "r") \
                             .replace("reverse", "U") \
                             .replace("merge_to_left", "r").replace("merge_to_right", "l") \
-                            .replace("none", " ").replace(";", "").split("|")
-                        t = ''.join(map(lambda e: " " if len(e) == 0 or e[0] != e[-1] else e[0], map(sorted, t)))
+                            .replace("left", "l").replace("right", "r") \
+                            .replace("none", "N").replace(";", "").split("|")
+                        t = ''.join(map(lambda e: "N" if len(e) == 0 else " " if e[0] != e[-1] else e[0], map(sorted, t)))
                         t = t.replace('U', '') # Ignore reverse
+                        # Ignore single none on the outside lanes: it could be a bus lane 
+                        t = t.replace('rN', 'r').replace('Nl', 'l').replace('N', ' ')
                         last_left = self.rindex_(t, "l")
                         first_space = self.index_(t, " ")
                         last_space = self.rindex_(t, " ")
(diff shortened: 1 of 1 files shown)


Famlam:
As a comment:

Probably there are better ways to do the "allow outer ways to be none"-replace step, but 
`t = t.replace('rN', 'r').replace('Nl', 'l').replace('N', ' ')`
seems to do the job well in a single line without lambda's.

But feel free to modify it (right now it'd allow `right||right`, which is very unlikely, at least from the areas where I encounter dedicated road lanes)

plugins/Highway_Lanes.py
```diff
@@ -111,14 +111,17 @@ def way(self, data, tags, nds):
...
                             .replace("reverse", "U") \
                             .replace("merge_to_left", "r").replace("merge_to_right", "l") \
-                            .replace("none", " ").replace(";", "").split("|")
-                        t = ''.join(map(lambda e: " " if len(e) == 0 or e[0] != e[-1] else e[0], map(sorted, t)))
+                            .replace("left", "l").replace("right", "r") \
+                            .replace("none", "N").replace(";", "").split("|")
+                        t = ''.join(map(lambda e: "N" if len(e) == 0 else " " if e[0] != e[-1] else e[0], map(sorted, t)))
                         t = t.replace('U', '') # Ignore reverse
+                        # Ignore single none on the outside lanes: it could be a bus lane
+                        t = t.replace('rN', 'r').replace('Nl', 'l').replace('N', ' ')
```
frodrigo, line 124: Should we not ensure N is first/last/only lane to remove it ? What about multiple `N` - Should we replace `N` at all ?

frodrigo:
Thank you to deep into this complex code.
Feel free to open new issue or PR to the other points, now or latter.

plugins/Highway_Lanes.py
Famlam: I've updated the PR to ensure it only replaces the outmost lanes.
I believe that it is very unlikely that a dedicated bus lane would be in between lanes, or that there would be multiple dedicated bus lanes on a single road. However, the check would lose a lot of its strength if all 'nones' would become unchecked, as it very often refers to "regular" lanes being "through"-lanes without road marking.
Hence I think it's better to keep treating `N` as `through` in the majority of cases, i.e. `left|||right|` very likely means `left|no turn = through|no-turn=through|right|some-special-lane`

Famlam:
Thanks for the review!
I think I addressed the in-line review, please let me know if you'd prefer something changed or whether it's ready to merge now!

The only bullet point left is actually the initial post on #873 (so no need to open a new issue). It turns out that all other comments in that issue (starting at the comment of dex2000) on that were caused by the issues I'm addressing here.

However, as I'll go on holiday soon, it'll probably take a while before I can go into that, hence I marked this pull request available to review/merge already :).

frodrigo:
Thank you. Merged but but not deployed yet.

I let you manage to create the new issue, as you have more clear idea on that than me for now.

Famlam:
> I let you manage to create the new issue, as you have more clear idea on that than me for now.

I've created issue #1306 to address the remaining issue, as well as three possible ways to resolve it (varying from simple to difficult, but thereby also from poor to good)

Thanks for merging!

Famlam:
> 1. Fix the replacing order of 'left' and 'right' (converting 'merge_to_left' to 'merge_to_l', for instance, causing it to become a ' ')

By fixing this sequence of replacing, a tiny issue popped up for roads where `through|merge_to_right|through` or `through|merge_to_left|through` was present. Previously it was (incorrectly) treated as `through|through|through` (and thus didn't raise an issue). Now it raises an issue (as now it's replaced by `through|right or left|through`). Fix is included in #1310.

PR #1307, opened 2021-07-26
Add test for issue 1288

dfaure:


plugins/TagFix_Opening_Hours.py
```diff
@@ -81,6 +81,7 @@ def test(self):
         # These are OK, no suggestion
         assert not a.node(None, {'opening_hours': 'Mo-Fr 10:00-19:00'})
         assert not a.node(None, {'opening_hours': 'Mo-Tu,Th-Fr 09:30-12:00; We 15:00-17:00; 2020 Dec 24,2020 Dec 31 off; Sa,Su off; PH off'})
+        assert not a.node(None, {'opening_hours': 'Mo off, Tu-Th 09:00-18:00; Fr 09:00-19:00; Sa 08:00-18:00; Su off'})
```
jocelynj, line 84: I'm wondering about the test: shouldn't we have "Mo off; ...", with `;` instead of `,` ?

Or is the meaning different?

dfaure: It makes no difference in this particular case, since the day selectors are different.

"," adds to the previous rule

";" replaces previous rules

No difference here. I just picked the exact string from the bug report.

jocelynj: Thanks for the clarification.

Issue #1323, opened 2021-07-15
Charte de toponymie file url is broken (404)


JackNUMBER:
On Toponymie issues the link in the details text is broken. 
I'm not talking about the IGN's website's 404 when you try to access the ressource, I'm talking about the link itself. It leads to `http://osmose.openstreetmap.fr/fr/map/education.ign.fr/sites/all/files/charte_toponymie_ign.pdf`.

Regarding the [source code](https://github.com/osm-fr/osmose-backend/blob/008d75162eccd306d844812c9a56f836fdd680cb/plugins/Name_Toponymy_FR.py#L38), the protocol is missing in the url.

An example issue [here](http://osmose.openstreetmap.fr/fr/map/#zoom=18&lat=48.938291&lon=2.174327&item=xxxx&level=1%2C2&issue_uuid=b0e76f27-cbf0-c6a6-b2f4-ddd6ee715aea).

![image](https://user-images.githubusercontent.com/2306550/125768269-30875ccd-84bf-4d9c-84ba-89dab944762d.png)

Issue #1322, opened 2021-08-27
Broken links in Name_Toponymy_FR (5040)


Ltrlg:
There are two broken links in the information panel for [this analyser](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_Toponymy_FR.py#L37-L42):

* The first one in “details” is incorrect Markdown and links to https://osmose.openstreetmap.fr/en/map/education.ign.fr/sites/all/files/charte_toponymie_ign.pdf instead of http://education.ign.fr/sites/all/files/charte_toponymie_ign.pdf.
* The second one (the same, except pointing to the intended location) in “resource” no longer points to anything on the IGN website.

Example: [b2b32ad0-4bdc-47d6-339a-1ea1b8980dc3](https://osmose.openstreetmap.fr/en/map/#issue_uuid=b2b32ad0-4bdc-47d6-339a-1ea1b8980dc3)

I’ve not been able to find the PDF on the reworked IGN website, but it is still accessible [on the Internet Archive](https://web.archive.org/web/2019/http://education.ign.fr/sites/all/files/charte_toponymie_ign.pdf).

PR #1324, opened 2021-08-28
Fix links in Name_Toponymy_FR

Ltrlg:
This commit fixes the two issues in #1322 and #1323 :

* The link in the “details” section now uses a correct absolute link.
* Both links point to the archived version at the Internet Archive.

Note : I have not updated the `*.po` files, I’m not sure what is the proper process. The only impacted language is Japanese, all other have no translation here, not even French (!?).


frodrigo:
It is fine like this. Thank you.

PR #1338, opened 2021-10-01
Accepts '/' in names of platforms and bus_stops

### LINKED ISSUE #1299: The reason for error 5030 two names being reported
Hello there :)

I have a question first. What is the reason for banning `/` character from being used in the `name=*` tag? I haven't found this explained on https://wiki.openstreetmap.org. Can someone clarify this to me, please?

The reason why I'm asking is because `5030 two names` error is reported for every bus stop in Bydgoszcz city of Poland. For example [errors](https://osmose.openstreetmap.fr/en/byuser/sarxos?username=sarxos&item=5030) were reported for few changesets I did in the past:

image

But those are valid official names. See [here](http://www.zdmikp.bydgoszcz.pl/pl/transport/rozklad-jazdy/2014-07-02-12-57-42). There a tons of bus stops with the same naming convention, e.g. "Akademicka / Romanowskiej", etc.

From the error message in here https://osmose.openstreetmap.fr/en/issue/d9470852-73c6-72c4-807d-93b3586c139c I can see that there are some mentions about streets having two names when `name:left=*` and `name:right=*` must be used accordingly. I suppose this may be the reason for the error. Are we forced to mark all of these errors as false positives or can this be somehow automated?

Famlam:
fixes #1299


Famlam:
Not sure if silence means you disagree or simply overlooked this PR, but please also let me know if you don't want it  ;) @frodrigo

frodrigo:
> Not sure if silence means you disagree or simply overlooked this PR, but please also let me know if you don't want it ;) @frodrigo

I need to review it, but so I need time review it.

Famlam:
No problem ;) . I just wondered as it's been a month since I submitted it

frodrigo:
Reviewed. It is OK. Thank.
Sorry for delay.

Merged, but not deployed yet.

Famlam:
No problem and thanks! :)

jocelynj:
I've just deplayed this fix to OSM-FR backends. It will take at least one day to have updated results on frontend.

PR #1352, opened 2021-10-29
Avoid duplicate warning for 1-member multipolygon

Famlam:
A single member multipolygon relation will always trigger both error [11704 ](https://github.com/osm-fr/osmose-backend/blob/2b171ea2f84dd84ed2508fae5b061bfdf75d4b3c/plugins/Structural_Multipolygon.py#L44)(`This multipolygon is a simple polygon`) as well as error [12001 ](https://github.com/Famlam/osmose-backend/blob/4e25c0b8927a6c8c55020ee5300894d8bbd21622/plugins/Structural_Useless_Relation.py#L30) (`1-member relation`)

This disables the less specific duplicate "1 member relation" warning.

### DIFF
--- a/plugins/Structural_Multipolygon.py
+++ b/plugins/Structural_Multipolygon.py
@@ -66,14 +66,15 @@ class Structural_Multipolygon(Plugin):
                 err_members.append(u"{0} - {1}".format(member['type'], member['role']))
 
         err = []
-        if len(err_members) > 0:
-            err.append({"class": 11702, "subclass": 1, "text": {"en": ', '.join(err_members)}})
         if len(err_roles) > 0:
             err.append({"class": 11701, "subclass": 1, "text": {"en": ', '.join(err_roles)}})
-        elif outer == 1 and inner == 0:
-            err.append({"class": 11704, "subclass": 1})
+        if len(err_members) > 0:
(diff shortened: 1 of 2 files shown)


plugins/Structural_Multipolygon.py
```diff
@@ -66,15 +66,15 @@ def relation(self, data, tags, members):
...
-
-        if outer == 0:
-            err.append({"class": 11703, "subclass": 1})
+        if len(err_roles) > 0:
+            err.append({"class": 11701, "subclass": 1, "text": {"en": ', '.join(err_roles)}})
         elif outer == 1 and inner == 0:
             err.append({"class": 11704, "subclass": 1})
+        if outer == 0:
+            err.append({"class": 11703, "subclass": 1})
+
```
frodrigo, line 77: Please can you review all this changes, not required.

Famlam: Done!

I assume you agree with that warning 11704 should not be triggered if there's a typo in a role, e.g. a multipolygon with:
`outer`: a way
`inne`: a way inside the outer with a typo in the role

If not, I'll get rid of the `len(err_roles) == 0` check again

frodrigo: thank, right.

frodrigo:

PR #1357, opened 2021-11-03
fix typo found with codespell

matkoniecz:
https://github.com/codespell-project/codespell


frodrigo:
Only one!

Thank.

matkoniecz:
It is possible that there is more of them, I found this one mostly by accident.

PR #1370, opened 2021-12-04
Fixes #1369 (also for bicycle parkings, and mixed bicycle/motocycle p…

### LINKED ISSUE #1369: Paris motorcycle parking not integrated: should also check for parkings mapped as areas
Hello,

osmose signals an unmapped motorcycle parking here:

http://osmose.openstreetmap.fr/en/issue/3d4a2e98-79e8-ab86-9854-0cdd2cfe73d7

but in fact it is mapped as an area. The problem seems to be that analyser_merge_bicycle_parking_FR_paris.py only queries nodes.

-Ralf.

treinen:
…arkings).

This is not tested (I have no working testbed).


frodrigo:
Thank you. Merged, but not deployed, yet.

Issue #1295, opened 2021-07-10
item 4030 class 900 Conflict between tags: amenity, landuse - Single school on one site


imagoiq:
Hi,

According to the wiki, to map a [single school on one site](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dschool#Single_school_on_one_site), you should map the entire area with amenity=school and optionally landuse=education. In Osmose, using both is reported as an error. Osmose also has a trap (I'm not familiar with the trap system): "Sometimes the object needs both tags.". But in the end, I'm not confident that the targeted object needs both.

Here is for example an object in this situation: https://osmose.openstreetmap.fr/en/issue/280dce44-432f-5681-e4c2-646c9b056434

frodrigo:
Should be added here https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagRemove_Incompatibles.py#L45

PR #1371, opened 2021-12-05
Whitelist combinations with landuse=education

### LINKED ISSUE #1367: Tag conflict for "landuse=education"
item | 4030
class | 900
Detail: http://osmose.openstreetmap.fr/it/issue/acbaa277-218d-3b47-4dd0-4bccfe1e92a6

"Conflict between tags: amenity, landuse", but the Wiki for "[landuse=education](https://wiki.openstreetmap.org/wiki/Tag:landuse=education?uselang=it)" says _"A landuse=education usually exist in conjunction with an amenity=kindergarten;school;college;university. This connection can either exist at the area itself, or as elements within this area, or through membership in a Relation:site."_

Famlam:
Whitelist combinations of `landuse=education` together with `amenity=school|university|college|kindergarten`
Per the [wiki](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Deducation#Use_cases): 
> For an area/ground that is only used by one facility. In this case it is not necessary to use landuse=education because the amenity=kindergarten;school;college;university already imply the landuse. However, it is optional that these areas are also marked with landuse=education.

Also added a note regarding the deprecated tag `landuse=school` and removed an impossible combination (typo I presume?) of `landuse=school` together with `landuse=education` (can't have two identical tags on a single object).

Fixes #1367 
Updates #1295

### DIFF
--- a/plugins/TagRemove_Incompatibles.py
+++ b/plugins/TagRemove_Incompatibles.py
@@ -42,8 +42,11 @@ class TagRemove_Incompatibles(Plugin):
         self.CONFLICT[4] = set(['information', 'place'])
         self.WHITE_LIST = {
             'landuse': [
-                ['school', 'amenity', 'school'],
-                ['school', 'landuse', 'education'],
+                ['school', 'amenity', 'school'], # deprecated
+                ['education', 'amenity', 'school'],
+                ['education', 'amenity', 'university'],
+                ['education', 'amenity', 'college'],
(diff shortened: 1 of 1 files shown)


Famlam:
Lint seems to fail on nearly every file, so I guess that's unrelated to this commit

frodrigo:
Thank. Merged but not deployed, yet.

PR #1375, opened 2021-12-24
Validate conditional restrictions and report outdated conditions

Famlam:
1. Implement a **basic validation of conditional restrictions**:
- `@` must be present
- `(` and `)` count/sequence must match
- value and condition must be present in the sequence `value @ condition` or `value @ (condition)`
- parentheses required if the condition itself contains `;`
- `AND` in a combined condition must be followed and preceded by a condition

2. **Search for expired conditions**. Only using the current year at the moment

**Outdated, use the other PR instead**

PR #1376, opened 2021-12-26
Validate conditional restrictions and report outdated conditions

### LINKED ISSUE #1373: Search for former (now-redundant) temporary road closures
In The Netherlands, there is a slight tendency to map temporary road closures too using conditional restrictions, i.e. [`access:conditional=no @ (2021 Jun 18-2021 Sept 15)`](https://www.openstreetmap.org/way/6932270/history). However, this eventually leads to many unneeded conditional restrictions, like the way I used as an example.
Hence I would propose adding a (very simplified) way of checking for "aged" conditional restrictions, using approximately the following flow diagram:

1. (Skip the test on non-ways and ways without `highway` key)
2. Get all tags with `*:conditional` on the highways. If none, exit.
3. For every of these tags, get the value; split it by `;`. Let's call the fragments **"A"** here. Note to self: not if `;` in brackets, e.g. `no @ (2021; 2025)`
4. For all fragments in **"A"**, get the part after the `@`. Let's call it **"B"** here. (Note to self: check if `@` exists)
5. For every **"B"**, search for the *last* occurrence of 4 consecutive numbers (a year or the end of a year range). Let's call the result **C** here.
6. If **"C"** is found, check if **"C"** < current year. If so, raise a warning

It's simplified to years-only as the conditional syntax is too complicated to check for week ranges / month ranges etc. as well.

Please let me know if you think this test is (or isn't) a good addition; if fine, I'll try to program it in the Christmas-New Year break

Famlam:
1. Implement a **basic validation of conditional restrictions**:
- `@` must be present
- `(` and `)` count/sequence must match
- value and condition must be present in the sequence `value @ condition` or `value @ (condition)`
- parentheses required if the condition itself contains `;`
- `AND` in a combined condition must be followed and preceded by a condition

2. **Search for expired conditions**. Only using the current year at the moment

Fixes #1373 
(Originally proposed in #1375 )

### DIFF
--- a/plugins/conditionalrestrictions.py
+++ b/plugins/conditionalrestrictions.py
@@ -1,0 +1,210 @@
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights osmose project 2021                                        ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from modules.OsmoseTranslation import T_
+from plugins.Plugin import Plugin
+import re
+from datetime import date
+from modules.Stablehash import stablehash64
+
+class ConditionalRestrictions(Plugin):
+  def init(self, logger):
+    Plugin.init(self, logger)
+
+    self.ReYear = re.compile(r'20\d\d') # Update in 2099
+    self.currentYear = date.today().year
+
+    self.errors[33501] = self.def_class(item = 3350, level = 2, tags = ['highway', 'fix:chair'],
+        title = T_('Bad conditional restriction'),
+        detail = T_('''Conditional restrictions should follow `value @ condition; value2 @ condition2` syntax.
+Combined restrictions should follow `value @ (condition1 AND condition2)`.
+Parentheses `()` must be used around the condition if the condition itself contains semicolons `;`, i.e. `value @ (date;date)`.'''))
+    self.errors[33502] = self.def_class(item = 3350, level = 3, tags = ['highway', 'fix:chair'],
+        title = T_('Combine conditions using `AND`'),
+        detail = T_('''`AND` (uppercase) is to be preferred over lowercase variants when combining restrictions.'''))
+    self.errors[33503] = self.def_class(item = 3350, level = 3, tags = ['highway', 'fix:chair'],
+        title = T_('Expired conditional'),
+        detail = T_('''This conditional was only valid up to a date in the past. It can likely be removed.'''),
+        trap = T_('''Other tags might need to be updated too to reflect the new situation.'''))
+
+  def way(self, data, tags, nds):
+    # Get the relevant tags with *:conditional
+    tags_conditional = {}
+    for tag in tags:
+      if tag[-12:] == ":conditional":
+        if "source:" in tag or "note:" in tag or "fixme:" in tag:
+          continue
+        tags_conditional[tag] = tags[tag]
+    if len(tags_conditional) == 0:
+      return
+
+    err = []
+    for tag in tags_conditional:
+      tag_value = tags_conditional[tag]
+      conditions = []
+      parentheses = 0
+      bad_tag = False
+
+      if not "@" in tag_value:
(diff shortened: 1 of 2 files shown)


plugins/conditionalrestrictions.py
```diff
@@ -0,0 +1,210 @@
...
+    Plugin.init(self, logger)
+
+    self.ReYear = re.compile(r'20\d\d') # Update in 2099
+    self.currentYear = date.today().year
+
+    self.errors[33501] = self.def_class(item = 3350, level = 2, tags = ['highway', 'fix:chair'],
+        title = T_('Bad conditional restriction'),
+        detail = T_('''Conditional restrictions should follow `value @ condition; value2 @ condition2` syntax.
+Combined restrictions should follow `value @ (condition1 AND condition2)`.
+Parentheses `()` must be used around the condition if the condition itself contains semicolons `;`, i.e. `value @ (date;date)`.'''))
```
jocelynj, line 39: Should there be a link to https://wiki.openstreetmap.org/wiki/Conditional_restrictions in the description?

Famlam: I'll add 
`resource="https://wiki.openstreetmap.org/wiki/Conditional_restrictions"`
to the two errors involving syntax when you're finished with the review, thanks!

plugins/conditionalrestrictions.py
```diff
@@ -0,0 +1,210 @@
+#-*- coding: utf-8 -*-
```
Famlam, line 1: Reminder: rename to ConditionalRestrictions.py (2 capitals)

plugins/conditionalrestrictions.py
```diff
@@ -0,0 +1,210 @@
...
+                  {"highway": "residential", "access:conditional": "delivery @ (Mo-Fr 06:00-11:00,17:00-19:00;Sa 03:30-19:00))"},
+                  {"highway": "residential", "access:conditional": "delivery @ Mo-Fr 06:00-11:00,17:00-19:00;Sa 03:30-19:00);yes@wet"},
+                  {"highway": "residential", "access:conditional": "delivery @ (Mo-Fr 06:00-11:00,17:00-19:00;Sa 03:30-19:00;yes@wet"},
+                  {"highway": "residential", "access:conditional": "delivery @ (Mo-Fr 06:00-11:00,17:00-19:00;Sa 03:30-19:00));yes@wet"},
+                  {"highway": "residential", "access:conditional": "yes@()"},
+                  {"highway": "residential", "access:conditional": "yes@"},
+                  {"highway": "residential", "access:conditional": "@wet"},
+                  {"highway": "residential", "access:conditional": "no @ (2099 May 22 AND AND 2099 Oct 7)"},
+                  {"highway": "residential", "access:conditional": "no @ (2099 May 22 AND 2099 Oct 7 AND); delivery @ wet"},
+                  {"highway": "residential", "access:conditional": "no @ (2099 May 22 and 2099 Oct 7); delivery @ wet"},
```
jocelynj, line 200: I would add one testcase with no space, like: "no@wet".

plugins/conditionalrestrictions.py
```diff
@@ -0,0 +1,210 @@
...
+    for tag in tags_conditional:
+      tag_value = tags_conditional[tag]
+      conditions = []
+      parentheses = 0
+      bad_tag = False
+
+      if not "@" in tag_value:
+        err.append({"class": 33501, "subclass": 0 + stablehash64(tag + '|' + tag_value), "text": T_("Missing `@` in \"{0}\"", tag)})
+        continue
+
```
jocelynj, line 69: I'm wondering if we should also check that we have ` @ ` in the tag value.

Famlam: Can be done fairly easily in a crude way:
`if tag_value.count("@") != tag_value.count(" @ "):`

I'm not so sure how much that adds though; this is a thing that bots can fix safely without needs for humans (unlike 'and' which may hypothetically also occur in a date-time range comment), and it's only readability, not invalid.

Famlam: (To avoid having to make a new branch again I'll await your and/or Frodrigo's comment on this so I can put everything in one patch ;) )

plugins/conditionalrestrictions.py
Famlam: If we'd indeed add a warning for this I'll add a test case to the 'invalid' tests. If we don't, I'll add it to the 'valid' tests.
I'll also update the existing test cases to not trigger on a missing space, as the assert doesn't check which error is triggered, if we add the "missing spaces" warning now or in the future.

Famlam:
I've implemented the feedback as far as it doesn't involve adding new checks. New checks can also be added in a follow-up PR I think.

plugins/conditionalrestrictions.py
Famlam: Done

plugins/conditionalrestrictions.py
Famlam: Resolved for now, check must be moved if the no-space-check must be added in this PR

jocelynj:
Thanks for the update, I'm fine with the proposed changes. We can check later if we want to check spaces too.

@frodrigo, is it fine for you too?

frodrigo:
Merged but not deployed yet.
Thank you @Famlam for all the work on this.

Famlam:
You're welcome :)
Thank you both for the reviews!

Famlam:
Found an false positive, see #1378

PR #1378, opened 2021-12-27
Conditional restrictions: fix false positives with ; in the value

Famlam:
### The false positive
Spotted this on https://www.openstreetmap.org/way/303019393 : a conditional `turn:lanes` with a semicolon in the value:
`turn:lanes:forward:conditional = left|through|through;right @ (Mo-Fr 06:00-09:00)`
This triggered the warning: `Missing condition or parentheses`

Fixing this by only calling `continue` would prevent a warning from appearing in cases like:
`maxweight:hgv:conditional = 27000 lbs (axles=2); 41400 lbs @ (axles=3); 48600 lbs @ (axles=4); 55800 lbs @ (axles=5); 62550 lbs @ (axles=6); 69750 @ (axles>=7)`, which should have been:
`maxweight:hgv:conditional = 27000 lbs @ (axles=2); 41400 lbs @ (axles=3); 48600 lbs @ (axles=4); 55800 lbs @ (axles=5); 62550 lbs @ (axles=6); 69750 @ (axles>=7)`
Hence, I added an additional (previously unneeded) check that there are no `(` allowed in the conditional-value of any `*:conditional`. (_This is an assumption, please correct me if I'm wrong here! If you prefer to use no assumptions, no problem, I will then just delete those lines of code_)

The following case - however - will no longer trigger and thus become a false negative:
`motor_vehicle:conditional = no @ Fr 16:00-24:00; Sa,Su 00:00-24:00; no @ wet` (which should be:
`motor_vehicle:conditional = no @ (Fr 16:00-24:00; Sa,Su 00:00-24:00); no @ wet`.
However, I would only be able to distinguish this from valid values by actually interpreting which values are valid (i.e. programming that `Sa,Su 00:00-24:00; no` is not a valid value for `motor_vehicle`). All I could possibly do without that is adding a recommendation to use `(` and `)` for readability (see #1380)

### Warning message improvement
I improved / clarified the warning message about a missing condition or parentheses as it also triggers on some cases where an `@` is missing, for example:
`vehicle:conditional = no @ (Apr 29-Sep 29); delivery (Apr 29-Sep 29 7:00-10:00)` instead of
`vehicle:conditional = no @ (Apr 29-Sep 29); delivery @ (Apr 29-Sep 29 7:00-10:00)`

### For discussion
Additionally ~~(not yet programmed, first for discussion only)~~, I'm considering whether the `and`-check should be removed, as only 1 out of 7 cases so far it finds is actually correct, the other 6 are just invalid conditions (for which a separate warning would be nice - and extremely challenging - but this one is just confusing)
https://osmose.openstreetmap.fr/nl/issues/open?source=&class=33502&useDevItem=all&username=&bbox=&limit=500

### DIFF
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -38,10 +38,6 @@ class ConditionalRestrictions(Plugin):
 Combined restrictions should follow `value @ (condition1 AND condition2)`.
 Parentheses `()` must be used around the condition if the condition itself contains semicolons `;`, i.e. `value @ (date;date)`.'''),
         resource="https://wiki.openstreetmap.org/wiki/Conditional_restrictions")
-    self.errors[33502] = self.def_class(item = 3350, level = 3, tags = ['highway', 'fix:chair'],
-        title = T_('Combine conditions using `AND`'),
-        detail = T_('''`AND` (uppercase) is to be preferred over lowercase variants when combining restrictions.'''),
-        resource="https://wiki.openstreetmap.org/wiki/Conditional_restrictions")
     self.errors[33503] = self.def_class(item = 3350, level = 3, tags = ['highway', 'fix:chair'],
         title = T_('Expired conditional'),
         detail = T_('''This conditional was only valid up to a date in the past. It can likely be removed.'''),
@@ -98,9 +94,7 @@ Parentheses `()` must be used around the condition if the condition itself conta
             err.append({"class": 33501, "subclass": 2 + stablehash64(tag + '|' + tag_value), "text": T_("Mismatch in the number of parentheses in \"{0}\"", tag)})
             bad_tag = True
             break
-        elif c == ";" and parentheses == 0:
-          if not condition_started:
-            continue # value contains semicolon, example turn:lanes:conditional = left;right @ ...
+        elif c == ";" and parentheses == 0 and condition_started:
           tmp_str = tmp_str.strip()
           if len(tmp_str) == 0:
             err.append({"class": 33501, "subclass": 3 + stablehash64(tag + '|' + tag_value), "text": T_("Missing condition, `@` or parentheses in \"{0}\"", tag)})
@@ -134,9 +128,6 @@ Parentheses `()` must be used around the condition if the condition itself conta
               err.append({"class": 33501, "subclass": 4 + stablehash64(tag + '|' + tag_value), "text": T_("Missing condition before or after AND combinator in \"{0}\"", tag)})
               bad_tag = True
               break
-
-          if not bad_tag and tmp_cond.count(" AND ") != tmp_cond.upper().count(" AND "):
-            err.append({"class": 33502, "subclass": 0 + stablehash64(tag + '|' + tag_value), "text": T_("Lowercase version of `AND` in \"{0}\"", tag)}) # Recommendation to use uppercase "AND". Not a true error
 
       if bad_tag:
         continue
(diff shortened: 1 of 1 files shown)


plugins/ConditionalRestrictions.py
```diff
@@ -88,16 +88,22 @@ def way(self, data, tags, nds):
...
+            bad_tag = True
+            break
         elif c == ")":
           parentheses -= 1
           if parentheses == -1:
             err.append({"class": 33501, "subclass": 2 + stablehash64(tag + '|' + tag_value), "text": T_("Mismatch in the number of parentheses in \"{0}\"", tag)})
             bad_tag = True
             break
         elif c == ";" and parentheses == 0:
+          if not condition_started:
```
Famlam, line 102: Note to self: move to elif line; that way tmp_str remains intact if we ever want to use it

frodrigo:
Please, move the note part to the issue. To save it, rather than lost it here, once the PR is close.

Famlam:
> Please, move the note part to the issue. To save it, rather than lost it here, once the PR is close.

Created #1380 for the follow-up after this PR with the most-needed fixes has landed.
If you'd agree with me on removing the `and`/`AND` check, I'd like to do so before it goes live for everyone though

frodrigo:
> > Please, move the note part to the issue. To save it, rather than lost it here, once the PR is close.
> 
> Created #1380 for the follow-up after this PR with the most-needed fixes has landed. If you'd agree with me on removing the `and`/`AND` check, I'd like to do so before it goes live for everyone though

Yes, go head, as you have better experiences with this tag.

Famlam:
> Yes, go head, as you have better experiences with this tag.
First thing I'll do tomorrow :). I'll just delete it, unless you prefer to keep that code as a comment for now

Famlam:
Done, I've also added the results so far to issue #1380 in case we wish to re-evaluate this rule

frodrigo:
Thank, merged.

Famlam:
Great! I'll check tomorrow (or when the results are updating) if there are no new issues due to this:
https://github.com/osm-fr/osmose-backend/blob/f9ba3af71536d7ca101155d5883707a4123e3686/plugins/ConditionalRestrictions.py#L87-L90

I'll let you know! If no new issues, I think it can become live for everyone!

frodrigo:
> Great! I'll check tomorrow

It is merged, but not deployed, yet.

Famlam:
Except for very odd ones (below) that triggered an correspondingly odd message, I didn't find any false positives in the countries that were re-analyzed after the re-deployment with this fix, so in my opinion it can go live.

The odd ones:
`maxweight:conditional = (0 @ destination)`
`turn:lanes:conditional = (through|through|through|through|right) @ (Tu-Fr 15:30-17:00)`

Maybe the error messages on the lines below may in the future have to be rephrased to a more general `Bad synthax in {key}` to cover all cases without being misleading by suggesting i.e. an missing `@` if people ask, but for now I'm (personally) fine with it.
https://github.com/osm-fr/osmose-backend/blob/36b55ba474471920dd7577ed719a6a191cfedb49/plugins/ConditionalRestrictions.py#L88
https://github.com/osm-fr/osmose-backend/blob/36b55ba474471920dd7577ed719a6a191cfedb49/plugins/ConditionalRestrictions.py#L100
https://github.com/osm-fr/osmose-backend/blob/36b55ba474471920dd7577ed719a6a191cfedb49/plugins/ConditionalRestrictions.py#L114

frodrigo:
It has been deployed today, update in progress.

Issue #1380, opened 2021-12-27
Conditional restrictions: style recommendations


Famlam:
There are three style recommendations in the wiki for conditional restrictions for which we could hypothetically warn:

1. Warn about missing space around `@` (suggested by @jocelynj in #1376 in the review comments). I personally don't mind too much about the space being present and fear that the number of markers would overwhelm more important edits, while bots can fix this and it's unlikely that people will make mistakes due to the absence of spaces.

2. Warn about `and` instead of `AND`? Initially implemented in #1376, the first results showed too many "bad" conditionals like `opening_hours:conditional = 24/7 @ ('camper and fisherman')` where this would be confusing. (It's meant as a comment & it's an invalid condition: one cannot be a camper and a fisherman simultaneously, unless you throw your rod out of the camper window, probably an "or"-like statement is meant, which (in dummy tags) would be `24/7 @ camper; 24/7 @ fishing` ). Making it `AND` would make no sense. For now, the error was disabled in [this commit](https://github.com/osm-fr/osmose-backend/pull/1378/commits/1676e79bc71aef78e3bb8797691c3884723e24fc). These were the results after roughly 3/4th of the world was checked:

> - 2x: `opening_hours:conditional = 24/7 @ ('camper and fisherman')` **(invalid warning)**
> - 2x: `restriction:conditional = no_left_turn @ bus and trucks` **(invalid warning)**
> - 2x: `access:conditional = yes @ MAY 1st - OCT 31 and 06:00 - 22:00` (inefficient, but maybe valid?)
> - 1x: `restriction:conditional = no_left_turn @ truck and bus` **(invalid warning)**
> - 1x: `hgv:conditional = no@(22:00-07:00 and maxweight>=1.5)` (valid warning, although probably `weight` is meant)
> - 1x: `hgv:conditional = no @ (19:00-07:00 and weight>7.5)` (valid warning)
> - 1x: `charge:conditional = $4.50 @ (axle >= 6); $4.00 @ (axle = 5); $3.50 @ (axle = 4); $3.00 @ (axle = 3); $2.50 @ (axle = 2 and wheel >= 6)` (valid warning)
[...]

PR #1382, opened 2021-12-29
Warn about missing parentheses in conditional restrictions

Famlam:
**Warn about missing non-required parentheses in conditional restrictions**:
- if there are multiple conditions (`a@b;c@d`) and it is not the final condition
- if the last condition is not a simple (composed of a-z, A-Z, 0-9 and _) condition (i.e. `yes @ wet` is OK, but `yes @ 06:00-20:00` would request for parentheses)

i.e. 
`yes @ wet` will not warn
`yes @ 06:00-18:00` will warn until it is converted into `yes @ (06:00-18:00)`
`yes @ wet; yes @ snow` will warn until it is converted into `yes @ (wet); yes @ snow` or `yes @ (wet); yes @ (snow)`
`yes @ snow; yes @ 06:00-18:00` will warn until it is converted into `yes @ (snow); yes @ (06:00-18:00)`
`yes @ snow; yes @ (06:00-18:00)` will warn until it is converted into `yes @ (snow); yes @ (06:00-18:00)`
`yes @ (06:00-18:00); yes @ snow` will not warn

Additionally, warning about these will help people to resolve cases like `no @ Fr 16:00-24:00; Sa,Su 00:00-24:00; yes @ wet` which is currently interpret as `no @ Fr 16:00-24:00` + `Sa,Su 00:00-24:00; yes @ wet` and considered valid (as we do not interpret the value to be an valid value)

Involves #1380

### DIFF
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -40,7 +40,7 @@ Combined restrictions should follow `value @ (condition1 AND condition2)`.
 Parentheses `()` must be used around the condition if the condition itself contains semicolons `;`, i.e. `value @ (date;date)`.'''),
         resource="https://wiki.openstreetmap.org/wiki/Conditional_restrictions")
     self.errors[33502] = self.def_class(item = 3350, level = 3, tags = ['highway', 'fix:chair'],
-        title = T_('Suboptimal conditional restriction'),
+        title = T_('Improve style of conditional'),
         detail = T_('''Although valid, it is recommended to format conditional restrictions with:
 - spaces around the `@`;
 - uppercase `AND` (in combined restrictions);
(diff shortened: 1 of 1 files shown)


plugins/ConditionalRestrictions.py
```diff
@@ -38,6 +39,15 @@ def init(self, logger):
 Combined restrictions should follow `value @ (condition1 AND condition2)`.
 Parentheses `()` must be used around the condition if the condition itself contains semicolons `;`, i.e. `value @ (date;date)`.'''),
         resource="https://wiki.openstreetmap.org/wiki/Conditional_restrictions")
+    self.errors[33502] = self.def_class(item = 3350, level = 3, tags = ['highway', 'fix:chair'],
+        title = T_('Suboptimal conditional restriction'),
```
frodrigo, line 43: Please, can you reword this.

Famlam: Done!

frodrigo:
Can I merge this PR ?

Famlam:
Yes, although maybe first under the `useDevItem` flag?

frodrigo:
Yes. It still under `useDevItem`.

Famlam:
Thanks! I'll keep an eye on the error detection list and let you know approximately 1 day after the first results appear online (unless I spot false positives of course)!

Famlam:
I noticed it's online as of about 22 o'clock yesterday (in the time zone that's currently 15:10) and hence about half of the world has been analyzed already with this new rule.
So far no false positives :)!

Just wondering about one tiny thing: currently the extra text message is:
`Add parentheses around the condition in {key}` (note: condition is singular)
This also appears for i.e. `*:conditonal=a @ 00:00-10:00; b @ 10:00-24:00`, i.e. with two conditions.

Do you think I should:
1. Leave it as it is?
2. Change it to `Add parentheses around the condition(s) in {key}` to be more specific regarding `:conditional` rules with multiple conditions?
3. Add two separate warnings: 
 - `Add parentheses around the condition in {key}` for a single condition (`a@b`) and 
 - `Add parentheses around the conditions in {key}` for multiple conditions (`a@b;c@d`) in a tag?

My personal favorite option would be the second option. (Translations would just have to figure out a way to deal with both singular and plural)

frodrigo:
Yes, 2 is fine.

Issue #1301, opened 2021-07-14
Freshness of "Highway intersecting small/large water piece" in spain_castilla_y_leon


Crashillo:
Since a couple of days I'm noticing there's no updates on the `item=1070&class=4,5`.

For instance, most of the markers you can find in this bbox http://osmose.openstreetmap.fr/es/map/#username=Hugoren%20Martinako&item=1070&class=4%2C5&zoom=11&lat=42.9503&lon=-5.1588&level=1%2C2%2C3 have been already fixed one week ago. However, the tooltip says the issue was reported yesterday and the update_matrix displays an age < 1 for the Osmosis_Highway_VS_Building analyzer

How's this possible? Are not all classes (of an item) updated at the same time?

frodrigo:
All classes of a source (aka country/area) are updated at the same time.
http://osmose.openstreetmap.fr/es/issues/open?country=spain_castilla_y_leon&item=1070&source=&class=4&username=&bbox=

OSM extract is up to date.
https://buildbot.osmose.openstreetmap.fr/#/builders/236/builds/320/steps/1/logs/stdio (#173)

The extract look ok (not corrupted), and the contain the new bridge 82224058 and the edited highway 805343918 in version #2.

cc @jocelynj

Crashillo:
Notified also in *spain_galicia*

Crashillo:
Notified also for item 5080 in _Castilla y León_. It looks like the comparison is being done against an outdated extract.

http://osmose.openstreetmap.fr/en/map/#item=5080&zoom=17&lat=42.328434&lon=-5.993289&level=1&tags=&fixable=&issue_uuid=2f5e14da-c864-69f3-a280-a9562b9de539

These errors show [this way](https://www.openstreetmap.org/way/520976111)'s name as _Calle La Calzada_, nevertheless, such name was changed a week ago in favour of _Calle la Calzada_ (lowercase articles).

frodrigo:
Sorry for not looking in details at it before, but lake of time.
Now I am away of computer for holidays... in Castilla y León.

Crashillo:
It looks like it has been fixed. No idea why.

> Now I am away of computer for holidays... in Castilla y León.

I've seen some notes around... :grin:

frodrigo:
Strange because the graph doe not have variation
http://osmose.openstreetmap.fr/es/issues/open?country=spain_castilla_y_leon&item=1070&source=&class=4&username=&bbox=
[...]

PR #1385, opened 2022-01-04
Resume fix

frodrigo:
Fix resume mode, also delete in result file touched objects

May fix https://github.com/osm-fr/osmose-backend/issues/1301 and in particular https://github.com/osm-fr/osmose-backend/issues/1301#issuecomment-1001766824 and others.

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -355,14 +355,31 @@ ANALYZE {0}.buildings;
             types = {'N': 'node', 'W': 'way', 'R': 'relation'}
             for t, ids in self.already_issued_objects.items():
                 if ids:
-                    sql = "SELECT v.id, l.id FROM (VALUES ({0})) AS v(id) LEFT JOIN {1}s AS l ON v.id = l.id WHERE l.id IS NULL;".format('),('.join(map(str, ids)), types[t])
+                    sql = """
+-- Touched
+SELECT id
+FROM touched_{1}s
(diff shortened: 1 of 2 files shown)


analysers/Analyser_Osmosis.py
```diff
@@ -355,14 +355,31 @@ def dump_delete(self):
...
+                    sql = sql.format('),('.join(map(str, ids)), types[t])
                     self.giscurs.execute(sql)
                     for res in self.giscurs.fetchall():
                         self.error_file.delete(types[t], res[0])
         else:
             # Change
             for t in ["node", "way", "relation"]:
-                sql = "(SELECT id FROM actions WHERE data_type='{0}' AND action='D') UNION ALL (SELECT id FROM touched_{1}s) EXCEPT (SELECT id FROM actions WHERE data_type='{0}' AND action='C')".format(t[0].upper(), t)
+                sql = """
+SELECT id FROM actions WHERE data_type='{0}' AND action IN ('D', 'C')
```
jocelynj, line 378: Does it make a difference to add 'C'?

Created nodes/ways/relations shouldn't already be present in osmose database - I'm not sure that osm ids can be re-used for a new object.

In any case, I'm fine with this change.

frodrigo: This part of the code if for change mode. Is not used.
But recreated object may also change it at the same time.
The new 'C' come with the remove of the  `EXCEPT` part.

PR #1389, opened 2022-01-14
Allow connection to highway=steps

Famlam:
Fixes the false positive of i.e. http://osmose.openstreetmap.fr/en/issue/c4d6b033-c145-57bb-214e-ff070ad857a0 , where a highway=steps directly connects to a platform (https://www.openstreetmap.org/node/3848754089)

Although technically it's better to adjust the map (by having an ~1m footway between the platform and the steps), the error incorrectly suggests to change the role to an invalid value. Hence I think it's better to disable it for highway=steps.
Also adjusted the class=8 warning comparably, although I don't have a real-life example for this


Famlam:
Lint error seems unrelated to me?

frodrigo:
Ok, Thank you.

PR #1398, opened 2022-02-04
Improve numerical testing

Famlam:
- Add support for `max(speed/width/length/...):(forward/backward)`
- Check that integer-only keys are truly integers
- Add special check for `width=0` on highways (a former bug in iD as far as I could determine, about 8k worldwide)
- Added `diameter`, `distance`, `capital`, `heritage`, `step_count`, `maxaxleload`, `maxspeed:advisory`, `minspeed`
- Added `mm` and `lbs` as units
- fix `capacity` check on volumes (i.e. `man_made=storage_tank`)

### DIFF
--- a/plugins/Number.py
+++ b/plugins/Number.py
@@ -49,10 +49,18 @@ be used if the value is valid.''')
         self.errors[3093] = self.def_class(item = 3091, level = 2, tags = ['value', 'fix:chair'],
             title = T_('Suspicious value'),
             detail = T_(
-'''The tag expects a positive, round number without unit'''),
+'''The tag expects a positive, round number without unit.'''),
             fix = T_(
 '''Check that the value is accurate. Consider whether another tag should
 be used if the value is valid.''')
+        )
+        self.errors[3094] = self.def_class(item = 3091, level = 2, tags = ['value', 'fix:chair'],
+            title = T_('Unknown unit'),
+            detail = T_(
+'''The tag uses an unexpected unit.'''),
+            fix = T_(
+'''Check that you have used the correct unit and a supported abbreviation of the unit.'''),
+            resource = "https://wiki.openstreetmap.org/wiki/Map_features/Units"
         )
 
         self.tag_number = ["diameter", "distance", "ele", "height", "length", "width"]
@@ -64,7 +72,11 @@ be used if the value is valid.''')
             self.tag_number.extend(list(map(lambda tag: tag + i, tag_number_directional)))
         self.tag_number.extend(self.tag_number_integer)
 
-        self.Number = re.compile(u"^((?:-?[0-9]+(?:[.][0-9]+)?)|(?:[.][0-9]+))(?: ?(?:m|cm|mm|km|nmi|km/h|mph|knots|t|kg|st|lbs|lt|cwt)|'(?:[0-9]*(?:[.][0-9]+)?\")?|\")?$")
+        self.units = ["m", "cm", "mm", "km", "nmi", # distance excluding feet'inch"
+                      "km/h", "mph", "knots", #speed
+                      "t", "kg", "st", "lbs", "lt", "cwt"] #weight
+
+        self.Number = re.compile(u"^((?:-?[0-9]+(?:[.][0-9]+)?)|(?:[.][0-9]+))(?: ?([a-zA-Z23/]{1,5})|'(?:[0-9]*(?:[.][0-9]+)?\")?|\")?$")
         self.MaxspeedExtraValue = ["none", "default", "signals", "national", "no", "unposted", "walk", "urban", "variable"]
         self.MaxspeedClassValue = re.compile(u'^[A-Z]*:')
         self.MaxheightExtraValue = ["default", "below_default", "no_indications", "no_sign", "none", "unsigned"]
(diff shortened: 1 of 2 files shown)


Famlam:
Seems both lint & tests are broken somewhere else

plugins/Number.py
```diff
@@ -46,11 +46,27 @@ def init(self, logger):
 '''Check that the value is accurate. Consider whether another tag should
 be used if the value is valid.''')
         )
+        self.errors[3093] = self.def_class(item = 3091, level = 2, tags = ['value', 'fix:chair'],
+            title = T_('Suspicious value'),
+            detail = T_(
+'''The tag expects a positive, round number without unit'''),
```
frodrigo:
```suggestion
'''The tag expects a positive, round number without unit.'''),
```

plugins/Number.py
```diff
@@ -46,11 +46,27 @@ def init(self, logger):
...
+'''Check that the value is accurate. Consider whether another tag should
+be used if the value is valid.''')
+        )
+
+        self.tag_number = ["height", "width", "length", "ele"]
+        self.tag_number_integer = ["population", "admin_level"] # Only positive integers (no units) allowed
+        tag_number_directional = ["maxheight", "maxheight:physical", "maxwidth", "maxlength", "maxweight", "maxspeed", "minspeed"]
+
+        # Add suffixes to the directional tags, add everything to tag_number
+        for i in ["", ":forward", ":backward"]:
```
frodrigo, line 71: The `""` does not make duplicate in the list ?

plugins/Number.py
```diff
@@ -59,19 +75,28 @@ def node(self, data, tags):
...
                     )) and
                     not (i == "maxheight" and tags[i] in self.MaxheightExtraValue)
                 ):
                     return {"class": 3091, "subclass": 1, "text": T_("Concerns tag: `{0}`", '='.join([i, tags[i]])) }
-                elif m and i == "height" and float(m.group(1)) > 500:
+                if not m:
+                    continue
+
+                # Below here only tags containing numbers with/without unit remain
+                if i in self.tag_number_integer and str(int(abs(float(m.group(1))))) != tags[i]:
```
frodrigo, line 104: `float([ramdom])` raise an exception. You cannot use it like this to check for interger. You have to catch ValueError.

plugins/Number.py
Famlam: This one was actually already fixed in the 3th commit :)

plugins/Number.py
Famlam: It shouldn't, because I'm copying from `tag_number_directional` (a local variable just to have the items that need suffixes) to `self.tag_number`. The eventual content of `self.tag_number` is `['diameter', 'distance', 'ele', 'height', 'length', 'width', 'maxaxleload', 'maxheight', 'maxheight:physical', 'maxlength', 'maxspeed', 'maxspeed:advisory', 'maxweight', 'maxwidth', 'minspeed', 'maxaxleload:forward', 'maxheight:forward', 'maxheight:physical:forward', 'maxlength:forward', 'maxspeed:forward', 'maxspeed:advisory:forward', 'maxweight:forward', 'maxwidth:forward', 'minspeed:forward', 'maxaxleload:backward', 'maxheight:backward', 'maxheight:physical:backward', 'maxlength:backward', 'maxspeed:backward', 'maxspeed:advisory:backward', 'maxweight:backward', 'maxwidth:backward', 'minspeed:backward', 'admin_level', 'capital', 'heritage', 'population', 'step_count']` 

p.s.: I did it this way so someone could easily add i.e. `:hgv`, like
`for i in ["", ":forward", ":backward", ":hgv"]:`
to also test for i.e. `maxspeed:hgv` (and identical for other common items; I'm considering `:hgv`, `:bus` and `:trailer` as they're very common to be special-cases)

plugins/Number.py
Famlam: "match group 1" of the regex is: `((?:-?[0-9]+(?:[.][0-9]+)?)|(?:[.][0-9]+))`
If I'm not mistaken, that should always end up with either:
- `[number(s)]` (e.g. `5`)
- `[number(s)][dot][number(s)]` (e.g. `22.3`)
- `[dot][number(s)]` (e.g. `.93`)

Which are all float-parsable. I don't see any case that could throw a ValueError?
(It's different for `int()`, which will throw in case I give a float string as input, i.e. `int("3.0")` throws a `ValueError`. Therefore I first use `float()` inside `int()`)

Would you still prefer that I make it 1) a separate function and 2) wrap it in try/catch? No problem of course :)

Famlam:
Thanks for the review @frodrigo !

plugins/Number.py
frodrigo: You are right. It is ok.

plugins/Number.py
frodrigo: My bad, it is ok.

frodrigo:
Thank you.
Merged, but not deployed yet.

plugins/Number.py
Famlam: Thank you for double-checking though :) !

stephankn:
@frodrigo this broke a larger number of analyzers. Hopefully the fix is working, so could be fine tomorrow. Is someone already looking at the broken lint/tests? We have to rely on on it. It did successfully report the broken file:

```
plugins/Capacity.py:151:85  SyntaxError: unexpected EOF while parsing [pylama]
```

frodrigo:
> @frodrigo this broke a larger number of analyzers. Hopefully the fix is working, so could be fine tomorrow. Is someone already looking at the broken lint/tests? We have to rely on on it. It did successfully report the broken file:
> 
> ```
> plugins/Capacity.py:151:85  SyntaxError: unexpected EOF while parsing [pylama]
> ```

Yes sure. But nobody look at the lint/test issue yet.

PR #1399, opened 2022-02-05
Update Capacity.py - syntax error

Famlam:
Very sorry about this, I added this "test code" without testing it and clearly forgot a `)`
(And since test and lint are broken I didn't spot it there either.......)

@frodrigo


frodrigo:
Thank, not a problem.

frodrigo:
Should be deployed.

PR #1405, opened 2022-02-18
New check: conflicting tags for sides

Famlam:
Add a check to validate conflicts between `tag:left` / `tag:right` and `tag` or `tag:both`, or between `tag:both` and `tag` (I.e. `sidewalk=no` + `sidewalk:left=yes`)
- Accepts `key=right/left/both/yes` (where `key` can be i.e. `sidewalk`, `shoulder`, `bus_bay`, ...) together with `key:side=more_specific_value`
- Considers `none` and `no` equal (deprecated, yet still common `sidewalk` value)
- Ignores `opposite_*` as it's dubious whether opposite_* should still be included when a side is used (i.e. https://wiki.openstreetmap.org/wiki/Key:cycleway#Problems_with_opposite.2A_values, and https://wiki.openstreetmap.org/wiki/Key:cycleway:left and others; also applies to i.e. `busway=opposite*`; some argue `x=opposite_lane` -> `x:left=lane` + oneway tags)

I didn't add redundancy checks (`cycleway:left = no` + `cycleway=no`, for instance) as there's often a good reason for it. For example, backward compatibility.

Examples:
https://www.openstreetmap.org/way/633261201 (parking:lane*)
https://www.openstreetmap.org/way/6920751 (parking:lane*)
https://www.openstreetmap.org/way/126484131 (sidewalk*)
https://www.openstreetmap.org/way/387380311 (sidewalk*)
https://www.openstreetmap.org/way/882834583 (cycleway*)
https://www.openstreetmap.org/way/295384472 (cycleway*)


Famlam:
It seems to work ok, so I'm marking it as ready for review.
If you think it would be a useful addition to Osmose, would it be possible to run it under the useDevItem flag?

If you don't think it adds anything, feel free to reject :)

Also, is the following efficient enough or should I add something like `if not (":left" in tag or ":right" in tag or ":both" in tag): /continue` first?
```
            if tag[-5:] == ":both":
              tag_default = tag[0:-5] # special handling because :both_ways also exists
            else:
              tag_default = tag.replace(":both:", ":").replace(":left", "").replace(":right", "")
```

frodrigo:
Sure, it is interesting.

Note, there is a dedicated plugin for `parking:*`. So, parking should be excluded here to avoid duplicate similar reports.

I pushed a fix in master for the linting issue.

plugins/Highway_Sides.py
```diff
@@ -0,0 +1,96 @@
...
+from modules.OsmoseTranslation import T_
+from plugins.Plugin import Plugin
+from modules.Stablehash import stablehash64
+
+class Highway_Sides(Plugin):
+    def init(self, logger):
+        Plugin.init(self, logger)
+
+        self.errors[33601] = self.def_class(item = 3360, level = 2, tags = ['highway', 'fix:chair'],
+            title = T_('Conflicting tag values'),
```
frodrigo, line 31: Could be more explicit.

Famlam:
Thank you for the review! I'll work on it!

Famlam:
Thanks! It'll be under useDevItem first, right?

frodrigo:
> Thanks! It'll be under useDevItem first, right?

Yes, once deployed.

PR #1411, opened 2022-02-26
Fixes for item 3360: Conflicting tags for sides of the way

Famlam:
https://osmose.openstreetmap.fr/en/issues/open?source=&class=&useDevItem=all&username=&bbox=&item=3360

The new item implemented in #1405, currently under the useDevItem flag
~I'll add fixes to this PR when encountering them. Currently:~

- [x] Disable warning for `name=A/B` + `name:right=A` + `name:left=B`, that's handled by Name_Multiple (ex: [way](https://www.openstreetmap.org/way/34723456), [issue](https://osmose.openstreetmap.fr/en/issue/4f5cca51-0a6a-6295-3f27-65e139e0b3d4))
- [x] Disable warning for `cycleway=opposite` + `cycleway:[side]=no` (ex: [way](https://www.openstreetmap.org/way/4845489), [issue](https://osmose.openstreetmap.fr/en/issue/6a81d93d-d41e-6e60-72fa-3b002d08161d))
- [x] Disable warning for `sidewalk=separate` + `sidewalk:right=separate` + `sidewalk:left=no` (Quebec style, makes some sense. Ex: [way](https://www.openstreetmap.org/way/909293109), [issue](https://osmose.openstreetmap.fr/en/issue/3b5f224c-0615-f0d1-3095-0efd50099cf4))
- [x] Disallow the combination of `xxx=both` + `xxx:side=no` (limitation in previous implementation)

This should fix the majority of the false positives / duplicate warnings / dubious issues. After deployment of this PR (still under useDevItem) I'll re-check if there are no more false positives remaining.

### DIFF
--- a/plugins/Highway_Sides.py
+++ b/plugins/Highway_Sides.py
@@ -54,6 +54,8 @@ class Highway_Sides(Plugin):
             if tag_default in tags:
                 tt = tags[tag].replace("none", "no").replace("opposite_", "")
                 ttd = tags[tag_default].replace("none", "no").replace("opposite_", "")
+                if tag[0:5] == "name:" and tt in ttd:
+                    continue # 'name' probably equals "name:left" + "/" + name:right, handled by Name_Multiple
                 if tt != ttd and not ttd in allowedAlternativeValues:
                     err.append({"class": 33601, "subclass": 1 + stablehash64(tag), "text": T_("Conflicting values of \"{0}\" and \"{1}\"", tag_default, tag)})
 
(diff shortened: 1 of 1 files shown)


Famlam:
Hi @frodrigo , I went over the results and I think the current implementation is safe to be released to the public (i.e. no 'useDevItem' flag needed anymore).

It may be that in the future a check needs to be added to allow cases like `key=a;b` + `key:left=a` + `key:right=b`, but this is undocumented behaviour, so unless someone reports this as being the desired way of tagging, I would leave it as it is.
The same for 10 ways worldwide where there's something like `*_name:left=X` + `*_name:right=Y` + `*_name=X / Y` (with *=alt/short/old/...). Can be fixed if desired by making line 61 `if (tag[0:5] == "name:" or "_name:" in tag) and tt in ttd:`, but I guess it's not worth it.

PR #1418, opened 2022-03-13
French minor distribution substations opendata extension

flacombe:
I propose this change regarding `analyser_merge_power_substation_minor_FR.py`.

It enlarge the scope of available opendata from Enedis only to all French distribution operators.
https://opendata.agenceore.fr/explore/dataset/postes-de-distribution-publique-postes-htabt/

The new files contains some substations' names, when available.
The operator's name is got in `NOM_GRD` column. I will propose to add a supplementary column with proper operator OSM value in near future.

### DIFF
--- a/analysers/analyser_merge_power_substation_minor_FR.py
+++ b/analysers/analyser_merge_power_substation_minor_FR.py
@@ -59,6 +59,6 @@ class Analyser_Merge_Power_Substation_minor_FR(Analyser_Merge):
                         "substation": "minor_distribution"},
                     mapping2 = {
                         "operator": "NOM_GRD",
-                        "source": lambda fields: self.source+"/"+fields["NOM_GRD"],
+                        "source": lambda fields: self.source + " - " + fields["NOM_GRD"],
                         "name": lambda fields: fields["NOM_POSTE"] if fields["NOM_POSTE"] not in ("") else None},
                 )))
--- a/analysers/analyser_merge_power_substation_minor_FR.py
+++ b/analysers/analyser_merge_power_substation_minor_FR.py
@@ -48,7 +48,7 @@ class Analyser_Merge_Power_Substation_minor_FR(Analyser_Merge):
                     types = ["nodes", "ways"],
                     tags = [
                         {"power": "substation", "substation": "minor_distribution"},
-                        {"power": None, "transformer": ["distribution", "main"},
+                        {"power": None, "transformer": ["distribution", "main"]},
                         {"power": "substation", "substation": "distribution"}]),
                 conflationDistance = 50,
                 mapping = Mapping(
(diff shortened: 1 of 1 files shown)


analysers/analyser_merge_power_substation_minor_FR.py
```diff
@@ -28,33 +28,37 @@ class Analyser_Merge_Power_Substation_minor_FR(Analyser_Merge):
...
                 conflationDistance = 50,
                 mapping = Mapping(
                     static1 = {
                         "power": "substation",
-                        "voltage": "20000",
-                        "operator": "Enedis"},
+                        "voltage": "20000"},
                     static2 = {
                         "substation": "minor_distribution",
                         "source": self.source},
```
frodrigo, line 60: We can also use `NOM_GRD` for source rather than the data aggregator, cannot ?

frodrigo: @flacombe ping

flacombe: Oops, sorry for forgetting this.

I've concatenated data aggregator and NOM_GRD, since it's complementary work between both

analysers/analyser_merge_power_substation_minor_FR.py
```diff
@@ -59,6 +59,6 @@ def __init__(self, config, logger = None):
                         "substation": "minor_distribution"},
                     mapping2 = {
                         "operator": "NOM_GRD",
-                        "source": self.source+"/"+fields["NOM_GRD"],
+                        "source": lambda fields: self.source+"/"+fields["NOM_GRD"],
```
frodrigo, line 62: Better use a standard `;`, doesn't ?

flacombe: Not aware of any other mean to use a field in mapping expression. What are you referring to?

frodrigo: `source=foo;bar` and not `source=foo/bar`

flacombe: Oh well.
They're not separate sources, let's say consortium :)

analysers/analyser_merge_power_substation_minor_FR.py
```diff
@@ -28,33 +28,37 @@ class Analyser_Merge_Power_Substation_minor_FR(Analyser_Merge):
...
-                        "voltage": "20000",
-                        "operator": "Enedis"},
+                        "voltage": "20000"},
                     static2 = {
-                        "substation": "minor_distribution",
-                        "source": self.source},
+                        "substation": "minor_distribution"},
+                    mapping2 = {
+                        "operator": "NOM_GRD",
+                        "source": lambda fields: self.source+"/"+fields["NOM_GRD"],
```
frodrigo, line 62:
```suggestion
                        "source": lambda fields: self.source+" - "+fields["NOM_GRD"],
```

analysers/analyser_merge_power_substation_minor_FR.py
frodrigo: Ok. Replaced by " - ".

PR #1422, opened 2022-03-19
Add Luxembourgish

Famlam:
https://en.wikipedia.org/wiki/Luxembourgish -> ISO 639-1 = `lb`
Fixes "Default and local language name not the same" errors in Luxembourg (i.e. [error](http://osmose.openstreetmap.fr/en/issue/d6d9439c-4562-e51c-8d55-b03beae627af) - [way](https://www.openstreetmap.org/way/128541299) )

Does not fix issue 1415; that requires a change to [plugins/Name_PoorlyWrittenWayType*](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_PoorlyWrittenWayType.py) as well


frodrigo:
> Does not fix issue 1415; that requires a change to [plugins/Name_PoorlyWrittenWayType*](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_PoorlyWrittenWayType.py) as well

It will "fix" also this, but ignoring the rules, due to
https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_PoorlyWrittenWayType_Lang_fr.py#L28
The rules only apply to french mapping only area

It also disable all default based language rules.

Famlam:
Hmm, ok, sounds like a bigger project then

PR #1424, opened 2022-03-25
ConditionalRestrictions: condition check for date/time and property comparison

Famlam:
I noticed several bad [conditionals](https://wiki.openstreetmap.org/wiki/Conditional_restrictions) not following the specifications of [`opening_hours`](https://wiki.openstreetmap.org/wiki/Key:opening_hours) such as 
- ~[`access:conditional=yes@(JUL 01-JAN 31)`](https://www.openstreetmap.org/node/3474720058) (uppercase instead of `Jul` and `Jan`)~
- [`yes @ (9.00 - 10.00)`](https://www.openstreetmap.org/way/865014901) (`.` instead of `:`)
- ~[`no @ (sa 06:00-15:00)`](https://www.openstreetmap.org/way/260042752) (`sa` instead of `Sa`)~

Hence I extended the ConditionalRestrictions test with:
- Test for bad date/time based conditions (such as `2022-03-25`, while the syntax says it should be `2022 May 25`), reusing the code from the `opening_hours` check. Determining whether it's a date/time is based on a scoring system: as we intent to discover bad date/timespans, we cannot check for valid spans only.
- Basic testing for bad usage of `<`, `>` or `=`
- also: make `AND`-check easier to read
- also: no longer triggering "expired" warning for i.e. `length > 120200 mm` (i.e. a bigger number containing 2020)

Warning: I was unable to get pykopeninghours working locally, so I'll test this using the github tests only

Note: 2 out of the 3 examples aren't detected because of a call to `lower()` in TagFix_Opening_Hours

### DIFF
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -161,10 +161,15 @@ For example, use `no @ (weight > 5 AND wet)` rather than `no@weight>5 and wet`.'
               sanitized = self.sanitize_openinghours(c)
               if not sanitized['isValid']:
                 if "fix" in sanitized:
+                  # Note: contains both invalid as optimizable cases
+                  # i.e. Monday 08:00-20:00 -> Mo 08:00-20:00 (invalid -> valid)
+                  # but also 2022 Mar 01 - 2022 Nov 02 -> 2022 Mar 01 - Nov 02 (valid -> better & valid)
+                  # Hence, don't set bad tag; allow i.e. expired conditionals to be found
                   err.append({"class": 33504, "subclass": 6 + stablehash64(tag + '|' + tag_value + '|' + c), "text": T_("Involves \"{0}\" in \"{1}\". Consider using \"{2}\"", c, tag, sanitized['fix'])})
                 else:
+                  # Only invalid cases
                   err.append({"class": 33504, "subclass": 6 + stablehash64(tag + '|' + tag_value + '|' + c), "text": T_("Involves \"{0}\" in \"{1}\"", c, tag)})
-                bad_tag = True
+                  bad_tag = True
                 break
             else:
               # Validate vehicle property comparisons
@@ -246,7 +251,7 @@ class Test(TestPluginCommon):
                   {"highway": "residential", "access:forward:conditional": "no @ (10:00-18:00 AND length>5)"},
                   {"highway": "residential", "access:conditional": "no @ 2099"},
                   {"highway": "residential", "access:conditional": "no @ (weight >= 12020 AND length < 20200)"},
-                  {"highway": "residential", "access:conditional": "no @ (2099 May 22-2099 Oct 7)"},
+                  {"highway": "residential", "access:conditional": "no @ (2098 May 22-2099 Oct 7)"},
                   {"highway": "residential", "access:conditional": "no @ (2010 May 22-2099 Oct 7)"},
                   {"highway": "residential", "turn:lanes:forward:conditional": "left|through|through;right @ (Mo-Fr 06:00-09:00)"},
                  ]:
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -24,7 +24,7 @@ from plugins.Plugin import Plugin
 import re
 from datetime import date
 from modules.Stablehash import stablehash64
-from plugins.TagFix_Opening_Hours import sanitize_openinghours
+from plugins.TagFix_Opening_Hours.TagFix_Opening_Hours import sanitize_openinghours
 
 class ConditionalRestrictions(Plugin):
   def init(self, logger):
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -161,15 +161,10 @@ For example, use `no @ (weight > 5 AND wet)` rather than `no@weight>5 and wet`.'
               sanitized = self.sanitize_openinghours(c)
               if not sanitized['isValid']:
                 if "fix" in sanitized:
-                  # Note: contains both invalid as optimizable cases
-                  # i.e. Monday 08:00-20:00 -> Mo 08:00-20:00 (invalid -> valid)
-                  # but also 2022 Mar 01 - 2022 Nov 02 -> 2022 Mar 01 - Nov 02 (valid -> better & valid)
-                  # Hence, don't set bad tag; allow i.e. expired conditionals to be found
                   err.append({"class": 33504, "subclass": 6 + stablehash64(tag + '|' + tag_value + '|' + c), "text": T_("Involves \"{0}\" in \"{1}\". Consider using \"{2}\"", c, tag, sanitized['fix'])})
                 else:
-                  # Only invalid cases
                   err.append({"class": 33504, "subclass": 6 + stablehash64(tag + '|' + tag_value + '|' + c), "text": T_("Involves \"{0}\" in \"{1}\"", c, tag)})
-                  bad_tag = True
+                bad_tag = True
                 break
             else:
               # Validate vehicle property comparisons
@@ -253,7 +248,7 @@ class Test(TestPluginCommon):
                   {"highway": "residential", "access:forward:conditional": "no @ (10:00-18:00 AND length>5)"},
                   {"highway": "residential", "access:conditional": "no @ 2099"},
                   {"highway": "residential", "access:conditional": "no @ (weight >= 12020 AND length < 20200)"},
-                  {"highway": "residential", "access:conditional": "no @ (2099 May 22-2099 Oct 07)"},
+                  #{"highway": "residential", "access:conditional": "no @ (2099 May 22-2099 Oct 07)"}, #https://bugs.kde.org/show_bug.cgi?id=452236
                   {"highway": "residential", "access:conditional": "no @ (2010 May 22-2099 Oct 07)"},
                   {"highway": "residential", "turn:lanes:forward:conditional": "left|through|through;right @ (Mo-Fr 06:00-09:00)"},
                  ]:
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -152,8 +152,6 @@ For example, use `no @ (weight > 5 AND wet)` rather than `no@weight>5 and wet`.'
             # For simplicity: ignore.
             continue
 
-          if not module_PyKOpeningHours:
-            continue
           for c in condition_ANDsplitted:
             # Validate time-based conditionals
             if self.isLikelyOpeningHourSyntax(c):
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -36,7 +36,7 @@ class ConditionalRestrictions(Plugin):
     self.currentYear = date.today().year
     self.comparisonOperatorChars = ["<", "=", ">"]
     # Following regex are to detect likely (possibly misspelled) opening_hours syntaxes
-    self.ReWeekdayMonthOpeningH = re.compile(r'\b[A-Z][a-z]') # i.e. Mar or Mo
+    self.ReWeekdayMonthOpeningH = re.compile(r'\b[A-Z][a-z]+') # i.e. Mar or Mo
     self.ReMonthDayOpeningH = re.compile(r'\w\w\w[\s-]\d') # i.e. sep 1
     self.ReTimeOpeningH = re.compile(r'\d\D[\d-]|sun[sr][ei][ts]') # i.e. 5:30 or 5h30 or 5h-8h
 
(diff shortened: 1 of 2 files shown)


plugins/ConditionalRestrictions.py
```diff
@@ -24,14 +24,24 @@
...
     Plugin.init(self, logger)
 
-    self.ReYear = re.compile(r'20\d\d') # Update in 2099
+    self.ReYear = re.compile(r'\b20\d\d\b') # Update in 2099
     self.ReSimpleCondition = re.compile(r'^\w+$', re.ASCII)
+    self.ReAND = re.compile(r'\band\b', re.IGNORECASE)
     self.currentYear = date.today().year
+    self.comparisonOperatorChars = ["<", "=", ">"]
+    # Following regex are to detect likely (possibly misspelled) opening_hours syntaxes
+    self.ReWeekdayMonthOpeningH = re.compile(r'\b[A-Z][a-z]') # i.e. Mar or Mo
```
frodrigo, line 39: Comment is about 2 or 3 chars, while only 2.

plugins/ConditionalRestrictions.py
```diff
@@ -128,16 +141,42 @@ def way(self, data, tags, nds):
...
-            if len(splittedANDpart.strip()) == 0:
-              err.append({"class": 33501, "subclass": 4 + stablehash64(tag + '|' + tag_value), "text": T_("Missing condition before or after AND combinator in \"{0}\"", tag)})
-              bad_tag = True
-              break
+          condition_ANDsplitted = list(map(str.strip, self.ReAND.split(condition)))
+          # Check the position of AND is ok
+          if "" in condition_ANDsplitted:
+            err.append({"class": 33501, "subclass": 4 + stablehash64(tag + '|' + tag_value), "text": T_("Missing condition before or after AND combinator in \"{0}\"", tag)})
+            bad_tag = True
+            break
```
frodrigo, line 151: Why break and not continue. Other conditions can still be checked, cannot ?

plugins/ConditionalRestrictions.py
```diff
@@ -161,6 +200,33 @@ def way(self, data, tags, nds):
     if err != []:
       return err
 
+  def isLikelyOpeningHourSyntax(self, condition):
+    # Use a scoring system to determine the likelyness of the condition being time/date based
+    # Not perfect, i.e. 'Mar' and '24/7' will fall through (and bad cases like JAN-APR are thus also not detected)
```
frodrigo, line 205: You can just add a case for `24/7`.

plugins/TagFix_Opening_Hours.py
```diff
@@ -38,28 +38,36 @@ def init(self, logger):
...
     def node(self, data, tags):
-        return self.sanitize_tags(tags)
+        if 'opening_hours' not in tags:
+            return
+        sanitized = self.sanitize_openinghours(tags['opening_hours'])
+        if not sanitized['isValid']:
+            if "fix" in sanitized:
+                return {"class": 32501, "subclass": 0, 'fix': {'opening_hours': sanitized['fix']}}
+            else:
+                return {"class": 32501, "subclass": 1, 'text': {'en': 'The opening_hours value is invalid and could not be parsed'}}
```
frodrigo, line 64: It was already like this. But translation should be used here: `T_()`.

frodrigo:
@dfaure FYI this PR reuse OpeningHours validator and make changes to `TagFix_Opening_Hours.py`.

plugins/ConditionalRestrictions.py
```diff
@@ -161,6 +200,33 @@ def way(self, data, tags, nds):
...
+    if len(condition) < 5:
+      return False # wet, snow, ..., not a time range
+    score = 0
+    treshold = 3
```
dfaure-kdab, line 209: typo: threshold

plugins/TagFix_Opening_Hours.py
```diff
@@ -38,28 +38,36 @@ def init(self, logger):
...
-
         parser = OpeningHours()
-        parser.setExpression(tags['opening_hours'])
+        parser.setExpression(openinghours_value)
         if parser.error() == Error.SyntaxError or parser.error() == Error.IncompatibleMode:
-            return {"class": 32501, "subclass": 1, 'text': {'en': 'The opening_hours value is invalid and could not be parsed'}}
+            return {"isValid": False}
         sanitized_field = parser.normalizedExpression()
         # Ignore trivial changes that can be fixed by bots rather than humans like spaces, case errors, etc
         simplify = lambda s: s.replace(' ', '').replace('24:00', '00:00').replace('0', '').lower()
```
dfaure-kdab, line 51: I would be OK with the .lower() being removed, this is from before my time.

It would however create more work for people applying fixes suggested by osmose, obviously.

plugins/ConditionalRestrictions.py
Famlam: Added a `+`
(For the regex (which isn't used to obtain capture groups, only to detect the presence of such a string) it doesn't change anything, in `Ma` `Mar` and `March` [2 of which are invalid and should be detected] it simply matched the first two chars as it's open-ended)

plugins/ConditionalRestrictions.py
Famlam: True. I'd prefer throwing a single "invalid" warning where it's easy to do; one would only see one marker on the front-end anyway for a single object, and the errors may be related to a single typo. Would you like me to adjust it?
(In that case I'll also have to add `condition` to the hash as multiple conditions might have the same error)

plugins/ConditionalRestrictions.py
Famlam: Done! (Although it'll always validate as true since it's a hard-coded, valid value)

plugins/TagFix_Opening_Hours.py
Famlam: Done! Also added the ` ` ` as it wasn't translated yet anyway.
Not sure if there's a better way to deal with the `T_()`-call in the tests?

plugins/ConditionalRestrictions.py
Famlam: Thanks!

plugins/TagFix_Opening_Hours.py
Famlam: I'd prefer to remove `lower()` too, as in practice I'm not aware of bots fixing case-errors (pretty error-prone I think, since anything can be in the fallback rule, like `9:00-18:00 || "by appointment made on monday"` shouldn't become `9:00-18:00 || "by appointment made on Mo"`, so any bot-fix should be very careful.
@frodrigo ?

Famlam:
Thanks for the review both!

@dfaure-kdab @dfaure would you perhaps also have any recommendations to improve this? Or would this require changes to kOpeningHours first?
https://github.com/osm-fr/osmose-backend/blob/5639e73bc10fd70322704e31897e1e170c4502db/plugins/ConditionalRestrictions.py#L164-L166
`normalizedExpression()` doesn't differentiate between invalid-but-convertible and valid-but-optimizable cases. Therefore it's not possible to give separate warnings for each of those cases.
(Which is actually also an issue with TagFix_Opening_Hours validation; some reported cases are truely invalid, others are just optimizable, but all trigger error 3250: `Invalid Opening Hours`)

plugins/TagFix_Opening_Hours.py
dfaure-kdab: Right. My semi-interactive "bot fixer" actually uses KOpeningHours for parsing and fixing ;)

Famlam: Haha oops, ok, in that case the question is indeed only if we should add case-errors to the UI or not

plugins/TagFix_Opening_Hours.py
frodrigo: I think you can just ignore the text part.

Famlam: You mean removing it from the error? (I agree if so as the title already says it's invalid, just checking I understand it correctly)

dfaure-kdab:
> `normalizedExpression()` doesn't differentiate between invalid-but-convertible and valid-but-optimizable cases. Therefore it's not possible to give separate warnings for each of those cases.

The KOpeningHours API is `normalizedExpression()` for "not as per grammar, must be fixed", and `simplifiedExpression()` for "valid but optimizable". I don't think we use the latter in osmose.

Famlam:
> > `normalizedExpression()` doesn't differentiate between invalid-but-convertible and valid-but-optimizable cases. Therefore it's not possible to give separate warnings for each of those cases.
> 
> The KOpeningHours API is `normalizedExpression()` for "not as per grammar, must be fixed", and `simplifiedExpression()` for "valid but optimizable". I don't think we use the latter in osmose.

Interesting that normalizedExpression() 'says' `2099 May 22-2099 Oct 07` must be modified and suggests it should be `2099 May 22-Oct 07` instead. (I believe the first one is 100% valid per https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#selector:month (specifically: `<monthday_range>` = `<date_from> [ <date_offset> ] - <date_to> [ <date_offset> ]`). See the current failing test. Is that a bug in kOpeningHours then?
(Also: https://osmose.openstreetmap.fr/nl/issue/591998f1-7f24-e071-2798-76419a48222b (`[-1,1]` vs `[1,-1]`) are both valid I think?)

dfaure-kdab:
> > > `normalizedExpression()` doesn't differentiate between invalid-but-convertible and valid-but-optimizable cases. Therefore it's not possible to give separate warnings for each of those cases.
> > 
> > 
> > The KOpeningHours API is `normalizedExpression()` for "not as per grammar, must be fixed", and `simplifiedExpression()` for "valid but optimizable". I don't think we use the latter in osmose.
> 
> Interesting that normalizedExpression() 'says' `2099 May 22-2099 Oct 07` must be modified and suggests it should be `2099 May 22-Oct 07` instead. (I believe the first one is 100% valid per https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#selector:month (specifically: `<monthday_range>` = `<date_from> [ <date_offset> ] - <date_to> [ <date_offset> ]`). See the current failing test. Is that a bug in kOpeningHours then?
> (Also: https://osmose.openstreetmap.fr/nl/issue/591998f1-7f24-e071-2798-76419a48222b (`[-1,1]` vs `[1,-1]`) are both valid I think?)

You're right, those are valid. KOpeningHours parses the string into its own data structures (i.e. it "understands" it, since it can also tell us when something is closed or open), and then we ask it to serialize it back to a string. So when the grammar offers multiple ways to express the same thing, it ends up normalizing to a single syntax, unless it stores additional info about which syntax was used. We already modified KOpeningHours to store such "useless" ;) information so that it doesn't turn 20:00-26:00 into 20:00-02:00 even though they are equivalent, to store "off" and "closed" differently, and to store "24/7" and "open" differently...

So yeah, I suppose you could say this is a bug / missing feature in KOpeningHours, if you think it should differenciate `2099 May 22-2099 Oct 07` from `2099 May 22-Oct 07` and only suggest the latter in `simplifiedExpression()`. Do you want to file an issue for KOpeningHours? See https://bugs.kde.org/buglist.cgi?list_id=2021757&product=KOpeningHours&resolution=--- for the current issue list.

Famlam:
> So yeah, I suppose you could say this is a bug / missing feature in KOpeningHours, if you think it should differenciate `2099 May 22-2099 Oct 07` from `2099 May 22-Oct 07` and only suggest the latter in `simplifiedExpression()`. Do you want to file an issue for KOpeningHours? See https://bugs.kde.org/buglist.cgi?list_id=2021757&product=KOpeningHours&resolution=--- for the current issue list.

Done, thanks! https://bugs.kde.org/show_bug.cgi?id=452236

frodrigo:
Sorry I merged the wrong PR !

plugins/ConditionalRestrictions.py
frodrigo: Yes I prefer. The Osmose frontend map is not the only to show reported issues.

plugins/TagFix_Opening_Hours.py
frodrigo: No, just from the test.

Famlam:
> Sorry I merged the wrong PR !
I think you can revert it via the button next to the merge link in this PR? (If you wish)

plugins/TagFix_Opening_Hours.py
Famlam: Currently there's no error given if you use `mo` or `JAN` rather than Mo or Jan. Would we want to enable this, so they become visible in the UI?

plugins/TagFix_Opening_Hours.py
Famlam: Ok, will try (later today)

plugins/ConditionalRestrictions.py
Famlam: Will do (later today)

Famlam:
I think you only reverted it locally in a branch? I can't make edits now ;)

plugins/TagFix_Opening_Hours.py
frodrigo: Yes we can.
The initial idea was to avoid boring contributor with obvious fix than parser are totally aware of.

frodrigo:
> I think you only reverted it locally in a branch? I can't make edits now ;)

I think the simplest if to make a new branch from mainstream and a new PR. Sorry.

Famlam:
> > I think you only reverted it locally in a branch? I can't make edits now ;)
> 
> I think the simplest if to make a new branch from mainstream and a new PR. Sorry.

No problem to make a new branch, but the revert is not in master yet.

frodrigo:
> > > I think you only reverted it locally in a branch? I can't make edits now ;)
> > 
> > 
> > I think the simplest if to make a new branch from mainstream and a new PR. Sorry.
> 
> No problem to make a new branch, but the revert is not in master yet (so making a new branch would already contain the changes and get a conflict as soon as the revert is merged)

I just try the revert button. But not validate it. We keep the master as is. Just continue on a new branch.

plugins/ConditionalRestrictions.py
Famlam: Implemented in #1431

plugins/TagFix_Opening_Hours.py
Famlam: You were right! Implemented in #1431

plugins/TagFix_Opening_Hours.py
Famlam: Implemented in #1431

PR #1429, opened 2022-04-03
Review OSM object mapping in analyser_merge_public_transport_FR_idfm #1377

frodrigo:
Implement https://github.com/osm-fr/osmose-backend/issues/1377#issuecomment-1082077960

".0" issue was previously fixed.


nlehuby (approved):
seems good to me :+1: thanks for the update

I've updated the doc on OSM wiki, maybe add a link
https://wiki.openstreetmap.org/wiki/France/Transports_en_%C3%8Ele-de-France#R.C3.A9f.C3.A9rences_Open_Data

PR #1431, opened 2022-04-06
Followup of 1424 - review comments of ConditionalRestrictions update

Famlam:
PR #1424 was accidentally merged to master, before all review comments were implemented
This should address the review comments:
- use continue rather than break
- remove `lower()` normalization
- remove `text: *` from tests


frodrigo:
Thank you.

frodrigo:
It is deployed and first results are here.

Famlam:
Thanks! So far the results look good, except for the [bug I filed with kOpeningHours](https://bugs.kde.org/show_bug.cgi?id=452236) about normalization of years (for example, https://osmose.openstreetmap.fr/nl/issue/34462a7c-900b-788e-edfc-e64f16f7cfc7 ; at the moment already 120 instances of this, about 3% of the errors)

Famlam:
I've proposed #1435 to temporary workaround the aforementioned bug, other than that I'm happy with the results

PR #1433, opened 2022-04-11
Check values for change:lanes

Famlam:
Adding a check to warn for `change:lanes(:direction)` with:
- unknown values (I allowed "" although officially that doesn't exist, as people do use it quite often*); [example](https://www.openstreetmap.org/way/700426183)
- values in impossible situations (changing to a non-existing lane); [example](https://www.openstreetmap.org/way/379391693)

(And fix a typo in "valid")

As the tag is not too commonly used, the number of errors will likely be low.

(* probably meaning "default", although there's no default for this tag defined)


frodrigo:
Thank you.

frodrigo:
Deployed

Famlam:
Results look good :)

PR #1435, opened 2022-04-14
Workaround for kOpeningHours normalization bug with years

Famlam:
For some regions, this false positive makes up over half of the detections, like https://osmose.openstreetmap.fr/nl/issues/open?source=24767&item=3350&class=33504&limit=500 (although quite a lot of them were expired). And overall, it's over 5% of all errors so far. Hence a workaround seems desirable I would say, until kOpeningHours had the time to resolve this issue on their side (I can't)

This regex simply converts `YYYY MMM DD - YYYY MMM DD` to `YYYY MMM DD - MMM DD`


frodrigo:
Ok, let's take this for now.

Famlam:
Seems this removed about 5.5% of the items as the numbers dropped from 3965 to 3747 warnings (assuming there were not too many changes to the database while the analysers re-ran). 
There's still 96 instances of e.g. `Jan 01-Jan 31` being simplified to Jan 01-31 (and similar month/date ranges): 1 in France, 4 in Germany, 2 in Greece, 1 in Italy, 70 in Netherlands, 1 in Spain and 16 in USA. I'll just mark these as False Positives for now as these are rule simplifications rather than wrong (and I guess same-month rules are not added as commonly as same-year rules).
@dfaure I've added this as a comment to the same [bug report](https://bugs.kde.org/show_bug.cgi?id=452236) as I guess they're very related?

PR #1438, opened 2022-04-20
ConditionalRestrictions: warn if there's something before '(' or after ')'

Famlam:
Example: https://www.openstreetmap.org/node/8365099879

Presently `a @ b(c)` (an invalid expression) is treated as `a @ bc`. If `bc` is an invalid condition (example: b=`_`, c=`2022 May 04`, giving `_2022 May 04`, which will trigger the invalid date warning) , the error will say the time is invalid. (Someone marked the above way as a false positive, hence I spotted this...)

Similar issue occurs if there's something after the closing `)`. This is especially common with `*:lanes:conditional`, when people use:
`tag:lanes:conditional = a @ (b)||; c @ (d)||` instead of
`tag:lanes:conditional = a|| @ (b); c|| @ (d)`
Example: https://www.openstreetmap.org/way/302595376

Instead, the current PR makes sure the entire condition is marked as invalid when there's something before or after parentheses. (So also `no @_(wet)_` will be found, currently that escapes detection)


Famlam:
Results look good!

PR #1442, opened 2022-04-30
Mapillary analysers: propose human-readable date format

Famlam:
Based on: https://github.com/osm-fr/osmose-backend/commit/d576d057a7a7926a53ec057882c81aa6a7e243bd#commitcomment-72445086

Also format the date in the proposed fix for Mapillary based analysers. Currently it displays a numerical string only. 
Careful: untested


frodrigo:
Looks good. Thank you.

PR #1444, opened 2022-04-30
Disable analysers for non-existing speed limits in NL

Famlam:
Speed limit signs of 35, 45 and 110 do not exist in The Netherlands, hence any detection must be a false positive. Therefore, I think disabling those is the easiest.


frodrigo:
Disabling think are always a good think ! Thank you.

Famlam:
This must've been the fastest merge ever ;)

PR #1451, opened 2022-05-07
Check the values of access tags

Famlam:
- Check for access tags using uncommon values
- Check for `access=unknown`
- Check for `access=[vehicle type]`


frodrigo:
> Hi @frodrigo , no problem if you don't have the time at the moment, but I was wondering if you could review this PR if you think it's interesting?

Yes, it is a time issue to review this PR.

Famlam:
No problem ;). Thanks for the update!

frodrigo:
Thank you. Interesting refactoring.

PR #1452, opened 2022-05-11
Add tex-mex to list of cuisine exceptions for TagFix_BadValue

emersonveenstra:
`cuisine` was added in https://github.com/osm-fr/osmose-backend/commit/50ee5f88952a0f7f15bfb11f231c4382bfba67ee, however `tex-mex` is a valid value that's specified in the NSI for Taco Bell, Chili's, etc. The PR adds an exception so they are skipped in the check.


frodrigo:
Thank

Famlam:
Overlooked it, my bad 🙃

frodrigo:
Deployed

PR #1454, opened 2022-05-12
Whitelist `highway=slipway`+`leisure=slipway` conflict

emersonveenstra:
According to https://wiki.openstreetmap.org/wiki/Tag:leisure=slipway this is a valid use case, and iD has it set as the preset for [Slipway (Drivable)](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/leisure/slipway_drivable.json) so I think it's worth whitelisting here.


Famlam:
Don't you mean highway=service?
https://wiki.openstreetmap.org/wiki/Tag:service%3Dslipway

emersonveenstra:
Yep, I do mean that 😅 I fixed it.

Famlam (approved):

frodrigo:
Merged, but not deployed yet.

PR #1462, opened 2022-05-23
Do not use UNLOGGED tables

frodrigo:
Data from unlogged tables are lost on server restart.


jocelynj:
Won't this decrease performance and increase SSD usage, by having more data going to disk?

frodrigo:
Yes. But this is on open data cache tables. Low frequency update.

PR #1481, opened 2022-06-15
Accept name=A / B with name:left and name:right

Famlam:
Fixes the first half of 1480

- Accepts `/` in `name` if `name` consists of `name:left / name:right` or the other way around
- Add a dedicated warning (not "multiple names", which is confusing for a typo) if the names don't match

### DIFF
--- a/plugins/Name_Multiple.py
+++ b/plugins/Name_Multiple.py
@@ -46,6 +46,12 @@ proper names of both part.'''),
 '''Some streets have not the same names on the each side, especially if
 the houses by both sides are on different city. In this case, you can use
 the tag `name:left=*` and `name:right=*`.'''))
+        self.errors[50301] = self.def_class(item = 5030, level = 2, tags = ['name', 'fix:chair'],
+            title = T_('Conflicting names'),
+            detail = T_(
+'''This is a street with different names on each side of the way, given by the keys `name:left` and `name:right`.
+These names do not match with the value of `name`.
+The tag `name` should have the value `name:right / name:left` or `name:left / name:right`.'''))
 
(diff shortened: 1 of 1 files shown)


plugins/Name_Multiple.py
```diff
@@ -82,6 +88,15 @@ def way(self, data, tags, nds):
             if "public_transport" in tags and tags["public_transport"] in ["platform", "stop_position"]:
                 return
 
+            if "name:left" in tags and "name:right" in tags:
```
frodrigo, line 91: Need to check `name` tag also exist.

plugins/Name_Multiple.py
```diff
@@ -82,6 +88,15 @@ def way(self, data, tags, nds):
...
+                # name:left and name:right may be combined in name, separated by a /
+                nameparts = [n.strip() for n in tags["name"].split("/")]
+                if tags["name:left"] in nameparts and tags["name:right"] in nameparts and len(nameparts) == 2:
+                    return
+                else:
+                    return {"class": 50301, "subclass": 1,
+                            "fix": {"+": {"name": tags["name:right"] + " / " + tags["name:left"]}}}
```
frodrigo, line 98:
```suggestion
                            "fix": {"~": {"name": tags["name:right"] + " / " + tags["name:left"]}}}
```

Famlam:
Thanks for the review!

frodrigo:
Thank.

Famlam:
So far, results seem good. A few cases where people made the understandable shortening of i.e `Route X / Route Y` to `Route X / Y`, but I'd say these can be marked as false positives

PR #1482, opened 2022-06-15
Ignore proposed roads

Famlam:
Example: http://osmose.openstreetmap.fr/nl/issue/f8eb8970-9b0a-3a9e-b836-fb34208c5846
Proposed roads do not reflect the current situation, but rather a future situation, hence it makes sense that they do not connect to the present road network exactly.

### DIFF
--- a/analysers/analyser_osmosis_highway_almost_junction.py
+++ b/analysers/analyser_osmosis_highway_almost_junction.py
@@ -45,7 +45,7 @@ FROM (
     FROM
       highways
     WHERE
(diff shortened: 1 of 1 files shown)


frodrigo:
I think it is better to use `is_construction`.
https://github.com/osm-fr/osmose-backend/blob/master/analysers/Analyser_Osmosis.py#L49

Famlam:
Thanks! Please check if I did it correctly, as I'm not too familiar with SQL (nor the scope of its tables/variables).

frodrigo:
Looks good, thank.

PR #1484, opened 2022-06-17
Exclude only one key of each pair in analyser_osmosis_tag_typo

Famlam:
- Exclude only one of each pair if possible, allowing typos to be found with the other key (as described & requested to be done with existing items in https://github.com/osm-fr/osmose-backend/pull/1470 :
> Interesting. Maybe can do it for other keys to un-blacklist some.

- Added `ship`, so that the (much more common key) shop could be used for matching
- Remove `static_caravan(s)` combination as almost all `static_caravan` have been converted to `static_caravans`
- Remove `colo(u)r` as all `color` have been converted to `colour` in a bulk edit early this year

### DIFF
--- a/analysers/analyser_osmosis_tag_typo.py
+++ b/analysers/analyser_osmosis_tag_typo.py
@@ -73,6 +73,7 @@ FROM
             'room', -- vs roof, rooms
             'house', -- vs horse
             'addr2', 'addr3',
+            'kern', -- vs kerb
             'lock_name', -- vs loc_name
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank. Let try it.

Famlam:
So far, only `wine` (vs. `line`) and `fdot` (a Florida tag; vs `foot`) seem to be new issues due to this change, so I'm positively surprised :). Included them in the open PR

PR #1487, opened 2022-06-19
Small fixes for multiple names, tag typos, and access value checks

Famlam:
Accidentally did everything in one branch, but since it's all simple things, I left it like this. Please let me know if I should split it up instead.

This fixes:
- Add `addr4`/`addr5` to exceptions (common in [Austria](https://osmose.openstreetmap.fr/nl/issues/open?source=13138&item=3150&class=1))
- Add `draft` and `fridge` to exceptions (false positives vs `craft` and `bridge`)
- Add `fdot` (Florida specific) and `wine` to exceptions (false positives vs `foot` and `line`)
- Fix a typo with `unleashed`
- Add a trap warning and prevent potential bogus messages for `name:left/name:right/name` if `name:left` equals `name:right`
- Fix the error when `tram=tram` is encountered (and similar x=x values, except when it's a valid value). ([Example](https://osmose.openstreetmap.fr/nl/issue/a36b97b9-9352-5cff-a17f-77b2e197d289))
- Add two dedicated `canoe` access values

### DIFF
--- a/analysers/analyser_osmosis_tag_typo.py
+++ b/analysers/analyser_osmosis_tag_typo.py
@@ -92,6 +92,7 @@ FROM
             'side', -- vs site, sides, hide
             'draft', -- vs craft
             'fridge', -- vs bridge
+            'moved', -- vs moped
             'name_1', 'name_2', 'name_3', 'name_4', 'name_5', 'name_6', 'name_7', 'name_8', 'name_9', -- Tiger mess
 
             -- Regional hiking/cycling/etc. routes. Lesser used ones like 'rhn' for horse riding trigger false positives:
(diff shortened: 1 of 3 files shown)


frodrigo:
Note, you can check linting localy with ./tools/pytest.sh lint

Famlam:
Thanks, I think this was just lazyness as I added that check directly via the web interface of Github 😇

I've included a couple of fixes / improvements in this PR, but they all only affect small numbers, so no rush to review / deploy :). 

(Also I think it may be good to ignore all numbered keys in tag_typo altogether (name1, addr1, shop1, contact1, ...), as they rarely are real typos, but usually "by design". However I don't know how to do regexes/patterns in SQL, so I'll leave that to you if you'd like).

frodrigo:
Thank.

PR #1489, opened 2022-06-22
Fix linz vs line false positive

Famlam:
`linz` (Land Information New Zealand) is a NZ governmental organisation, used in many keys in that country (causing 855 false positives that want to convert `line` to `linz`)

Let's hope this is the last one. Most of the UK, Texas, Florida and North-Carolina still have to update since the merge of #1484


frodrigo:
Thank.

> Most of the UK, Texas, Florida and North-Carolina still have to update since the merge of https://github.com/osm-fr/osmose-backend/pull/1484

Yes. Not deployed yet.

PR #1497, opened 2022-07-22
Update analyser_merge_milestone_FR_metropole.py

didier2020:
missing filter on "logical" ref , not real milestone can be seen on the road


didier2020 (commented):
space before ":"

PR #1500, opened 2022-07-25
Add dedicated citycentre geocoding feature as to get rid of Addok

flacombe:
i'd like to propose a new French city centre geocoding ability as to solve Addok problems when looking for centroïd or city hall point of a city.

Currently `analyser_merge_power_plant_FR.py` uses `Analyser_Merge_Geocode_Addok_CSV.py` which doesn't deliver appropriate results with city names/codes only.
As Addok client may be useful for further anlyzers, I've added `Analyser_Merge_Geocode_City_CSV.py` to use official https://api.gouv.fr/documentation/api-geo.
I can remove `Analyser_Merge_Geocode_Addok_CSV.py` if required by contribution guidelines.

Instead of Addok geocoder, official Geo API doesn't support csv data to make bulk geocoding, so the input file is read line by line. Is it ok?

### DIFF
--- a/analysers/Analyser_Merge_Geocode_City_CSV.py
+++ b/analysers/Analyser_Merge_Geocode_City_CSV.py
@@ -31,7 +31,7 @@ import json
 
 class Geocode_City_CSV(Source):
 
-    def __init__(self, source, logger, citycode = None, delimiter = ',', encoding = 'utf-8'):
+    def __init__(self, source, logger, citycode, delimiter = ',', encoding = 'utf-8'):
         self.source = source
         self.delimiter = delimiter
         self.citycode = citycode
--- a/analysers/Analyser_Merge_Geocode_City_CSV.py
+++ b/analysers/Analyser_Merge_Geocode_City_CSV.py
@@ -51,7 +51,7 @@ class Geocode_City_CSV(Source):
         infile = self.source.open()
         header = None
         i = 0
-        
+
         for linestr in infile:
             outline = linestr.split(self.delimiter)
 
@@ -70,7 +70,7 @@ class Geocode_City_CSV(Source):
                 rData = json.loads(r.text)
 
                 outline.extend(rData["centre"]["coordinates"])
-            
+
             outfile.write(self.delimiter.join(outline))
             i += 1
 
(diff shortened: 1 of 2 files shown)


analysers/Analyser_Merge_Geocode_City_CSV.py
```diff
@@ -0,0 +1,77 @@
...
+        self.delimiter = delimiter
+        self.citycode = citycode
+        self.encoding = encoding
+        self.logger = logger
+
+    def __getattr__(self, name):
+        return getattr(self.source, name)
+
+    def open(self):
+        return open(downloader.update_cache('citycoded://' + self.source.fileUrl, 60, fetch=self.fetch))
```
flacombe, line 45: @frodrigo can you confirm this is appropriate to cache geocoding results, right?

Geocoded files may have several lines involving a given city.
I expect the API will be queried once and then use cached results for further records.
I'm not sure how cache is handled exactly.

frodrigo: No. It is to cache the result of the geocoding.

flacombe: So for several queries `GET https://geo.api.gouv.fr/communes/74010?fields=centre&format=json&geometry=centre` the API will be queried only once, right?

frodrigo: By the result I mean the global CSV. The output of the processing.

If you want to cache query you can use https://github.com/osm-fr/osmose-backend/blob/master/modules/downloader.py#L140 or one of previous in the module.

flacombe: Got it, thanks.

I've edited the code to use urlread instead of Python requests.Session.

frodrigo:
You can fetch all the data and make the join locally.

https://geo.api.gouv.fr/communes?zone=metro&fields=nom,code,codesPostaux,siren,codeEpci,codeDepartement,codeRegion,population&format=geojson&geometry=centre

frodrigo:
Merged manually, using full download of city geocoding data.

flacombe:
Great, thank you for the final touch on this point

Power plant analysis is now at the best functional level it can run

frodrigo:
> Great, thank you for the final touch on this point
> 
> Power plant analysis is now at the best functional level it can run

I think there is a next level. The file content IRIS code.

PR #1502, opened 2022-07-27
disable merge bus stops in Kerala, India :(

### LINKED ISSUE #1486: Merge Kerala Bus Stops DataSource no more available
The geojson about the bus stop positions in Kerala is no more available from 200d.

@nlehuby do you know about it status ? It will be back or we need to disable this analyser ?

nlehuby:
Fix #1486


frodrigo:
Ok. Thank.

PR #1503, opened 2022-07-27
Update merge public_transport platforms in IDF, France

nlehuby:
* `ref:FR:STIF` are int
* disable for trains as you may have several `ref:FR:STIF` on one platform

about #1377


nlehuby:
I've disabled the analyser for trains because we do have several objects in IDFM data for the same platform:
- in the case of a central platform serving two lanes (one on each side), there are generally two IDFM objects. Maybe the `railway=platform_edge` object (not used in Île-de-France) is closer in terms of modeling?
- an extra object for each station which I have no idea of the use and which creates false positives
- SNCF maintains the `ref:FR:STIF` on all the platforms of the stations of its perimeter so there are only those of RATP RER stations which are still missing in OSM

frodrigo:
Thank.

Issue #1506, opened 2022-08-02
False positive for parking:condition:[side] without parking:lane:[side] value


rskedgell:
The tests in Highway_Parking_Lane.py reflect the situation before Proposed features/Parking lane conditionals (https://wiki.openstreetmap.org/w/index.php?oldid=2289049) was approved on 2022-01-21. 

This moved the values no_parking and no_stopping for parking:lane=* to parking:condition=*, so a false positive if raised for e.g. parking:condition:*=no_parking|no_stopping, but not for the now deprecated parking:lane:*=no_parking|no_stopping.

frodrigo:
Ok, should be fixed here
https://github.com/osm-fr/osmose-backend/blob/master/plugins/Highway_Parking_Lane.py

frodrigo:
Done.

blef00fr:
Osmose continues to report a warning when parking:condition:[side] is used without parking:lane:[side] 

According to the wiki at Key:parking:condition: "Both can be used independently of each other, they do not necessarily require the other "

So, what about this warning?

Discostu36:
Issue should be re-opened, I still get this warning.

Famlam:
Unlike the parking:condition (which often requires ground surveys), parking:lane can mostly be deduced from areal imagery. 
So I personally, when I fixed the bug in the initial post, kept this check in despite being an "allowed" situation. It's a good suggestion for things that could be improved easily. (Just like for instance `signals should have a railway:signal:direction=* tag`, `Unfinished power minor line`, `Unmapped surveillance camera`, `Operator can be completed with operator:wikidata` and several other errors).

frodrigo:
Can we close this one now?

blef00fr:
As far as my comment is concerned, I think so.

Issue #1479, opened 2022-06-14
False positive: parking:lane:[side]=no + parking:condition:[side]=no_stopping


vincentvd1:
Osmose gives a false positive when a highway is tagged with both `parking:lane:[side]no + parking:condition:[side]=no_stopping`.

No parking lane does not equal no parking. Depending on the situation (for example in the Netherlands), the absence of a parking lane does not prohibit you to stop. In the Netherlands, if a cycleway is present, you are not allowed to stop but on other roads it is for example allowed to stop to let out a passenger but it is not allowed to park, while still not saying anything about the presence of parking lanes.  

Link to a location: http://osmose.openstreetmap.fr/en/map/#zoom=17&lat=52.267875&lon=6.150364&item=xxxx&level=1,2,3&tags=&fixable=&issue_uuid=6d062189-672c-b3da-2438-d8ea93f08b32
item details: http://osmose.openstreetmap.fr/en/issue/6d062189-672c-b3da-2438-d8ea93f08b32

frodrigo:
Done

PR #1511, opened 2022-08-08
Make parking:*:[side] reflect current wiki

Famlam:
Related to: #1506, #1479

- `no_stopping`, `no_parking` are now `parking:condition:[side]` values, `fire_lane` is a `parking:condition:[side]:reason` value
- warn if `parking:condition:[side]` is set to a value belonging to `parking:lane:[side]`
- make error 31615 more clear in case `parking:lane:[side]` is set to a `parking:condition:[side]` or `parking:condition:[side]:reason` value
- Check `parking:lane:[left/right]` + `parking:condition:both` mismatches (as it was very easy to implement)
- Fix part of the check for error 31616 which never returned true

Marked as ready for review, but it may be good to wait until the checks are fixed...

### DIFF
--- a/plugins/Highway_Parking_Lane.py
+++ b/plugins/Highway_Parking_Lane.py
@@ -27,6 +27,11 @@ class Highway_Parking_Lane(Plugin):
 
     def init(self, logger):
         Plugin.init(self, logger)
+
+        self.parkingLaneValues = ["parallel", "diagonal", "perpendicular", "marked", "yes", "no", "separate"]
+        self.parkingConditionValues = ["free", "ticket", "disc", "residents", "customers", "private", "disabled", "loading", "no_parking", "no_standing", "no_stopping", "no"] # or custom
+        self.parkingConditionReasonValues = ["bus_stop", "crossing", "driveway", "dual_carriage", "fire_lane", "junction", "loading_zone", "narrow", "passenger_loading_zone", "priority_road", "street_cleaning", "turnaround", "turn_lane"] # or custom
+
         self.errors[31611] = self.def_class(item = 3161, level = 3, tags = ['highway', 'parking', 'fix:imagery'],
             title = T_('Bad parking:lane:[side]'),
             detail = T_(
@@ -41,9 +46,8 @@ class Highway_Parking_Lane(Plugin):
 sides.'''))
         self.errors[31615] = self.def_class(item = 3161, level = 3, tags = ['highway', 'parking', 'fix:chair'],
             title = T_('Bad parking:lane:[side] value'),
-            fix = T_(
-'''See values at
-[`parking:lane=*`](https://wiki.openstreetmap.org/wiki/Key:parking:lane).'''))
+            fix = T_('''Use any of the following values: `{0}`''', "`, `".join(self.parkingLaneValues)),
+            resource = "https://wiki.openstreetmap.org/wiki/Key:parking:lane")
         self.errors[31616] = self.def_class(item = 3161, level = 3, tags = ['highway', 'parking', 'fix:survey'],
             title = T_('parking:condition:[side] without parking:lane:[side] value'),
             detail = T_(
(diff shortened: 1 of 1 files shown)


frodrigo:
> Marked as ready for review, but it may be good to wait until the checks are fixed...

Master is fixed, you can now rebase.

plugins/Highway_Parking_Lane.py
```diff
@@ -41,9 +46,8 @@ def init(self, logger):
 sides.'''))
         self.errors[31615] = self.def_class(item = 3161, level = 3, tags = ['highway', 'parking', 'fix:chair'],
             title = T_('Bad parking:lane:[side] value'),
-            fix = T_(
-'''See values at
-[`parking:lane=*`](https://wiki.openstreetmap.org/wiki/Key:parking:lane).'''))
+            fix = T_('''Use any of the following values: `{0}`''', "`, `".join(self.parkingLaneValues)),
```
frodrigo, line 49:
```suggestion
            fix = T_('''Use any of the following values: `{0}`.''', "`, `".join(self.parkingLaneValues)),
```

plugins/Highway_Parking_Lane.py
```diff
@@ -58,6 +62,10 @@ def init(self, logger):
 '''A parking condition is set for a parking:lane:[side] value that indicates
 that the parking area is mapped separately. Any parking conditions should
 be tagged on that object instead.'''))
+        self.errors[31619] = self.def_class(item = 3161, level = 3, tags = ['highway', 'parking', 'fix:survey'],
+            title = T_('Bad parking:condition:[side] value'),
+            fix = T_('''Use any of the following values: `{0}`''', ", ".join(self.parkingConditionValues)),
```
frodrigo, line 67:
```suggestion
            fix = T_('''Use any of the following values: `{0}`.''', ", ".join(self.parkingConditionValues)),
```

plugins/Highway_Parking_Lane.py
```diff
@@ -67,7 +75,7 @@ def way(self, data, tags, nds):
 
         if (("parking:condition:right" in tags and not "parking:lane:right" in tags and not "parking:lane:both" in tags) or
             ("parking:condition:left" in tags and not "parking:lane:left" in tags and not "parking:lane:both" in tags) or
-            ("parking:condition:both" in tags and not "parking:condition:both" in tags)):
+            ("parking:condition:both" in tags and not "parking:lane:both" in tags and not ("parking:lane:left" in tags and "parking:lane:right" in tags))):
```
frodrigo, line 78: I am not sure, but the condition should not be :
```suggestion
            ("parking:condition:both" in tags and not "parking:lane:both" in tags and not "parking:lane:left" in tags and not "parking:lane:right" in tags)):
```

Famlam: Hi Frodrigo,
Thanks for the review!

I think either could be good.

- In both cases, `parking:condition:both` without `parking:lane:*` is considered bad
- In both cases, `parking:condition:both` together with `parking:lane:both` is considered good
- In both cases, `parking:condition:both` together with `parking:lane:left` + `parking:lane:right` is considered good

However, in my case, `parking:condition:both` with only `parking:lane:left` OR `parking:lane:right` is considered 'bad' (because one side `parking:lane` is not specified, only `parking:condition`); in your case, the latter is considered good too.

I can live with either of the two options, so you can choose which one you prefer :).

frodrigo: ok, keep as is.

frodrigo:
Thank. Merged.
Deploy already occur today. So will be in next batch.

Famlam:
Thanks! No problem :)

Quick question: https://osmose.openstreetmap.fr/nl/issues/false-positive?item=3161&class=31611 shows there's 505 false positives, but the list with the false positives at the bottom does not load. Is this known? (I usually track the list of false positives a couple of weeks after making an edit, to see if many new issues arise)

frodrigo:
I make some changes in the frontend API, now it is not synchro with top of the page. It comes, but take times.

Famlam:
In case you think we're too early with also warning about the old syntax (given the complaints in 1476 about the marker carpet), easiest is to comment out lines 99 and 101 (and making it `pass`). I'm going to be on a little holiday for the remainder of the week so can't fix it myself :S

PR #1516, opened 2022-08-11
Suggestion for wording changes

Famlam:
I think these texts are slightly more clear in describing what the issue is. Feel free to reject if you disagree :)

I'll be on holiday starting in a few hours, so it'll take a few days before I can reply. Sorry about that :)


frodrigo:
Yes, better, thank you.

PR #1520, opened 2022-08-16
Increase priority of guaranteed errors in Highway_Parking_Lane

Famlam:
I checked the results after the recent modification. All seemed fine.
However, I spotted that suggestions for optimization got the same priority as 100% certainly errors (like 31611)

I didn't touch the others as they are either suggestions (31614,31616 for example: not wrong, just suboptimal), error prone (31617 and above: they can have custom values added in the future) or currently very very common (31615, due to the `no_stopping`/`no_parking` move). However, 31611 is very unlikely to contain false positives: anything but `left`, `right` or `both` is unlikely to be added, ever.

`31611` (unknown [side]) has 508 occurances at the moment


frodrigo:
I personal think it is a too technical and complicated subject to annoying contributor with that at level 1. We can promote it to level 2 if you want.

Famlam:
Sure! But feel free to reject as well if you disagree, it's a minor thingy :)

frodrigo:
Thank. Merged

Issue #1499, opened 2022-07-22
Warn about dead-end drive-throughs


Famlam:
I today discovered someone in my area mixed up driveway and drive-through. Given that many people use the presets, it's not that difficult to do so (both start with drive*), so I suspect there might be many more of these.

Unlike `service=driveway` highways, `service=drive-through` ways should by definition not be dead-end (unless there's a turning_circle at the end of it, or something similar). Hence I thought it may be possible to have a check similar to [analyser_osmosis_highway_deadend](https://github.com/osm-fr/osmose-backend/blob/c6c0e8a4598495c3b21c308081dbda92be0d1717/analysers/analyser_osmosis_highway_deadend.py) to also find any dead-end drive-through ways, which are then probably mistagged driveways.

I can't code this myself given that I don't 100% understand the SQL...

PR #1521, opened 2022-08-16
Implement drive-through check

Famlam:
First attempt to implement #1499 (based on 'if you want it, do it' I figured I had to try it myself)

Check for potential driveway/drive-through mix-ups.
- Filter on ways with `highway=service` & `service=drive-through`
- ~Pass if oneway (as then there's usually an exit; if not, another error will trigger)~
- Pass if there's a turning facility at the end
- Pass if there's a building entrance at the end
- ~Pass if under construction~
- If all conditions fulfilled and dead-ended: probably not a drive-through

*Important note:*
I have no experience with SQL and I can't get any SQL analysers to run at home, so this code is untested and probably needs thorough inspection prior to merging, if you wish.

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -242,7 +242,7 @@ GROUP BY
     nodes.geom
 HAVING
     COUNT(*) = 1
-)
+) AS t
 """
 
 class Analyser_Osmosis_Highway_DeadEnd(Analyser_Osmosis):
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -242,6 +242,7 @@ HAVING
     LEFT JOIN highways AS drivethrough ON
         drivethrough.linestring && t.geom AND
         t.nid = ANY(drivethrough.nodes) AND
+        drivethrough.id = t.wid AND
         drivethrough.tags?'service' AND
         drivethrough.tags->'service' = 'drive-through'
 WHERE
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -275,7 +275,7 @@ lead to somewhere and in particular to a network of minor roads.''')),
 '''Review the classification of road or draw the local road network.'''))
         self.classs[3] = self.def_class(item = 1210, level = 1, tags = ["highway", "fix:chair"],
             title = T_('One way inaccessible or missing parking or parking entrance'))
-        self.classs_change[4] = self.def_class(item = 1210, level = 2, tags = ['highway', 'fix:chair'],
+        self.classs_change[5] = self.def_class(item = 1210, level = 2, tags = ['highway', 'fix:chair'],
             title = T_('Unconnected drive-through'),
             detail = self.merge_doc(detail, T_(
 '''Drive-throughs are usually not dead-ended. Make sure the full drive-through path was drawn, including i.e. turning circles and covered areas.
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -211,8 +211,7 @@ sql40 = """
 SELECT
     wid,
     nid,
-    ST_AsText(geom),
-    highway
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -206,6 +206,45 @@
...
+    nodes.geom,
+    MIN(way_ends.highway) AS highway
+FROM
+    {0}highway_ends AS way_ends
+    JOIN highways ON
+        highways.linestring && way_ends.linestring AND
+        way_ends.nid = ANY(highways.nodes) AND
+        NOT highways.is_construction AND
+        NOT highways.is_oneway AND -- oneway implies an exit
+        highways.tags->'service' = 'drive-through'
```
frodrigo, line 229: You need the full network to check connectiveness.

frodrigo: You nee to filter the highway_ends, not the highway.

Famlam: Thank you! I see what you mean. 
I tried to implement it differently now, but I don't know if the new method is correct

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -206,6 +206,48 @@
...
+        (NOT nodes.tags?'entrance' OR nodes.tags->'entrance' = 'no') -- i.e. indoor part not drawn
+WHERE
+    way_ends.highway = 'service'
+GROUP BY
+    nodes.id,
+    nodes.geom
+HAVING
+    COUNT(*) = 1
+) AS t
+    LEFT JOIN highways AS drivethrough ON
```
frodrigo, line 242: Need to exclude the highway it self, need not it?

```sql
drivethrough.id = t.wid
```

Famlam: I've added the check, thank you!

Do I understand correctly that this would be needed in case `t` is a connection between a service road and a non-service road (like residential)?

Famlam: My impression was that `t` already contains only nodes that are not connected to any other highways (and not turning circles and such)

Here I indeed just wanted to have a filter on drive-throughs. As far as I understand the code that generates `highway_ends` (and thus `t`), `t` doesn't contain the tags, correct?

But as mentioned, I'm really a beginner in SQL

Famlam: (Note that I do have some concerns about the WHERE block in the inner FROM, i.e. are connections between service and unclassified filtered out of `t` or not)

frodrigo: So. There is a more efficient way to do this.
The idea is to start with `highways` as highways_services filtered on required services, join on `highway_ends`, then join on all (other) highways and then join on `nodes`. It can filter the checked highways from start of the process.

On a such complex query you can ignore `{0}` and the `.format` part.

To work on this, run osmose once with --no-clean option and jump to the database to get postgres shell to test query directly. Once here you can run query, with timing (`\timing`) and more important `ANALYSER` (`ANALYSER SELECT ...`).

Famlam: I'm afraid I'll have to park this for a while then, until I figure out how to run the SQL queries locally.

frodrigo: Look at this https://github.com/osm-fr/osmose-backend/tree/master/docker#develop-on-osmose-using-docker

Famlam: Oh wow, it took me 3 hours to discover that in 
`./osmose_run.py --no-clean --country=monaco --analyser=osmosis_highway_deadend`
you have to use the filename *excluding* the prefix `analyser_`...
And I haven't noticed that any earlier time that I tried to build it either...

Famlam: Thank you very much for the feedback earlier!
Now that I can run them locally, I rewrote the tests; tested them, and they seem to work great on the 7 areas where I ran them.
I'll commit them in a new pull request to allow more easy comparison of what has changed from the original code: #1528

PR #1528, opened 2022-08-27
Implement drive-through check (attempt 2)

### LINKED ISSUE #1499: Warn about dead-end drive-throughs

Famlam:
New attempt based on feedback in #1521 (new PR as it's mostly a rewrite; keeps things clean :) )
Fixes #1499  -  Check for potential driveway/drive-through mix-ups.

1. Start from `highway=service` + `service=drive-through` (and not a oneway, which is just a shortcut as oneways are theoretically never dead-ended)
2. Get the end nodes of this way
3. Filter out (via `LEFT JOIN` + `WHERE ... IS NULL`) the end nodes that are connected to any other way
4. Filter out (the last `JOIN`) i.e. turning circles (which could make it a valid dead-end drivethrough) and entrances (if e.g. the indoor part isn't mapped)

Results of some countries [attached](https://github.com/osm-fr/osmose-backend/files/9437767/results.zip); no false positives so far; several "bad" drive-throughs found. (Note: in my local tests I disabled the class=1&2&3 analysers, hence their results are not listed)

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -211,7 +211,7 @@ SELECT
   wid,
   nid,
   (SELECT ST_AsText(geom) FROM nodes WHERE id = nid)
-FROM 
+FROM
   (
   SELECT
     id AS wid,
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -206,6 +206,43 @@ WHERE
   r.nid IS NULL
 """
 
+sql40 = """
+SELECT
+  wid,
+  nid,
+  (SELECT ST_AsText(geom) FROM nodes WHERE id = nid)
+FROM 
+  (
+  SELECT
+    id AS wid,
+    linestring,
+    nodes
+  FROM
+    highways as drivethroughs
+  WHERE
+    drivethroughs.highway = 'service' AND
+    drivethroughs.tags?'service' AND
+    drivethroughs.tags->'service' = 'drive-through' AND
+    NOT drivethroughs.is_oneway
+  ) as highways_drivethrough
+  JOIN highway_ends AS way_ends ON
+    highways_drivethrough.linestring && way_ends.linestring AND
+    way_ends.nid = ANY(highways_drivethrough.nodes)
+  LEFT JOIN highways AS other_highways ON
+    way_ends.nid = ANY(other_highways.nodes) AND
+    other_highways.id != highways_drivethrough.wid
+  JOIN nodes ON
+    nodes.id = way_ends.nid AND
+    (NOT nodes.tags?'highway' OR (
+      nodes.tags->'highway' != 'turning_circle' AND
+      nodes.tags->'highway' != 'turning_loop' AND
+      nodes.tags->'highway' != 'mini_roundabout'
+    )) AND
+    (NOT nodes.tags?'entrance' OR nodes.tags->'entrance' = 'no') -- i.e. indoor part not drawn
+WHERE
+  other_highways.id IS NULL
+"""
+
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -206,6 +206,43 @@
...
+  JOIN nodes ON
+    nodes.id = way_ends.nid AND
+    (NOT nodes.tags?'highway' OR (
+      nodes.tags->'highway' != 'turning_circle' AND
+      nodes.tags->'highway' != 'turning_loop' AND
+      nodes.tags->'highway' != 'mini_roundabout'
+    )) AND
+    (NOT nodes.tags?'entrance' OR nodes.tags->'entrance' = 'no') -- i.e. indoor part not drawn
+WHERE
+  other_highways.id IS NULL
```
frodrigo, line 243: Better SQL, simplest. Also join on highway_ends ends geom (point) and not linestring.

Better explain
```diff
-Nested Loop  (cost=963.52..57287.91 rows=1 width=48)
+Nested Loop  (cost=963.81..42095.65 rows=1 width=48)
```
@Famlam please review my suggest SQL code anyway. I do not have validated the result.


```suggestion
SELECT
  drivethroughs.id,
  nid,
  ST_AsText(nodes.geom)
FROM
  highways AS drivethroughs
  JOIN highway_ends ON
    drivethroughs.linestring && highway_ends.geom AND
    highway_ends.nid = ANY(drivethroughs.nodes)
  LEFT JOIN highways AS other_highways ON
    other_highways.linestring && highway_ends.geom AND
    highway_ends.nid = ANY(other_highways.nodes) AND
    other_highways.id != drivethroughs.id
  JOIN nodes ON
    nodes.id = highway_ends.nid AND
    (NOT nodes.tags?'highway' OR (
      nodes.tags->'highway' != 'turning_circle' AND
      nodes.tags->'highway' != 'turning_loop' AND
      nodes.tags->'highway' != 'mini_roundabout'
    )) AND
    (NOT nodes.tags?'entrance' OR nodes.tags->'entrance' = 'no') -- i.e. indoor part not drawn
WHERE
  drivethroughs.highway = 'service' AND
  drivethroughs.tags?'service' AND
  drivethroughs.tags->'service' = 'drive-through' AND
  NOT drivethroughs.is_oneway AND
  other_highways.id IS NULL```

Famlam: Hi @frodrigo thank you very much!
I tested your code on my local computer and it works perfect. (I checked with the same countries as in the initial message and the results are identical to the results of my initial code)

Would you be able to edit your commit suggestion to remove the three ``` at the end? Then it can be directly committed via the suggestion

frodrigo: > Would you be able to edit your commit suggestion to remove the three ``` at the end? Then it can be directly committed via the suggestion
done

Famlam: Applied, thanks!

frodrigo:
Thank you, merged.

Famlam:
Results look nice so far!

PR #1534, opened 2022-08-31
Clarify error message

bubeck:
The previous error message indicates, that there is an error in the way type (attribute "highway") and not in the name. Clarify this.


frodrigo:
Deployed

PR #1542, opened 2022-09-04
Don't remove oneway from roundabout if oneway=no

Famlam:
Prevent proposing to remove `oneway` from a roundabout if `oneway=no`. 

The proper fix would be to tag `junction=circular` instead of `junction=roundabout` and keep the `oneway` tag, however, I believe a check for this case belongs in JOSM ( filed: https://josm.openstreetmap.de/ticket/22343 ), hence I'm just adding the exception to Osmose.
Example issue: https://osmose.openstreetmap.fr/nl/issue/38388bc5-48a0-8cdb-7891-81cbb7f3fc4b
815 cases worldwide

(p.s.: theoretically this entire plugin could be in mapcss, but I find it too debatable if having oneway=yes is bad to propose it to JOSM myself. In fact, I always disable this check when I'm looking at Osmose issues)

### DIFF
--- a/plugins/TagRemove_Roundabout.py
+++ b/plugins/TagRemove_Roundabout.py
@@ -35,7 +35,7 @@ The tag `name=*` must be present if this is the name of the roundabout
 and not a road connected, same thing for `ref=*`.'''))
 
     def way(self, data, tags, nds):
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank you.

frodrigo:
Deployed

PR #1544, opened 2022-09-06
duplicated geom with unusal tags

didier2020:
unusual tags like 'SHAPE', 'LENGTH', only tag "area" or "name" are duplicated and or with another with "normal" tag
tested on indonesia and some african country


frodrigo:
Thank. Merged, but not deployed, yet.

frodrigo:
Deployed

didier2020:
thank's 
fix issue for thailand and paraguay

PR #1545, opened 2022-09-06
only area on way

didier2020:
a lot of in indonesia, i don't know why


frodrigo:
Need to adjust the `detail` in the doc section.

didier2020:
je n'ai pas trouvé ou se trouve "detail"

frodrigo:
```python
            detail = T_(
'''An object without relevant tag (no other tag than `source`,
`created_by`, `note:qadastre` nor `name`) nor a relation member with a
role.'''),
```

You can just push a new commit on the branch to fix it.

Issue #1538, opened 2022-09-03
Item #1050 : Reverse Roundabout


altilunium:
Is there a database where Osmose keeps track of whether a country is "drive on the right" or "drive on the left"? 

Here is an example from [Indonesia](https://osmose.openstreetmap.fr/en/map/#zoom=18&lat=-6.357587&lon=106.930055&item=xxxx&level=1&issue_uuid=fd861f98-431f-9adf-7868-b8bbb170238e), a "drive on the left" country. Clockwise roundabout is valid. But it's marked as invalid on Osmose.

![image](https://user-images.githubusercontent.com/70379302/188251315-eb916e81-1553-4baa-8a50-7651f25eff1f.png)

frodrigo:
It set in configuration

https://github.com/osm-fr/osmose-backend/blob/master/osmose_config.py#L1146

It is the `driving_side` options at `right` by default.

The configuration for Indonesia is missing (or lost) in the configuration file.

frodrigo:
Done.

PR #1550, opened 2022-09-08
Update left-handed driving countries

Famlam:
Based upon https://en.wikipedia.org/wiki/Left-_and_right-hand_traffic
See issue #1538

Will test tomorrow/this weekend, then release for review/merging.


Famlam:
Works like a charm :)

frodrigo:
Thank you for this and check all other countries.

Famlam:
(Technically I did not check all countries: I only compared the ones that had many issues about inversed roundabouts in Osmose)

Issue #1549, opened 2022-09-08
Missing roundabout check doesn't check the presence of junction=circular


Famlam:
Just spotted this one: https://osmose.openstreetmap.fr/nl/issue/dbe595e7-abb0-95df-4f67-7f3c7e9780ec

Item 2010
Class 1
https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_roundabout.py#L88

The check for potentially missing roundabouts doesn't exclude roads tagged with `junction=circular`

I'll propose a fix tomorrow/this weekend. Probably just adding:
```sql
    NOT (ways.tags?'junction' AND ways.tags->'junction' = 'circular')
```

PR #1551, opened 2022-09-10
Exclude junction=circular in missing roundabout check

Famlam:
Fixes issue #1549.
Also add the `nodes.tags != ''::hstore` check to the new check as recommended by the docs

---------

@frodrigo I have some issues with debugging the 'diff'-supporting analysers locally.
When I run it the first time (i.e. via `./osmose_run.py --no-clean --country=sweden_gotland --skip-frontend-check --skip-upload --analyser=osmosis_roundabout`) all results (1) are present. When I re-run the analyser (using the same command) the results are empty, probably because it uses the diff mode (and the relevant ways haven't changed).
Is there some option to ensure you always get the full results for i.e. debugging purposes?
Example using the command above:
First time running, I got https://www.openstreetmap.org/way/5134430 as a result for class 1
Second time running, I got no results

### DIFF
--- a/analysers/analyser_osmosis_roundabout.py
+++ b/analysers/analyser_osmosis_roundabout.py
@@ -45,6 +45,7 @@ WHERE
     NOT ways.is_construction AND
     (NOT ways.tags?'name' OR ways.tags->'name' LIKE 'Rond%' OR ways.tags->'name' LIKE 'Giratoire%') AND -- no name or start with 'Rond' or 'Giratoire' (French)
     NOT ways.tags->'oneway' = 'no' AND
+    NOT ways.tags->'junction' = 'circular' AND
     -- geometry
     ways.is_polygon AND -- It's a polygon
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank.

If you down not use `--resume-analyser=RESUME_ANALYSER` or `--resume`, it does not run in diff mode.

Famlam:
I must say I am surprised by the amount of results that have disappeared: it went from 16k results to 8k results. I initially got the impression junction=circular was only about 1 out of 10 of the "matches"

Famlam:
In fact, something IS wrong...
`NOT ways.tags->'junction' = 'circular'` (and more surprisingly `NOT (ways.tags->'junction' = 'circular')`) evaluates as false if `junction` is absent.
Working on a fix. Also I realized I don't need to check for `circular` as any other value of junction would not be possible together with `junction=roundabout`

PR #1552, opened 2022-09-10
More left-hand-driving countries

Famlam:
Based upon https://en.wikipedia.org/wiki/Left-_and_right-hand_traffic
(Checked a few more countries that had issues with reversed roundabouts and noticed some were valid, but not set to left-hand driving)


frodrigo:
Thank you.

PR #1548, opened 2022-09-08
Add bad junction=roundabout check

Famlam:
Roundabouts per (OSM) definition must have the right of way. If not, they should be tagged as `junction=circular`. Alternatively, it can also be a roundabout where the (i.e.) traffic signals are located on the connection between the roundabout and the incoming way. In this case, the (i.e.) traffic_signals should be moved to the incoming way only. This check locates roundabouts that contain either `highway=give_way`, `highway=stop` sign, or `highway=traffic_signals`

Results for Luxembourg look good to me:
```xml
<error class="2">
	<way id="74889583" version="32" user="">
		<tag k="lit" v="yes" />
		<tag k="ref" v="N 51" />
		<tag k="foot" v="no" />
		<tag k="name" v="Rond-Point Serra" />
		<tag k="lanes" v="2" />
		<tag k="width" v="10" />
		<tag k="highway" v="primary" />
		<tag k="surface" v="asphalt" />
		<tag k="junction" v="roundabout" />
		<tag k="maxspeed" v="50" />
		<tag k="wikidata" v="Q25583450" />
		<tag k="wikipedia" v="lb:Rond-point Serra" />
		<tag k="cycleway:right" v="no" />
		<tag k="name:etymology:wikidata" v="Q321245" />
	</way>
	<node id="2574889088" lat="49.635280300000005" lon="6.1770361000000005" version="10" user="">
		<tag k="highway" v="traffic_signals" />
		<tag k="traffic_signals:direction" v="forward" />
	</node>
	<location lat="49.635280300000005" lon="6.1770361" />
</error>
<error class="2">
	<way id="87335903" version="11" user="">
		<tag k="lit" v="yes" />
		<tag k="lanes" v="1" />
		<tag k="highway" v="motorway_link" />
		<tag k="surface" v="asphalt" />
		<tag k="junction" v="roundabout" />
		<tag k="maxspeed" v="50" />
		<tag k="motorroad" v="no" />
	</way>
	<node id="3199708780" lat="49.7288678" lon="6.4908228" version="5" user="">
		<tag k="highway" v="stop" />
	</node>
	<location lat="49.7288678" lon="6.4908228" />
</error>
<error class="2">
	<way id="87335903" version="11" user="">
		<tag k="lit" v="yes" />
		<tag k="lanes" v="1" />
		<tag k="highway" v="motorway_link" />
		<tag k="surface" v="asphalt" />
		<tag k="junction" v="roundabout" />
		<tag k="maxspeed" v="50" />
[...]

PR #1554, opened 2022-09-11
Make analyser_osmosis_roundabout_reverse use highways rather than all ways

Famlam:
Actually I just needed a rapid analyser to test some things out with diff-based analyzers locally 🤫, but this commit should also ensure it's no longer matching on `area:highway=*` + `junction=roundabout`, if the `area:highway` was drawn in the "opposite" direction compared to the driving direction. (Example: https://www.openstreetmap.org/way/458628428 ).
Also, if I read the docs correctly, it should be more efficient to use highways rather than ways.

Will keep it as draft as long as I'm playing around with the analyser (on my local computer).
Feel free to reject this PR if you believe it doesn't add much to Osmose

### DIFF
--- a/analysers/analyser_osmosis_roundabout_reverse.py
+++ b/analysers/analyser_osmosis_roundabout_reverse.py
@@ -43,6 +43,8 @@ class Analyser_Osmosis_Roundabout_Reverse(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
+        if not "proj" in self.config.options:
+            return
         self.classs_change[1] = self.def_class(item = 1050, level = 1, tags = ['highway', 'roundabout', 'fix:chair'],
             title = T_('Reverse roundabout'),
             detail = T_(
@@ -68,12 +70,10 @@ Clockwise rotation.'''))
             self.driving_side = ""
 
     def analyser_osmosis_full(self):
-        if "proj" in self.config.options:
-            self.run(sql10.format("", self.driving_side), self.callback10)
+        self.run(sql10.format("", self.driving_side), self.callback10)
 
     def analyser_osmosis_diff(self):
-        if "proj" in self.config.options:
-            self.run(sql10.format("touched_", self.driving_side), self.callback10)
+        self.run(sql10.format("touched_", self.driving_side), self.callback10)
 
 ###########################################################################
 
--- a/analysers/analyser_osmosis_roundabout_reverse.py
+++ b/analysers/analyser_osmosis_roundabout_reverse.py
@@ -68,9 +68,11 @@ Clockwise rotation.'''))
             self.driving_side = ""
 
     def analyser_osmosis_full(self):
+      if "proj" in self.config.options:
         self.run(sql10.format("", self.driving_side), self.callback10)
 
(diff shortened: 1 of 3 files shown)


frodrigo:
Yes. It is a valuable change.

frodrigo:
> test some things out with diff-based analyzers locally

Do not brother you too much with the diff thing. It is a complex subject and hard to test. I mostly implement it by systematic approach rather than testing it.

Famlam:
> Yes. It is a valuable change.

Thank you :)

> Do not brother you too much with the diff thing. It is a complex subject and hard to test. I mostly implement it by systematic approach rather than testing it.

Thanks! Then I'll not focus too much on this (yet)

----------

One question: in order to use `highway`, `proj` must be set. Is it necessary to wrap the tests in `if "proj" in self.config.options:` or is `proj` always present? (And if it's necessary, should I also do this for the two new tests I added in #1528 and #1548?

frodrigo:
Yes, need to introduce `proj`.

> Funny detail, at home the test will now fail with key '@lat' is different: '43.9533100018163' != '43.953310001816334' [.analysers.analyser.error.0.location] 

Not remember seeing this before. Yes may be linked to OS or other.

Famlam:
Thanks! I'll update the other tests as well to check for proj, but I'll do that in a separate PR

analysers/analyser_osmosis_roundabout_reverse.py
```diff
@@ -67,9 +68,11 @@ def __init__(self, config, logger = None):
             self.driving_side = ""
 
     def analyser_osmosis_full(self):
+      if "proj" in self.config.options:
         self.run(sql10.format("", self.driving_side), self.callback10)
```
frodrigo, line 72:
```suggestion
        if "proj" in self.config.options:
            self.run(sql10.format("", self.driving_side), self.callback10)
```

analysers/analyser_osmosis_roundabout_reverse.py
```diff
@@ -67,9 +68,11 @@ def __init__(self, config, logger = None):
             self.driving_side = ""
 
     def analyser_osmosis_full(self):
-        self.run(sql10.format("", self.driving_side), self.callback10)
+        if "proj" in self.config.options:
+            self.run(sql10.format("", self.driving_side), self.callback10)
 
     def analyser_osmosis_diff(self):
+      if "proj" in self.config.options:
         self.run(sql10.format("touched_", self.driving_side), self.callback10)
```
frodrigo, line 76:
```suggestion
        if "proj" in self.config.options:
            self.run(sql10.format("touched_", self.driving_side), self.callback10)
```

analysers/analyser_osmosis_roundabout_reverse.py
```diff
@@ -67,10 +68,12 @@ def __init__(self, config, logger = None):
             self.driving_side = ""
```
frodrigo, line 68: Also need the definition to be put under a if

Famlam: Done

Famlam:
This must've been the PR with the most reverted lines throughout the PR so far :)
If you decide to merge this, please merge as a single commit :)

frodrigo:
Merged, thank.

PR #1555, opened 2022-09-12
Fix analyser_osmosis_roundabout

Famlam:
See #1551 
`NOT ways.tags->'junction' = 'circular'` evaluates to false if junction is not present. However, as no other value of `junction` can work together with `junction=roundabout`, just check that `junction` is absent.

Alternative is:
`NOT (ways.tags?'junction' AND ways.tags->'junction' = 'circular')`

(This very likely means that the line above it also doesn't act as it's intended to, although the effect makes more sense there)

Will run over some more areas overnight to ensure the results are ok now. Apologies... when debugging locally this I was blaming this on my non-understanding of the diff mode. (I should've understood something was wrong when frodrigo told me that diff mode isn't run unless `--resume` is used)

### DIFF
--- a/analysers/analyser_osmosis_roundabout.py
+++ b/analysers/analyser_osmosis_roundabout.py
@@ -45,7 +45,7 @@ WHERE
     NOT ways.is_construction AND
     (NOT ways.tags?'name' OR ways.tags->'name' LIKE 'Rond%' OR ways.tags->'name' LIKE 'Giratoire%') AND -- no name or start with 'Rond' or 'Giratoire' (French)
     NOT ways.tags->'oneway' = 'no' AND
-    NOT ways.tags->'junction' = 'circular' AND
+    NOT ways.tags?'junction' AND
     -- geometry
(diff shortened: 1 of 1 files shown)


frodrigo:
Ok.
Note in SQL `NULL = NULL` is false. Comparing with NULL is always false.

Famlam:
Thanks! That explains a lot!
Knowing this, may I ask what the benefit of `NOT ways.tags->'oneway' = 'no'` (the line above my change) over `ways.tags->'oneway' = 'yes'` (or `ways.is_oneway`) is? Or was this likely an accident?

frodrigo:
> Knowing this, may I ask what the benefit of NOT ways.tags->'oneway' = 'no' (the line above my change) over ways.tags->'oneway' = 'yes' (or ways.is_oneway) is? Or was this likely an accident?

It allows other values than yes (like `-1`)

`highway.is_oneway` was introduced long after this analyser. But yes, can be replaced.

PR #1557, opened 2022-09-13
Check for proj in analysers using highway; add test for analyser_osmosis_roundabout

Famlam:
1. Add a test to ensure the same error as with junction before can't happen again (and yes, I did have some drawing fun while making the test case :) )
2. Protect the class-"2"-test against `proj` being unset. In line with the class-"1"-test, I also put the error description behind `proj`, although that's technically not needed I think
3. Use `is_oneway` rather than `NOT ways.tags->'oneway' = 'no'`

### DIFF
--- a/analysers/analyser_osmosis_cycleway_track.py
+++ b/analysers/analyser_osmosis_cycleway_track.py
@@ -61,6 +61,8 @@ class Analyser_Osmosis_Cycleway_track(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
+        if not "proj" in self.config.options:
+            return
         self.classs[1] = self.def_class(item = 1180, level = 2, tags = ['geom', 'highway', 'cycleway', 'fix:chair'],
             title = T_('Duplicated cycle tracks, `highway=*`+`cycleway=track` and `highway=cycleway`'),
             detail = T_(
--- a/analysers/analyser_osmosis_highway_almost_junction.py
+++ b/analysers/analyser_osmosis_highway_almost_junction.py
@@ -98,6 +98,8 @@ class Analyser_Osmosis_Highway_Almost_Junction(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
+        if not "proj" in self.config.options:
+            return
         self.classs[1] = self.def_class(item = 1270, level = 1, tags = ['highway', 'fix:chair'],
             title = T_('Almost junction, join or use noexit tag'))
 
--- a/analysers/analyser_osmosis_highway_area_access.py
+++ b/analysers/analyser_osmosis_highway_area_access.py
@@ -59,6 +59,8 @@ class Analyser_Osmosis_HighwayAreaAccess(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
+        if not "proj" in self.config.options:
+            return
         self.classs_change[1] = self.def_class(item = 2130, level = 3, tags = ['highway', 'routing'],
             title = T_('Inconsistent Access'),
             detail = T_(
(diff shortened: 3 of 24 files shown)


frodrigo:
> put the error description behind proj, although that's technically not needed I think

It avoid sending empty result to the fronend, as everything is ok. To be checked, but I think if the analyser has no analyse if avoid to try running it.

Famlam:
Interesting. I assume you're referring to the following? 

https://github.com/osm-fr/osmose-backend/blob/fd0e51bdbca93e2365ebafb740a42f91546fd5c7/analysers/Analyser_Osmosis.py#L182-L183

(And similar in `analyser_change`). Which prevent `analyser_osmosis_common`, `analyser_osmosis_full` and `analyser_osmosis_diff` from starting if classes aren't defined.

Wouldn't it be simpler to use:
```py
  Analyser_Osmosis.__init__(self, config, logger)
  if not "proj" in self.config.options:
    return
```
at the start of every analyser that uses `highways`?
And thus avoid indenting and wrapping everything in `if "proj" in self.config.options:` blocks?

Just asking, because I noticed the following analysers all have the issue of not checking (always) if `proj` exists and it'd be fairly easy to just update them all at once:
> FAILED analysers/analyser_osmosis_cycleway_track.py
> FAILED analysers/analyser_osmosis_highway_almost_junction.py
> FAILED analysers/analyser_osmosis_highway_area_access.py
> FAILED analysers/analyser_osmosis_highway_bad_intersection.py
> FAILED analysers/analyser_osmosis_highway_broken_level_continuity.py
> FAILED analysers/analyser_osmosis_highway_cul-de-sac_level.py
> FAILED analysers/analyser_osmosis_highway_deadend.py
> FAILED analysers/analyser_osmosis_highway_floating_islands.py
> FAILED analysers/analyser_osmosis_highway_link.py
> FAILED analysers/analyser_osmosis_highway_motorway.py
> FAILED analysers/analyser_osmosis_highway_noexit.py
> FAILED analysers/analyser_osmosis_highway_traffic_signals.py
> FAILED analysers/analyser_osmosis_highway_turn_lanes.py
> FAILED analysers/analyser_osmosis_highway_without_ref.py
> FAILED analysers/analyser_osmosis_highway_zone.py
> FAILED analysers/analyser_osmosis_parking_highway.py
> FAILED analysers/analyser_osmosis_relation_associatedStreet.py
> FAILED analysers/analyser_osmosis_relation_enforcement.py
> FAILED analysers/analyser_osmosis_relation_public_transport.py
> FAILED analysers/analyser_osmosis_relation_restriction.py
> FAILED analysers/analyser_osmosis_roundabout_level.py


I tried the following:
- In the config, I (temporarily) deleted the `proj` of jersey`
- Ran the analyser osmosis_highway_traffic_signals without modifications = 153 seconds and `psycopg2.errors.UndefinedColumn: column "none" does not exist` (because proj is undefined)
```
<?xml version="1.0" encoding="UTF-8"?>
<analysers>
<analyser timestamp="2022-09-13T14:08:22Z" analyser_version="1522408399" version="(unknown)">
</analysers>
```

- Ran the analyser with the modification of only an early return in `__init__` = 153 seconds (no error)
```
<?xml version="1.0" encoding="UTF-8"?>
<analysers>
</analysers>
```
- For comparison, the analyser with proj set `and` no return is also 153 seconds, but yields the full results file of 32kB with all class definitions and all results

Would you prefer me to update all analysers listed above with

1. Just this:
```
  Analyser_Osmosis.__init__(self, config, logger)
  if not "proj" in self.config.options:
    return
```

or:
2. with
```
  Analyser_Osmosis.__init__(self, config, logger)
  if "proj" in self.config.options:
            self.classs_change[1] = ...

def analyser_osmosis_common(self):
  if "proj" in self.config.options:
    self.run(...)

def analyser_osmosis_full(self):
  if "proj" in self.config.options:
    self.run(...)

  if "proj" in self.config.options:
    self.run(...)
```

frodrigo:
> Wouldn't it be simpler to use:

Yes, that an option, if required to disable everything.

Note, with OSM, I think we always use a projection. Projection less was to experiment on analysing sea (yes) or OpenGeoFiction https://github.com/osm-fr/osmose-backend/blob/master/osmose_config.py#L1796

Famlam:
Will do :)

Famlam:
Ah ok, I'll close this PR and split them into two: one for the changes to analyser_osmosis_roundabout and one for the proj checks elsewhere

Famlam:
Created #1558 for the first part.
I'm not sure how to do a rebase via the web interface, and it'd be a lot of copy-pase effort for little gain (I think) so I'll leave that for now if that's fine? (Given that it went unnoticed for so long that many tests don't check for `proj`, the impact on the server probably isn't too big :) )

PR #1558, opened 2022-09-13
Apply feedback analyser_osmosis_roundabout and add test case

Famlam:
1. Use `is_roundabout` (https://github.com/osm-fr/osmose-backend/pull/1555#issuecomment-1245242048)
2. Add check vs missing `proj` for the class 2 warning too by avoiding analyser initialization (https://github.com/osm-fr/osmose-backend/pull/1554#issuecomment-1244614338)
3. Add test case to ensure the bug I introduced earlier will never occur again undetected ;) #1555


frodrigo:
Thank, merged.

PR #1560, opened 2022-09-14
Require highway_ends in 'common' analysers

Famlam:
- Require highway_ends as common table for common analysers (fixes diff mode run)
- Also ensure `proj` is set

(I suspect it's not needed to remove highway_ends from `requires_tables_full`? I see the code checks if it already exists, and if so, doesn't build it twice, so it'd have little additional benefit)


frodrigo:
Looks good to me. Thank.

PR #1561, opened 2022-09-14
Add requires_tables_common to analyser_osmosis_roundabout

Famlam:
In case it's preferred to have requires_tables_common, even though requires_tables_full and requires_tables_diff both already load `highways` (based on #1560)


frodrigo:
No, as common is "common" for full and diff. If you add to common you have to remove from full and diff.

Issue #1564, opened 2022-09-22
False positive: bad tag value for sport=shot-put


bramoo:
The way 1004803671 is tagged `sport=athletics;rugby_union;hockey;shot-put;long_jump`
Osmose is flagging it as bad tag value, apparently for the dash in shot-put, which [appears to be valid ](https://wiki.openstreetmap.org/wiki/Tag%3Asport%3Dshot-put)

[OSM link](https://www.openstreetmap.org/way/1004803671)
[Osmose details](https://osmose.openstreetmap.fr/en/issue/99110277-b7b5-519a-c5a1-efd8b8e49d62)

frodrigo:
Done

PR #1565, opened 2022-09-22
Accept whitelisted values in ;-separated combinations

Famlam:
Accept whitelisted values also if they occur in a `;`-separated list. (Only for "open" keys; not for "closed" keys with fixed values like `oneway`)

See #1564

Note, the easy way out would be:
`if any(x in tags[k].split(";") for x in self.exceptions_open[k]):`
but that would fail if a whitelisted and 'bad' entry are combined. Given that this part of the code is only ran in rare cases, I guess it doesn't harm performance to have the check as detailed as possible.

### DIFF
--- a/plugins/TagFix_BadValue.py
+++ b/plugins/TagFix_BadValue.py
@@ -111,10 +111,10 @@ However, this should probably still conform to the typical format used for value
         keys = set(keyss) & check_list_open
         for k in keys:
             if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
-                   not k in self.exceptions_open or # no exceptions exist for the key
-                   any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
+                not k in self.exceptions_open or # no exceptions exist for the key
+                any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
             )):
-               err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
+                err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
 
         keys = set(keyss) & self.check_list_closed
         for k in keys:
--- a/plugins/TagFix_BadValue.py
+++ b/plugins/TagFix_BadValue.py
@@ -111,8 +111,9 @@ However, this should probably still conform to the typical format used for value
         keys = set(keyss) & check_list_open
         for k in keys:
             if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
-               not k in self.exceptions_open or # no exceptions exist for the key
-               any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) )): # Check each value in a multiple-value key for not being whitelisted (or normal)
+                   not k in self.exceptions_open or # no exceptions exist for the key
+                   any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
+            )):
                 err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
 
         keys = set(keyss) & self.check_list_closed
--- a/plugins/TagFix_BadValue.py
+++ b/plugins/TagFix_BadValue.py
@@ -112,8 +112,23 @@ However, this should probably still conform to the typical format used for value
         for k in keys:
             if not self.Values_open.match(tags[k]):
                 if k in self.exceptions_open:
-                    if tags[k] in self.exceptions_open[k]:
-                        # no error if in exception list
+                    isWhitelisted = tags[k] in self.exceptions_open[k]
+                    if not isWhitelisted and ";" in tags[k]:
+                        for val in tags[k].split(";"):
+                            if val in self.exceptions_open[k]:
+                                # Whitelisted value
+                                isWhitelisted = True
+                                continue
+                            elif self.Values_open.match(val):
+                                # Normal value
+                                continue
+                            else:
+                                # Bad value
+                                isWhitelisted = False
+                                break
+
+                    if isWhitelisted:
+                        # No error if in exception list
                         continue
                 err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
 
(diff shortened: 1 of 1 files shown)


plugins/TagFix_BadValue.py
```diff
@@ -112,8 +112,23 @@ def check(self, data, tags, check_list_open):
...
+                                continue
+                            else:
+                                # Bad value
+                                isWhitelisted = False
+                                break
+
+                    if isWhitelisted:
+                        # No error if in exception list
                         continue
                 err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```
frodrigo, line 133: Try to simplify, not tested code. We probably can simply more, as sub-if are just `and`.
```suggestion
                    if tags[k] not in self.exceptions_open[k] and ";" in tags[k]:
                        if any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";")):
                            err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```

Famlam: I could not directly use this code as it would not trigger an error in case there's no `;` in the code, but I've added a commit that serves a similar shortening of the code

plugins/TagFix_BadValue.py
```diff
@@ -111,11 +111,10 @@ def check(self, data, tags, check_list_open):
...
             if not self.Values_open.match(tags[k]):
-                if k in self.exceptions_open:
-                    if tags[k] in self.exceptions_open[k]:
-                        # no error if in exception list
-                        continue
-                err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
+                if (not k in self.exceptions_open or # no whitelists exist for the key
+                   (not ";" in tags[k] and not tags[k] in self.exceptions_open[k]) or # single-valued tag and not whitelisted
+                   (";" in tags[k] and any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))))): # multi-valued tag and one or more bad values
+                    err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```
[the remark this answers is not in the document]
Famlam: My reasoning against that was that it's probably much more efficient to use a `a in b` (as I guess the majority of the cases is without `;`) rather than the full call with map, split, match, but I can change it, no problem!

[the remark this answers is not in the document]
Famlam: Done, also the requested combination with the previous 'if'. Fortunately there's no limit on the number of parentheses :)

plugins/TagFix_BadValue.py
```diff
@@ -110,11 +110,9 @@ def check(self, data, tags, check_list_open):
...
         keys = set(keyss) & check_list_open
         for k in keys:
-            if not self.Values_open.match(tags[k]):
-                if k in self.exceptions_open:
-                    if tags[k] in self.exceptions_open[k]:
-                        # no error if in exception list
-                        continue
+            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
+               not k in self.exceptions_open or # no exceptions exist for the key
+               any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) )): # Check each value in a multiple-value key for not being whitelisted (or normal)
```
frodrigo, line 115:
```suggestion
            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
                   not k in self.exceptions_open or # no exceptions exist for the key
                   any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
            )):
```

plugins/TagFix_BadValue.py
```diff
@@ -110,11 +110,10 @@ def check(self, data, tags, check_list_open):
...
-            if not self.Values_open.match(tags[k]):
-                if k in self.exceptions_open:
-                    if tags[k] in self.exceptions_open[k]:
-                        # no error if in exception list
-                        continue
+            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
+                   not k in self.exceptions_open or # no exceptions exist for the key
+                   any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
+            )):
                 err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```
frodrigo, line 117:
```suggestion
               err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```

plugins/TagFix_BadValue.py
```diff
@@ -110,12 +110,11 @@ def check(self, data, tags, check_list_open):
...
-                if k in self.exceptions_open:
-                    if tags[k] in self.exceptions_open[k]:
-                        # no error if in exception list
-                        continue
-                err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
+            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
+                   not k in self.exceptions_open or # no exceptions exist for the key
+                   any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
+            )):
+               err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```
frodrigo, line 117:
```suggestion
            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
                not k in self.exceptions_open or # no exceptions exist for the key
                any(map(lambda val: val not in self.exceptions_open[k] and not self.Values_open.match(val), tags[k].split(";"))) # Check each value in a multiple-value key for not being whitelisted (or normal)
            )):
               err.append({"class": 3040, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
```

frodrigo:
Thank you.

Famlam:
Regarding the change in indent after a newline (e.g.: https://github.com/osm-fr/osmose-backend/pull/1565/commits/9e5f06d9b2e6d669c8f0c8c26f33ebed0d567620 ) perhaps lint has a rule for this?

Famlam:
My bad, I was not clear. I meant, maybe Lint has a rule for:
```py
            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
               not k in self.exceptions_open or # no exceptions exist for the key
```
(i.e. my code) instead of: (your code)
```py
            if (not self.Values_open.match(tags[k]) and ( # value has a non-standard character
                not k in self.exceptions_open or # no exceptions exist for the key
```
I.e. that it automatically tells me that I should add a space in the second line.
Both these fragments of code were fine according to the present lint settings.
Anyway, not the most important thing :)

frodrigo:
Deployed.

PR #1584, opened 2022-09-26
Resolve duplicate warning: transformer on way

Famlam:
https://osmose.openstreetmap.fr/nl/issue/7524de41-d16f-1f94-bd18-d83c8a6790c1 https://osmose.openstreetmap.fr/nl/issue/86f129a2-fe93-9263-080f-9ae6a895712f

First one is from power.validator.mapcss (which this PR removes). Second one is from [JOSM](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/geometry.mapcss#L102). The rules are practically identical

### DIFF
--- a/plugins/Power.py
+++ b/plugins/Power.py
@@ -55,27 +55,6 @@ class Power(PluginMapCSS):
 
         return err
 
-    def way(self, data, tags, nds):
-        capture_tags = {}
-        keys = tags.keys()
-        err = []
-
-
-        # way[power=transformer]
-        if ('power' in keys):
-            match = False
-            if not match:
-                capture_tags = {}
-                try: match = ((mapcss._tag_capture(capture_tags, 0, tags, 'power') == mapcss._value_capture(capture_tags, 0, 'transformer')))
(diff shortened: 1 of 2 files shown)


frodrigo:
Not sure to understand. The python file is not in synch with the mapcss ?

Famlam:
The same rule exists in both JOSM ( https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/geometry.mapcss#L102 , which is imported into Osmose) and in Osmose's own mapcss file ([power.validator.mapcss](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Power.validator.mapcss))

Hence, there were two mapcss files with exactly the same rule.


This PR only removes the rule from one of the two mapcss files to avoid two warnings about the same. There is no syncing issue :)

frodrigo:
Ok, I see.
Please, can add a comment into to power mapcss file, to avoid re-adding it.

Famlam:
Sure!

frodrigo:
Deployed.

PR #1585, opened 2022-10-01
Update Colour.validator.mapcss

### LINKED ISSUE #1567: Extend check of coulor
I would like to propose to extend the check for colors to not only test for hexadecimal values, but also check if real color names are used.
For example: https://www.openstreetmap.org/node/8131853250 uses red-white rather than red;white. And taginfo shows a large amount of 'interesting' key values.

This can be done in two ways: 
- By updating the mapcss plugin
- Via a new python plugin

**Advantages of mapcss:**
 - Easy to add to the current file [colour.validator.mapcss](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Colour.validator.mapcss) as something like:
`[colour][colour!~/^[a-z]+(; ?[a-z]+)*/][colour!~/^#/]`
- Catches a significant amount of the errors, especially if we add some more keys to the check

**Advantages of python:**
- Not limited to specific keys, we can just loop over all keys and find all that start or end with `colour`
- We can actually check for existing values, rather than just pattern recognition (e.g. if someone uses `rood` (Dutch) rather than `red`, it'd go undetected in pattern detection) <details><summary>color list</summary>```aliceblue, antiquewhite, aqua, aquamarine, azure, beige, bisque, black, blanchedalmond, blue, blueviolet, brown, burlywood, cadetblue, chartreuse, chocolate, coral, cornflowerblue, cornsilk, crimson, cyan, darkblue, darkcyan, darkgoldenrod, darkgray, darkgreen, darkgrey, darkkhaki, darkmagenta, darkolivegreen, darkorange, darkorchid, darkred, darksalmon, darkseagreen, darkslateblue, darkslategray, darkslategrey, darkturquoise, darkviolet, deeppink, deepskyblue, dimgray, dimgrey, dodgerblue, firebrick, floralwhite, forestgreen, fuchsia, gainsboro, ghostwhite, gold, goldenrod, gray, grey, green, greenyellow, honeydew, hotpink, indianred, indigo, ivory, khaki, lavender, lavenderblush, lawngreen, lemonchiffon, lightblue, lightcoral, lightcyan, lightgoldenrodyellow, lightgray, lightgreen, lightgrey, lightpink, lightsalmon, lightseagreen, lightskyblue, lightslategray, lightslategrey, lightsteelblue, lightyellow, lime, limegreen, linen, magenta, maroon, mediumaquamarine, mediumblue, mediumorchid, mediumpurple, mediumseagreen, mediumslateblue, mediumspringgreen, mediumturquoise, mediumvioletred, midnightblue, mintcream, mistyrose, moccasin, navajowhite, navy, oldlace, olive, olivedrab, orange, orangered, orchid, palegoldenrod, palegreen, paleturquoise, palevioletred, papayawhip, peachpuff, peru, pink, plum, powderblue, purple, red, rosybrown, royalblue, saddlebrown, salmon, sandybrown, seagreen, seashell, sienna, silver, skyblue, slateblue, slategray, slategrey, snow, springgreen, steelblue, tan, teal, thistle, tomato, turquoise, violet, wheat, white, whitesmoke, yellow, yellowgreen```</details>

To be able to avoid false positives (e.g. the python plugin shouldn't flag on items that are flagged by the mapcss test), it would be most convenient to deprecate the mapcss if the python plugin-method is chosen and have both the new and the existing checks implemented in the python plugin

Any preference? (If you like the idea of course)

Famlam:
Fixes #1567

- Implements a check for named colors
- Adds keys with > 50k uses `colour:back`, `colour:text`, `light:colour`, `colour:arrow`, `seamark:buoy_lateral:colour`, `seamark:light:colour`, `seamark:topmark:colour`
- Remove (rare, but incorrect) warning for `;`-separated colours in #code-colours and named colours and mixtures thereof. (In case of missing `#` (the middle rule) combined with a `;`-separation, I just let it get through to the third check: unknown color name)
- Switch `[{key}=~/^#/]` (which otherwise would've been changed to `[{key}=~/#/]`) to `[{key}*="#"]` as string operations are usually faster


frodrigo:
Ok, looks good to me.

Can you also generate the python code, and commit it, plase.

python -m mapcss.mapcss2osmose plugins/Colour.validator.mapcss

An then, you can also run the unitary tests.

Famlam:
I'll upload the generated mapcss. However, I somehow can't run the pytests for generated mapcss files so I'll let it run on github instead.

On my personal PC I always get (for any mapcss) this result from pytest:
```
self = <plugins.Colour.Colour object at 0x7f1669a82220>, logger = None

    def init(self, logger):
        super().init(logger)
        tags = capture_tags = {} # noqa
>       self.errors[30911] = self.def_class(item = 3091, level = 2, tags = mapcss.list_('tag'), title = mapcss.tr('Colour code should start with \'#\' followed by 3 or 6 digits'))
E       AttributeError: module 'modules.mapcss_lib' has no attribute 'list_'

plugins/Colour.py:16: AttributeError
```

On the dev server I can't compile (any file, here I used Phone2 for demonstration):
```
famlam@osm110:/home/famlam/osmose1329$ python -m mapcss.mapcss2osmose plugins/Phone2.validator.mapcss
  File "/usr/lib/python2.7/runpy.py", line 163, in _run_module_as_main
    mod_name, _Error)
  File "/usr/lib/python2.7/runpy.py", line 119, in _get_module_details
    code = loader.get_code(mod_name)
  File "/usr/lib/python2.7/pkgutil.py", line 281, in get_code
    self.code = compile(source, self.filename, 'exec')
  File "/home/famlam/osmose1329/mapcss/mapcss2osmose.py", line 249
    t['declarations'] = sorted(t['declarations'], key = lambda d: (d.get('property') and [rule_declarations_order_map.get(d['property']) or print("W: Unknow property: " + d['property']) and -1, str(d['value'])]) or [-1, -1])
                                                                                                                                                ^
SyntaxError: invalid syntax
```

frodrigo:
> /usr/lib/python2.7/runpy.py

You need to use pyhton 3.7 at least.

You can do this with virtualenv
https://github.com/osm-fr/osmose-backend/blob/master/INSTALL.md

Famlam:
Ah true, somehow I just assumed the server would be running python 3. Anyway, given that I don't expect to use it often, I guess I'll be a bit lazy and just let Github run it for me. I had extensively tested it in JOSM already.

frodrigo:
Note, JOSM can also run mapcss test directly. No remember how, but it is in the doc.

Famlam:
> Note, JOSM can also run mapcss test directly. No remember how, but it is in the doc.

Correct, you have to set `validator.check_assert_local_rules` to `true` in advanced preferences, then you can see any errors in i.e. the status report. 
I use this all the time for my private rules as well as for [the Dutch validation rules that I maintain](https://github.com/Famlam/OsmMapcssValidationNL/blob/main/netherlands.validator.mapcss).

frodrigo:
Deployed.

Issue #1570, opened 2022-09-26
False-positive for name='Rue de la Salle des Fêtes'


iiitomo:
Currently, Osmose suggests to add 'amenity=community_centre' when name='Rue de la Salle des Fêtes' or 'Parc de la Salle des fêtes' or similar.
See attached screenshot.
It should do that suggestion only when name='Salle des Fêtes', or name contains 'Salle des Fêtes d' (e.g. 'Salle des fêtes du village').

![osmose_bad_suggestion_for_rue_salle_des_fetes](https://user-images.githubusercontent.com/58949423/192227298-25bc2f5c-2f04-4d58-ad7d-87e4affe8d68.PNG)
![osmose_error_suggestion_for_salle_des_fetes](https://user-images.githubusercontent.com/58949423/192329765-b709007b-f6d0-4ac8-b975-1f49d7056f4d.PNG)

frodrigo:
Done.

PR #1587, opened 2022-10-02
fix issue 1570

didier2020:
#1570
no word before "Salle des fete" in the regex


frodrigo:
No. That not a fix.
We must add more guard condition L64.

Issue #1590, opened 2022-10-03
Mapcss generation fails with [key?!] selector


Famlam:
I encountered that on the following rule:
```
way[oneway:bicycle?!][!oneway:mofa][oneway?][inside("NL")] {
  throwWarning: tr("{0} without {1}", "{0.tag}", "{1.key}");
  group: tr("NL overige problemen");
}
```
the mapcss generator fails (in a cascade of errors, but primarily due to 
`line XXX:19 no viable alternative at input 'way[oneway:bicycle?!'`
), after which everything fails.

As far as I could debug it, it seems related to `?!` not being defined as an existing selector in: 
https://github.com/osm-fr/osmose-backend/blob/ec6bc95e5091f2c77ba615e33abf9a5374fbc26f/mapcss/MapCSS.g4#L203-L206

I suspect that 
```
OP_NOT ? (osmtag | quoted) QUESTION_MARK ?
```
should be replaced by
```
OP_NOT ? (osmtag | quoted) ? |
(osmtag | quoted) QUESTION_MARK OP_NOT ?
```

frodrigo:
I think you are right.

But unlike the other part, this new operator is not well documented.

I think it rather should be `(QUESTION_MARK OP_NOT ?) ?` or by adding a new operator `?!` as (eg) `QUESTION_MARK_NEGATED` (or BANG).

Famlam:
> But unlike the other part, this new operator is not well documented.

The operator seems to be [defined](https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation) as meaning `not [key?]`:

> The value is evaluated to true, if it is either "yes", "true", or "1". All other values are evaluated to false.

Which probably means that nothing else needs to be changed on the `mapcss2osmose` side (I hope)

---------------------

Regarding: 

> I think it rather should be (QUESTION_MARK OP_NOT ?) ?

I would expect they produce the same result: my solution would exclude the possibility of `[!key?!]` (which I have no clue how that's supposed to evaluate, but in practice that'll probably never occur), but otherwise the result should be the same.

---------------------

My local system cannot run anything beyond ANTLR 4.9.3 (I probably have to switch to openjava at some point: still running the (up to date) 'regular' java). Am I allowed (and able) to install antlr on the dev server? Or would you be able to regenerate the parser for me with this change so I can test it?

PR #1591, opened 2022-10-05
Support [key?!] selectors

Famlam:
See #1590 

First commit:
- Only updates the mapcss so I could run the script and only see the differences due to the new parsing. No code changes

Second commit:
- Fixes `[key?!]` selectors, see example below. Didn't affect any of the current JOSM rules

Third commit:
- Fixes a missing space I spotted in the `set` comments. Affects many (comments of) JOSM rules

<details>
<summary>Example for commit 2</summary>
The example rule (from my Dutch filters; currently not in Osmose)

```
way[oneway:bicycle?!][!oneway:mofa][oneway?][inside("NL")] {
  throwWarning: tr("{0} without {1}", "{0.tag}", "{1.key}");
  group: tr("NL overige problemen");
}
```

became 
```
if ('oneway' in keys and 'oneway:bicycle' in keys):
            match = False
            if not match:
                capture_tags = {}
                try: match = ((mapcss._tag_capture(capture_tags, 0, tags, 'oneway:bicycle') not in ('yes', 'true', '1')) and (not mapcss._tag_capture(capture_tags, 1, tags, 'oneway:mofa')) and (mapcss._tag_capture(capture_tags, 2, tags, 'oneway') in ('yes', 'true', '1')) and (mapcss.inside(self.father.config.options, 'NL')))
                except mapcss.RuleAbort: pass
            if match:
                # group:tr("NL overige problemen")
                # throwWarning:tr("{0} without {1}","{0.tag}","{1.key}")
                err.append({'class': 8, 'subclass': 2063031692, 'text': mapcss.tr('{0} without {1}', mapcss._tag_uncapture(capture_tags, '{0.tag}'), mapcss._tag_uncapture(capture_tags, '{1.key}'))})
```


mapcss/MapCSS.g4
```diff
@@ -77,7 +78,7 @@ OP_PLUS: '+';
 OP_MINUS: '-';
 OP_NOT: '!'; // NOTE: boolean not -> !(expr)
 
-SET: ('s' | 'S') ('e' | 'E') ('t' | 'T');
+SET: ('s' | 'S') ('e' | 'E') ('t' | 'T') (' ' | '\t') ?;
```
frodrigo, line 81: No, parsing space is not the right option.
It probably an issue with some missing space in the output printing.

Famlam: `set` itself is never defined to contain a space. 
Already in the first (not-generated-by-ANTLR) file, in the function `exitDeclaration`, it looks like this without the patch:
`{'type': 'declaration', 'text': 'setarea_yes_autofix', 'set': 'area_yes_autofix', 'property': None, 'value': None}`
while, with the patch, it looks like this:
`{'type': 'declaration', 'text': 'set area_yes_autofix', 'set': 'area_yes_autofix', 'property': None, 'value': None}`

The `text` property is what's printed out in `to_p` in mapcss2osmose:
`selectors_text = "# " + "\n# ".join(map(lambda s: s['text'], t['selectors']))`
Adding exceptions for `set` there is probably going to be very ugly.

Spaces are just discarded everywhere:
`WS:           (' ' | '\t' | '\n' | '\r' | '\f') -> channel(HIDDEN);`
(That's also why for example [this](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Colour.validator.mapcss#L70) many-space-containing mapcss rule is [changed to](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Colour.py#L99) `# *[colour:arrow][colour:arrow=~/^([0-9A-Fa-f]{3}|[0-9A-Fa-f]{6})$/]`, without spaces: spaces are dropped unless explicitly captured)

So either we have to explicitly capture it (like in the quoted strings (used in i.e. `[a="b c"]`) where spaces must also be retained, e.g.:
`DQUOTED_STRING: '"' (' ' | '!' | '#'..'[' | ']'..'~' | '°' | UNICODE | EDQUOTE | EBACKSLASH)* '"';`
or we would have to change exitDeclaration to get a special case for set, for example:
```
    def exitDeclaration(self, ctx:MapCSSParser.DeclarationContext):
        v = self.stack.pop()
        self.declarations.append({
            'type': 'declaration',
            'text': ctx.getText(),
            'set': ctx.cssident() and ctx.cssident().getText(),
            'property': ctx.declaration_property() and ctx.declaration_property().getText(),
            'value': v['single_value'] or v['functionExpression']
        })
        if self.declarations[-1]["set"]:
            self.declarations[-1]["text"] = self.declarations[-1]["text"].replace("set", "set ", 1)
```

I would argue it's neater to capture it than to have such special cases. Would you prefer the latter?

frodrigo: There also the same issue with `import` (but not used/implemented).
I look at how to implement it. I think it may be something like this (not tested)
```diff
         self.declarations.append({                                                                                       
             'type': 'declaration',                                                                                       
-             'text': ctx.getText(),                                                                                       
+             'text': ctx.getText() + (" " if self.declarations[-1]["set"] else ""),                                                                                       
             'set': ctx.cssident() and ctx.cssident().getText(),                                                          
             'property': ctx.declaration_property() and ctx.declaration_property().getText(),                             
             'value': v['single_value'] or v['functionExpression']                                                        
         })  
```

Famlam: > I look at how to implement it.

Thank you!

Be aware that `self.declarations[-1]["set"]` is only defined the line after you use it in your example code, so I guess you'd need to create a variable like `set_value`, as below (also untested)
```
         set_value = ctx.cssident() and ctx.cssident().getText()
         self.declarations.append({                                                                                       
             'type': 'declaration',                                                                                       
             'text': ctx.getText() + (" " if set_value else ""),                                                                                       
             'set': set_value,                                                          
             'property': ctx.declaration_property() and ctx.declaration_property().getText(),                             
             'value': v['single_value'] or v['functionExpression']                                                        
         })  
```

Famlam: > I look at how to implement it.



In case you need inspiration, you could consider 
`'text': re.sub(r'^[sS][eE][tT]', 'set ', ctx.getText()),`

Note that regular `.replace('set', 'set ', 1)` wouldn't work because:
1. the definition of `SET` also matches SeT, sET, ..., 
2. `declaration` (and thus the text property) could, besides set declarations, also be i.e. `declaration_property COLON declaration_value` where `declaration_value` may contain `set` in a word, i.e. 
```
way[surface=cobblestone] {
  throwWarning: 'Prefer sett or unhewn_cobblestone over cobblestone';
}
```
would result in `ctx.getText()` outputting `throwWarning:'Prefer sett or unhewn_cobblestone over cobblestone'` (so including the word `set` in `sett`, where you shouldn't add a space)

frodrigo:
Thank. Merged manually.

Famlam:
@frodrigo 
I think you didn't merge the part to support `[key?!]` right?

frodrigo:
> @frodrigo I think you didn't merge the part to support `[key?!]` right?

I intended. What is missing ?

Famlam:
correction, you did https://github.com/frodrigo/osmose-backend/commit/a68b74d3abafb11d59bfa6587cf27c8d3ed692d1 (but somehow it doesn't show up in this PR, only the other commit https://github.com/frodrigo/osmose-backend/commit/880002c1e4760ae0a5f25d126cea5d751dc75b48). Apologies!

PR #1592, opened 2022-10-05
Update mapcss README.md

Famlam:
- Add download link for ANTLR
- Update command so it also works when ran in the Windows terminal (if not, I got all kind of unknown char messages as it doesn't default to utf8)
**Please do check that this doesn't affect anything poorly in Linux**

### DIFF
--- a/mapcss/MapCSSListenerL.py
+++ b/mapcss/MapCSSListenerL.py
@@ -1,3 +1,5 @@
+#-*- coding: utf-8 -*-
+
 from .generated.MapCSSListener import MapCSSListener
 from .generated.MapCSSParser import MapCSSParser
 from typing import Dict, List, Optional
--- a/mapcss/README.md
+++ b/mapcss/README.md
@@ -20,5 +20,6 @@ Update all external MapCSS
 
 From root directory
 ```
-python -X utf8 -m mapcss.update
+python -m mapcss.update
 ```
+(Windows users should set the system variable `PYTHONUTF8` to `1` first)
--- a/mapcss/README.md
+++ b/mapcss/README.md
@@ -6,6 +6,7 @@ Generate the parser
 ```
 java -jar antlr-4.10.1-complete.jar -o generated/ -Dlanguage=Python3 MapCSS.g4
 ```
+ANTLR can be downloaded [here](https://github.com/antlr/antlr4/releases)
 
 Convert one MapCSS
 ==================
@@ -19,5 +20,5 @@ Update all external MapCSS
 
 From root directory
 ```
-python -m mapcss.update
+python -X utf8 -m mapcss.update
 ```
--- a/mapcss/mapcss2mapcss.py
+++ b/mapcss/mapcss2mapcss.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+#-*- coding: utf-8 -*-
 
 import sys
 from antlr4 import FileStream, CommonTokenStream, ParseTreeWalker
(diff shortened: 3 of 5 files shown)


frodrigo:
Maybe just add this at the start of faulting file
```
#-*- coding: utf-8 -*-
```

frodrigo:
That nice to see some one deeping into that part of Osmose. Probably the most complicated one!

Famlam:
> Maybe just add this at the start of faulting file
> 
> ```
> #-*- coding: utf-8 -*-
> ```

I just tried this, adding it to all Osmose files involved (leaving the python files that handle the read/write etc untouched), unfortunately doesn't work. I really need to specify the -X option.
```
C:\[...]\osmose-backend-master>python -m mapcss.update
https://josm.openstreetmap.de/josmfile?page=Rules/FranceSpecificRules
  File "C:\Program Files\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Program Files\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\[...]\osmose-backend-master\mapcss\update.py", line 31, in <module>
    main(sys.argv[1:])
  File "C:\[...]\osmose-backend-master\mapcss\update.py", line 22, in main
    open(f, 'w').write(r.text)
  File "C:\Program Files\Python310\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2289' in position 13245: character maps to <undefined>
```


> That nice to see some one deeping into that part of Osmose. Probably the most complicated one!

Yeah, very complicated README, was a huge challenge to get it pass all tests :)

frodrigo:
Not sure about adding the `-X` option. Does we support windows ? I even surprised you are able to run some (all ?) part of it under windows. @jocelynj what do you think about this ?

Famlam:
>I even surprised you are able to run some (all ?) part of it under windows

Yep, I was surprised too :)

jocelynj:
On windows, is it possible to set LANG to something like en_US@UTF-8 to remove the need for "-X utf8"? Or setting variable PYTHONUTF8?

If not, and if this option doesn't break Linux, then we should add this option in the README, maybe with a comment to tell that it is for windows.

On the other side, we should add the missing `#-*- coding: utf-8 -*-` in all python files, just to be cleaner.

Famlam:
> Or setting variable PYTHONUTF8?

I didn't know about this one yet, but this works perfectly!

frodrigo:
> > Or setting variable PYTHONUTF8?
> 
> I didn't know about this one yet, but this works perfectly! I adjusted the PR, and incorporated the suggestion of @jocelynj. If you prefer not to mention Windows-specific things, that's fine for me too, then I'll remove that line

It is not mapcss specific and could be set in an other READEM/INTALL/doc part I think.

The `#-*- coding: utf-8 -*-` you added still necessary ? (I removed this such line in the frontend)

Famlam:
>The #-*- coding: utf-8 -*- you added still necessary ? (I removed this such line in the frontend)

As far as I'm aware, it's only mandatory if you use unicode characters in the .py files directly. It however doesn't hurt either. So I'll let you and Jocelynj decide and adjust the code accordingly ;)

> It is not mapcss specific and could be set in an other READEM/INTALL/doc part I think.

Suggestions? (I haven't encountered issues with this anywhere else yet, but of course most files don't have read/write operations on external files containing i.e. ⊇ and such, except mapcss ;) )

Issue #1533, opened 2022-08-30
The tag area=yes is unnecessary is flagged twice for 1 'error' (since longer)


sekerob:
When mapping areas of landuse, the removal of the standard area=yes tag is occasionally forgotten as not required. It gets flagged with "landuse`, already make this an area". The same though gets also flagged with "highway={{2.value}} is unnecessary for area", which I think is incorrect.

Below a partial copy of the daily extract list of personal Osmose issues showing both errors refer to the same way link: 

<html>
<body>

2416 | 3 | 3200 area | 32001 | E | 14.09 42.33 | w1090301726  (j) | Tags, `landuse`, already make this an area. | ✘ / ✔
-- | -- | -- | -- | -- | -- | -- | -- | --
2416 | 3 | 3200 area | 32001 | E | 14.09 42.33 | w1090301727  (j) | Tags, `landuse`, already make this an area. | ✘ / ✔
2416 | 3 | 3200 area | 32001 | E | 14.09 42.33 | w1090301728  (j) | Tags, `landuse`, already make this an area. | ✘ / ✔
2416 | 3 | 9010 unnecessary | 9010001 | E | 14.09 42.33 | w1090301726  (j) | highway={{2.value}} is unnecessary for area | ✘ / ✔
2416 | 3 | 9010 unnecessary | 9010001 | E | 14.09 42.33 | w1090301727  (j) | highway={{2.value}} is unnecessary for area | ✘ / ✔
2416 | 3 | 9010 unnecessary | 9010001 | E | 14.09 42.33 | w1090301728  (j) | highway={{2.value}} is unnecessary for area

</body>
[...]

Issue #1304, opened 2021-07-21
wrong interpretation in class 9010001


Klumbumbus:
all details see https://josm.openstreetmap.de/ticket/21149

frodrigo:
http://osmose.openstreetmap.fr/en/issue/f7b50ad2-69fd-69c9-a91c-df137f0ff980

|k|v
-- | --
area | yes
barrier | fence
leisure | dog_park
opening_hours | Mo-Su 00:00-24:00; 09:00-17:00; PH closed
wheelchair | no


unnecessary tag
highway={{2.value}} is unnecessary for area

PR #1594, opened 2022-10-09
Record proper selector indices

Famlam:
Per the wiki on tag capture: (https://josm.openstreetmap.de/wiki/Help/Validator/MapCSSTagChecker) "Classes and pseudoclasses do also count."

Rather than counting in mapcss2osmose (which is fairly difficult as some tags get 'cleaned' away (ex `:completely_downloaded`) and the sequence of the tags gets shuffled around for optimization, count it where it's freshly parsed and store the index as a property of the captured selector.

See #1535, #1533, #1304

### DIFF
--- a/mapcss/MapCSSListenerL.py
+++ b/mapcss/MapCSSListenerL.py
@@ -65,6 +65,7 @@ class MapCSSListenerL(MapCSSListener):
         self.predicates: List[Dict] = []
         self.predicates_function_base: Optional[List[Dict]] = None
         self.pseudo_class = [] # : List[Dict]
+        self.selector_index = 0
 
     # Exit a parse tree produced by MapCSSParser#simple_selector.
     def exitSimple_selector(self, ctx:MapCSSParser.Simple_selectorContext):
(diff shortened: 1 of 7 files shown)


mapcss/MapCSSListenerL.py
```diff
@@ -184,8 +195,10 @@ def exitBooleanExpression(self, ctx:MapCSSParser.BooleanExpressionContext):
             'operands':
                 v['booleanExpressions'] or # Juste get operands array
                 (v['functionExpression'] and [v['functionExpression']]) or
-                []
+                [],
+            'selector_index': self.selector_index
```
frodrigo, line 199:
```suggestion
                'selector_index': self.selector_index
```

frodrigo:
Look good to me.
Can you just review my change suggestion?

mapcss/MapCSSListenerL.py
Famlam: Are you sure about this? It should have the same indentation as the other keys ('type', 'operator', 'operands') of the dictionary right?

frodrigo: Right!

frodrigo:
Thank you for looking at this complicated thing.

Famlam:
> Thank you for looking at this complicated thing.

I enjoyed it ;)

PR #1597, opened 2022-10-12
Improve error message

### LINKED ISSUE #1567: Extend check of coulor

Famlam:
Closes #1567

Only relevant change is in plugins/Colour.validator.mapcss; the other changes are just mapcss regeneration (which we forgot for the local files in /plugins/ when we edited the mapcss parser)
Changes the title from "Unknown color name" to "Bad colour name" and the throwWarning from "Unknown colour name in tag 'X'" to "Unknown or invalid colour name in tag 'X'".

(Also regenerated the python code of indoor.validator.mapcss and Bicycle.validator.mapcss; doesn't effectively change anything as these do not use `{n}` capture groups in messages)


frodrigo:
Ok. Thank.
Good catch for missing regeneration.

Note, preferably make this kind of changes in two commits.

Issue #1603, opened 2022-10-15
Mapcss key not parsed


Famlam:
This rule from unnecessary.mapcss (JOSM)
```css
/* #14256, #15664, #20681 */
way:closed[highway=rest_area][area?],
way:closed[highway=services][area?],
way:closed[aeroway=aerodrome][area?],
way:closed[waterway=riverbank][area?],
way:closed[aeroway=helipad][area?] {
  throwWarning: tr("{0} is unnecessary for {1}", "{2.tag}", "{1.tag}");
  group: tr("unnecessary tag");
  fixRemove: "{2.key}";
}
```
Applied on way https://www.openstreetmap.org/way/926894128 (or https://www.openstreetmap.org/way/98570914 )
```
area	yes 
highway	services 
name	Paterson
```

https://osmose.openstreetmap.fr/nl/issue/fb7bf10b-4376-9a97-5120-4c0a19bec84c
https://osmose.openstreetmap.fr/en/issue/8a03e8a6-2c8e-bc26-3576-e112440580a0
Message: `area=yes is unnecessary for {{1.key}}=services`

`{{1.key}}` should be parsed as `highway`

I don't know if this is a regression of the recent mapcss changes or a preexisting issue

Famlam:
Patch ready :)

Famlam:
I see it's deployed

PR #1605, opened 2022-10-16
Deal with non-sequential indices

Famlam:
Issue #1603 

`mapcss._tag_capture(...)` captured `[key, value]` correctly, but `mapcss._value_capture(...)` then overwrote the capture with `[None, None]` (and only set the value itself, resulting in eventually `[None, value]`)

Also added a test to spot such issues in the future: if a `{` or `}` appears in the string, it'll complain.


Famlam:
The effect of this PR on two countries can be seen here: https://github.com/Famlam/osmose-backend/commit/8f35ba67448cd137689fe142bf340aaa60e5f858

frodrigo:
Ok.
Thank you for looking into this.

About the test. What do you thing of using a subclass for test for generated mapcss plugin in place of `disallowed_str_in_text` anywhere?

frodrigo:
Ok.
Thank you for looking into this.

About the test. What do you thing of using a subclass for test for generated mapcss plugin in place of `disallowed_str_in_text` anywhere?

Famlam:
> Ok. Thank you for looking into this.
> 
> About the test. What do you thing of using a subclass for test for generated mapcss plugin in place of `disallowed_str_in_text` anywhere?

I think that's a good idea ;). Initially I thought it may be a useful feature, but I can't really think of other plugins that would benefit. Done!

frodrigo:
> > About the test. What do you thing of using a subclass for test for generated mapcss plugin in place of `disallowed_str_in_text` anywhere?
> 
> I think that's a good idea ;). Initially I thought it may be a useful feature, but I can't really think of other plugins that would benefit. Done!

Thank to change it this way.

PR #1607, opened 2022-10-20
Add major german supermarket brands to UpperCase Validator whitelist

malteger:
As listed at https://nsi.guide/index.html?t=brands&k=shop&v=supermarket&cc=de


frodrigo:
Thank you. Merged, but not deployed, yet.

Not, Osmose already use NSI, this may be done globally and automatically.

malteger:
Oh, I did not see that. Does the NSI check for brands prevent this validator from running, or will the uppercase brands still be marked?

frodrigo:
> Oh, I did not see that. Does the NSI check for brands prevent this validator from running, or will the uppercase brands still be marked?

There is no linked between the two. That what I suggested in my note.

malteger:
So automatically adding relevant brands from NSI to the whitelist. I will have a look and see if I can implement it.

PR #1613, opened 2022-10-22
Populate UpperCase validator whitelist with NSI data

### LINKED ISSUE #1428: Exclude NSI entries from "uppercase name" warning
Item 5010
Class 803
https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_UpperCase.py

Example: 
- [error](https://osmose.openstreetmap.fr/en/issue/21256ea0-c863-9236-b486-0ef34d229685)
- [node](https://www.openstreetmap.org/node/5086281758)
- `name = "ICI PARIS XL"`
- name suggestion index [suggests](https://nsi.guide/index.html?t=brands&k=shop&v=perfumery&tt=ICI%20paris%20xl) `name=ICI PARIS XL`

According to the Name Suggestion Index, the name of this store is in capitals. It would be a nice enhancement if Osmose would in this case not warn about an upper case name.

malteger:
This adds support for NSI data also to the UpperCase plugin. The whitelist from NSI data and the whitelist as defined in the plugin are merged and used in the validator.

Follow up to #1607 

Fixes #1428

### DIFF
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -22,12 +22,13 @@
 from modules.OsmoseTranslation import T_
 from plugins.Plugin import Plugin
 import regex as re
+from modules.downloader import urlread
+import json
 
 # Whitelist of allowed capitals by country code
 UpperCase_WhiteList = {
-    "FR": ["CNFPT", "COSEC", "EHPAD", "MACIF", "MEDEF", "URSSAF"],
+    "FR": ["CNFPT", "COSEC", "EHPAD", "MEDEF", "URSSAF"],
     "ES": ["FREMAP"],
-    "DE": ["ALDI", "EDEKA", "MARKTKAUF", "PENNY", "REWE"],
 }
 
 class Name_UpperCase(Plugin):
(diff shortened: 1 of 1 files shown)


Famlam:
#1428  
Are the "hardcoded" whitelists still needed?

frodrigo:
> #1428 Are the "hardcoded" whitelists still needed?

FR and ES list are not brand part of NSI.

frodrigo:
Adding NSI the brand list should be rewrite like this
```
    "FR": ["CNFPT", "COSEC", "EHPAD",  "MEDEF", "URSSAF"],
```

frodrigo:
@malteger Thank you to do this. Can you change the whitelists content, please.

malteger:
I changed the french whitelist according to your suggestion and removed the german one again, as it only contains brand names which are covered by NSI.

I think it is good to keep the hardcoded whitelist for things which are not brands.

Famlam:
> I think it is good to keep the hardcoded whitelist for things which are not brands.

Agree, apologies for being unclear, I indeed only meant removing the "now dealt-with via NSI" items

plugins/Name_UpperCase.py
```diff
@@ -22,12 +22,13 @@
...
 from plugins.Plugin import Plugin
 import regex as re
+from modules.downloader import urlread
+import json
 
 # Whitelist of allowed capitals by country code
 UpperCase_WhiteList = {
-    "FR": ["CNFPT", "COSEC", "EHPAD", "MACIF", "MEDEF", "URSSAF"],
+    "FR": ["CNFPT", "COSEC", "EHPAD", "MEDEF", "URSSAF"],
     "ES": ["FREMAP"],
```
frodrigo, line 31: Spanish is also a brand.

```suggestion
```

frodrigo:
Merged, but not deployed yet. Thank you.

PR #1406, opened 2022-02-20
Numerical name plugin

arrival-spring:
Fixes #1319 

First time contributing and haven't been able to test this.

PR #1614, opened 2022-10-23
Mapcss rules specific for NL

Famlam:
I would like to propose to add 'my' mapcss rules that I maintain [for the Dutch community](https://forum.openstreetmap.org/viewtopic.php?id=74817) to Osmose.

Feel free to reject if you disagree.


frodrigo:
I am OK with that.
Note test are failing.

Note, you can add `-osmose*` fields to have a better integration into Osmose. Not just all in one "default" item, as adding osmose tag and use thematic items/class is better.

Famlam:
> I am OK with that.

Thanks!

> Note test are failing.

Noticed, I had to rearrange my rules a bit to not use a class. In JOSM's internal tests, it seems classes set by 'earlier' rules are not evaluated when evaluating asserts. It's fixed now, I just rewrote the rules to not use a class.

> Note, you can add -osmose* fields to have a better integration into Osmose.

I must admit that I'm not too much a fan of this, as it would mean that every time that I change an item (in the rules or in Osmose) I first have to search all existing numbers, spread over the different files, to see if an item/class combination already exists (to avoid i.e. https://github.com/osm-fr/osmose-backend/pull/1406#issuecomment-1094119547 ). Furthermore I agree with Klumbumbus (in the other PR) that in normal use you cannot really filter on classes anyway, so random classes are fine with me. So I like to have it simple: in one item, with automatically generated classes. 
(But if there's a way to say in item_map.py that i.e. group "NL Addresses and Contacts" should be connected to tag "addr" in Osmose, I'm more than happy to do so. Doing so for every individual rule using -OsmoseTags is a bit overzealous, as everything in a group will have the same tags; that's why they're grouped together anyway)

frodrigo:
Update in progress
http://osmose.openstreetmap.fr/en/issues/open?item=9020&useDevItem=all

Famlam:
Results look as expected, just discovered that I accidentally also included Caribbean Netherlands (for which several rules don't apply) by selecting `NL`; made #1620 to fix that

frodrigo:
When ready. Ping me or @jocelynj to make the new item item public.

Famlam:
Will do! I'll wait until the Caribbean have disappeared from the results, then have a last look for major issues, then report back here.
Thanks!

frodrigo:
Caribbean will not disappear. We have to remove manually from the frontend. Tomorrow.

Famlam:
Oops.. Sorry for the extra work...

Famlam:
I'm happy with the current results!
Yesterday and today I removed some minor issues that Osmose allowed me to spot, which are included in #1623 together with the fix for the French mapcss rules

So if you could remove the results for the Caribbean and make it public, that would be appreciated :)

Famlam:
> When ready. Ping me or @jocelynj to make the new item item public.

New ping if needed :)

frodrigo:
Done

Issue #1600, opened 2022-10-14
Osmose needs a control with higher prio than typo_check


iiitomo:
Currently by default (ie with at least all Severe controls enabled), Osmose overall gives bad suggestions for the tags below:
link  (suggests 'line' instead of 'website')
short (suggests 'sport' instead of 'short_name')
sale (suggests 'salt')
retail
ramp (suggests 'lamp')
mine
marking
brunch
And probably more

These bad suggestions come from the 'typo_check' control, and this is not a bug of the typo_check control (which works as expected by giving typo fix suggestions), but it is an issue of the Osmose current design which assumes these are typos, instead of assuming these are bad tag usages.
Osmose should have other controls (new or existing) that make correct suggestions for these tags, and that have higher priority/Severity than 'typo_check' , so that correct suggestions appear first by default.

iiitomo:
Thx Famlam, but as far as i understand, https://github.com/osm-fr/osmose-backend/issues/1483 will not solve this issue, because typo_check will continue to propose tags that are close. To solve this issue, we need to blacklist some tags in typo_check, and have another control to propose the correct fix.

Famlam:
Well, yeah, but suggesting (as example) line rather than link is correct behavior and should not be blacklisted. 1483 will prevent converting of valid tags to invalid tags (ramp to lamp for example)

iiitomo:
I've seen hundreds usage of 'link' where the value was a URL. And none where it was a tipo for 'line'. So if we don't blacklist tipo_fix, then at least we should have another control to propose a more appropriate fix (e.g. 'website') before tipo_fix.

Famlam:
So how would you propose we can distinguish between a typo and a wrong key? If someone has `marking=yes`, how do we know it should "actually" be lane_marking, road_marking or parking? Same for link: it can be a typo for line, but it could be website, contact:website, brand:website, or someone who wants to indicate that it's a linking feature between any kind of route.
Remember, it are suggestions, not requirements. Your Word (or whatever text editor) spell checker also cannot read minds on what is actually meant.

If it happens a lot, you might want to go through the deprecation process to get a key formally deprecated, then have i.e. JOSM (whose rules are imported in Osmose) create an warning.

frodrigo:
Done using tagingo.

Issue #1569, opened 2022-09-26
False-positive for tag 'grades'


iiitomo:
hi
Osmose currently proposes to change 'grades' into 'grade'.
See screenshot below.
It shouldn't as 'grades' is a valid tag, see https://wiki.openstreetmap.org/wiki/Key:grades

![osmose_error_suggestion_for_grades](https://user-images.githubusercontent.com/58949423/192223581-52b33192-9b9d-4d8e-8cca-cd503e74af36.PNG)

frodrigo:
All this suggestions are based on statistics base on territorial data. It may be biased on local tag usage. But yes, keep reporting false positives.

Famlam:
This will also be fixed with #1483

frodrigo:
`grade` is not a documented tag.

It is proposed because it use in the surrounding. The issue is with the surrounding data. Maybe it should be explained in the popup.

frodrigo:
Keep this one open to remind to add more doc in the popup.

Famlam:
> Keep this one open to remind to add more doc in the popup.

I hope the suggested change in #1631 matches what you were thinking of?
(Together with the information in the side bar I think it should be clear enough that it's just about counting which tag is more common)

PR #1616, opened 2022-10-23
Tag_Typo: check keys with documented keys using TagInfo

### LINKED ISSUE #1483: Use TagInfo API to determine whether keys exist or might be typos
Per https://github.com/osm-fr/osmose-backend/pull/1470#issuecomment-1156929246

TagInfo has a method to obtain all existing keys with wiki pages: https://taginfo.openstreetmap.org/taginfo/apidoc#api_4_keys_wiki_pages

We could use this (rather than statistics and manual whitelisting) to find misspelled keys in [analyser_osmosis_tag_typo](https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_tag_typo.py). This is much less likely to result in false positives

- [x] Check if TagInfo allows this usage (I think it does - see its about-page - as long as we mention its name)
- [ ] Fetch and cache the results of the existing wiki pages on the Osmose servers
- [ ] Adjust check to use the keys with wiki pages to check if it exists. If so -> pass, if not -> propose replacing with a close match if any

Famlam:
Fixes #1483, see #1569 and #1600

Implementation:
1. Download the list with existing wiki pages from TagInfo, filter to only get the relevant keys
2. Make sure that the "less-abundant key" is NOT a documented key
3. Also added a generic filter for numbered keys (addr2 etc)
4. Add new tags found during tests that are documented/import noise but do not have their own wiki page: `dcgis`, `linen`, `lwn_ref` and generic filter for `expected_*n_route_relations`
5. Grouped the remaining exceptions in:
 - Keys resulting from imports (which would just generate noise)
 - Keys that are documented on pages, without their own Wiki page (yet)
 - Legacy keys from early versions where I couldn't find the origin (added in https://github.com/osm-fr/osmose-backend/commit/c5b3270f8078116ac895c26f3dcf088cbc15e4d9, https://github.com/osm-fr/osmose-backend/commit/13079eacbda2ea04e42b0a3674dd794b152836f5 ) or used by third-party presets

Checked with: ecuador, germany_bayern_mittelfranken, israel, indonesia_west_java, canada_british_columbia, russia_moscow, usa_illinois, portugal, ireland, sierra_leone, denmark, france_ile_de_france_seine_et_marne, japan_chubu, netherlands_noord_brabant, netherlands_gelderland, senegal
Verified that i.e. `ramp -> lamp` (e.g. in MittelFranken) nor `survey -> surrey` (Columbia) do not exist anymore (and some others), all numbered keys are gone (e.g. Israel), and all valid errors remained

### DIFF
--- a/analysers/analyser_osmosis_tag_typo.py
+++ b/analysers/analyser_osmosis_tag_typo.py
@@ -148,7 +148,7 @@ class Analyser_Osmosis_Tag_Typo(Analyser_Osmosis):
 
     def get_keys_wiki_taginfo(self):
         # See https://taginfo.openstreetmap.org/taginfo/apidoc#api_4_keys_wiki_pages
-        # TagInfo has a JSON file with all wiki pages
+        # TagInfo has a JSON file with all keys with wiki pages
         taginfo_url = "https://taginfo.openstreetmap.org/api/4/keys/wiki_pages"
         json_str = urlread(taginfo_url, 30)
         json_entrylist = json.loads(json_str)['data']
--- a/analysers/analyser_osmosis_tag_typo.py
+++ b/analysers/analyser_osmosis_tag_typo.py
@@ -23,6 +23,8 @@
 from modules.OsmoseTranslation import T_
 from modules.Stablehash import stablehash64
 from .Analyser_Osmosis import Analyser_Osmosis
+from modules.downloader import urlread
+import json
 
 sql10 = """
 CREATE TEMP TABLE rtag_{type} AS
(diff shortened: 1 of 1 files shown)


Famlam:
Maybe good to consider: what's the desired behavior if TagInfo changes it's structure (but keeps the same url)? 
Should it fail (with/without error) or silently proceed without the whitelists? Currently it tries to proceed.

frodrigo:
> Should it fail (with/without error) or silently proceed without the whitelists? Currently it tries to proceed.

No. No silently continue on error. Especially here, as it will raise lot of false positives.

Famlam:
> No. No silently continue on error.

Done 👍

Famlam:
If it helps the reviewing, attached are the [results](https://github.com/osm-fr/osmose-backend/files/9889751/results.zip)

frodrigo:
Nice and simple! Looks good to me.
And I learn a new SQL key word `SIMILAR TO`, probably more efficient than regex!

I have doubt on perf at `t1.key NOT IN ({wikikeys})` with a very long list, but it will probably be ok.

frodrigo:
Thank you for solving this recurring issue.

Famlam:
> I have doubt on perf at t1.key NOT IN ({wikikeys}) with a very long list, but it will probably be ok.

I was playing with the idea of maybe making a temp-table to pre-filter `t1`, before using it in `sql20`, but I don't know enough about how SQL evaluates its conditions to know if this would really improve performance. (If `sql20` is practically a double "for loop" over t1 and t2 it would definitely be more efficient)

frodrigo:
Deployed

Marc-marc-marc:
```
 -- Documented keys without their own key:* wiki page. Between () where it's documented
 'dist', -- vs dirt, list, diet (golf=hole)
 'hide', -- vs site, sides, side (amenity=hunting_stand)
```

I have added a wiki redirect-to-doc page for those

`            'levels', -- vs level (building:levels)`
I don't understand this one : according to https://wiki.openstreetmap.org/wiki/Key:building:levels , levels is an invalid tag (the correct one if often building:levels, some may be level), not a "Documented keys without their own key:* wiki page"

Famlam:
Nice, so after the cache period these exceptions can go

Regarding 'levels': I considered it as documented as a deprecated tagging method. The suggestion (level) was bad anyway

Famlam:
From 90k to 60k, pretty effective :)
https://osmose.openstreetmap.fr/en/issues/graph.png?item=3150

frodrigo:
Nice. thank you. And it will stop a flux of new github issues!

PR #1619, opened 2022-10-24
Check if country exists

Famlam:
And provide a human readable error if not (and skip the country)

I accidentally ran Osmose with `--country=aruba` rather than `--country=netherlands_aruba` and got
```
  File "/opt/osmose-backend/./osmose_run.py", line 589, in <module>
    sys.exit(main(options))
  File "/opt/osmose-backend/./osmose_run.py", line 477, in main
    country_conf = config.config[country]
KeyError: 'aruba'
```

I prefer a human readable error instead :)

### DIFF
--- a/osmose_run.py
+++ b/osmose_run.py
@@ -474,7 +474,12 @@ def main(options):
     # analyser
 
     for country in options.country:
-        country_conf = config.config[country]
+        if country in config.config:
+            country_conf = config.config[country]
+        else:
+            logger.log("Failed to load country {0}".format(country))
+            continue
+
         if os.getenv('SENTRY_DSN'):
             sentry_sdk.set_tag('country', country)
 
--- a/osmose_run.py
+++ b/osmose_run.py
@@ -478,7 +478,7 @@ def main(options):
             country_conf = config.config[country]
         else:
             logger.log("Failed to load country {0}".format(country))
-            return 2
+            return 8
 
         if os.getenv('SENTRY_DSN'):
             sentry_sdk.set_tag('country', country)
--- a/osmose_run.py
+++ b/osmose_run.py
@@ -478,7 +478,7 @@ def main(options):
             country_conf = config.config[country]
         else:
             logger.log("Failed to load country {0}".format(country))
-            continue
+            return 2
 
         if os.getenv('SENTRY_DSN'):
             sentry_sdk.set_tag('country', country)


frodrigo:
Ok. But would prefer to stop with error on bad country.

Famlam:
No problem, do you have a preference for a specific return code?
`return 1` I assume?

frodrigo:
> No problem, do you have a preference for a specific return code? `return 1` I assume?

No, there is already many return codes. Pick a new, or a similar one.

frodrigo:
> > Pick a new, or a similar one.
> 
> Ok. It seems `return 2` is still free, so I'll use that ;)

I code 2 here https://github.com/osm-fr/osmose-backend/blob/master/osmose_run.py#L335

Famlam:
8 then?

jocelynj:
Looks good, thanks!

Just a remark: you could also have used try/except instead of a if. I will still merge this change, as it is good enough :)

jocelynj (approved):

jocelynj:
I just saw after merge that you used logger.log instead of logger.err: using `err` has the main advantage of making the printed string in red. Do you want to do this change?

Famlam:
Thanks!

> Just a remark: you could also have used try/except instead of a if. I will still merge this change, as it is good enough :)

Yeah, I usually try to avoid try/except if it can be guarded against easily :)

> I just saw after merge that you used logger.log instead of logger.err: using err has the main advantage of making the printed string in red. Do you want to do this change?

I can do that, do you also want me to do it for (any or all of) the following lines:
https://github.com/osm-fr/osmose-backend/blob/5ee4af8056891e028a28dc429f1d2e1324c21e47/osmose_run.py#L433
https://github.com/osm-fr/osmose-backend/blob/5ee4af8056891e028a28dc429f1d2e1324c21e47/osmose_run.py#L459
https://github.com/osm-fr/osmose-backend/blob/5ee4af8056891e028a28dc429f1d2e1324c21e47/osmose_run.py#L464
https://github.com/osm-fr/osmose-backend/blob/5ee4af8056891e028a28dc429f1d2e1324c21e47/osmose_run.py#L468
https://github.com/osm-fr/osmose-backend/blob/5ee4af8056891e028a28dc429f1d2e1324c21e47/osmose_run.py#L480

jocelynj:
Yes, good idea to change these lines too.

I'm not sure why these lines use "log_av_b" to print in blue. @frodrigo : do you agree that we should remove the blue color on these error lines, which means that they will be printed in red?

PR #1620, opened 2022-10-24
Limit Dutch rules to continental NL

Famlam:
I hadn't realized that `NL` also included the islands we have at the other side of the ocean, but on those islands part of the rules do not apply, hence I'm limiting the execution to continental Netherlands only. (The Caribbean part has country code "NL" in Osmose, hence I cannot use that as a unique filter)

It also includes a little change I made today to the mapcss file due to calling `mapcss.update`. Edit: and one fix which I spotted while randomly clicking through some Osmose issues (which was actually an issue in the mapcss).

### DIFF
--- a/mapcss/item_map.py
+++ b/mapcss/item_map.py
@@ -14,7 +14,7 @@ item_map = \
                    'url_display': 'https://josm.openstreetmap.de/wiki/Rules/ItalySpecific'},
  'DutchSpecific': {'prefix': 'Josm_',
                    'item': 9020,
-                   'only_for': ['NL'],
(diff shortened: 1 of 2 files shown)


Famlam:
When I look at https://osmose.openstreetmap.fr/en/issues/open?item=9020&limit=1&useDevItem=all it seems that Caribbean is still updating (age column shows 2h). When I ran it locally, it was skipped. Any clue what may cause this 🤔?

frodrigo:
> When I look at https://osmose.openstreetmap.fr/en/issues/open?item=9020&limit=1&useDevItem=all it seems that Caribbean is still updating (age column shows 2h). When I ran it locally, it was skipped. Any clue what may cause this thinking?

Was not manually cleaned yet. Just done. Note, there is cache on the page.

PR #1622, opened 2022-10-26
Update merge_charging_station_FR to use IRVE v2 scheme

PanierAvide:
Input data has changed scheme since last year, so this PR intends to use latest format for charging stations in France.

For reference : https://wiki.openstreetmap.org/wiki/France/data.gouv.fr/Bornes_de_Recharge_pour_V%C3%A9hicules_%C3%89lectriques


nlehuby:
Fix #1622

frodrigo:
Ok. Thank you. Merged but not deployed yet.

Issue #1618, opened 2022-10-24
country not France for item=9019 and class=9019004 


didier2020:
9019 is for "spécific France"
but there is another country : see at http://osmose.openstreetmap.fr/fr/issues/open?item=9019&class=9019004

Famlam:
Seems it needs to add `[inside("FR")]` to `node[highway=motorway_junction][ref=~/^\D/]`

Famlam:
Update is failing on an assert... @frodrigo do you know who to cc for the French mapcss rules?

frodrigo:
From history https://josm.openstreetmap.de/wiki/Rules/FranceSpecificRules?action=history
all last edit are from @didier2020 

There is one `assertMatch` in place of `-osmoseAssertNoMatchWithContext`, it is not the issue ?

Famlam:
Ah the topic starter, oops 0:) 
Yeah, I had a similar issue (but just discarded the asserts) that Osmose evaluates inside(...) as false in asserts, while JOSM evaluates them as true.

Famlam:
I figured I could also change this myself as from the remainder of the file it's clear that the `-osmoseAssert*` syntax is preferred over the `assert*`. Now it works https://github.com/osm-fr/osmose-backend/pull/1623

After the update has been live (and all non-France country results are gone again), it may be an idea to limit it to France by default in the config, similar to that Brazilian rules are limited to Brazil, Russian to RU, German to DE? Saves the server a lot of checks that only end when `and (mapcss.inside(self.father.config.options, 'FR'))` is evaluated, which is done after all other mapcss rule evaluations. Practically all rules are restricted to `[inside("FR")]` anyway. But as a Dutch person, I'll leave this to someone who's actually from France :)

Brazil example:
https://github.com/osm-fr/osmose-backend/blob/master/mapcss/item_map.py#L86

frodrigo:
Done & fronted clean.

PR #1629, opened 2022-10-29
Restrict Italian and France mapcss rules to those countries

Famlam:
Given that all rules in the Italian an France mapcss rules have either country specific items in the selectors (ex. `*["ref:FR:SFR"]["telecom:medium"!=fibre],`), or `[inside("FR"/"IT")]`-clauses (e.g. `*[operator=ERDF][inside("FR")]`), don't run the mapcss plugins on other countries than those. 
Should be more efficient for the server as it doesn't need to execute on other countries. Additionally it prevents possible issues like #1618 in the future

For example, in any country but France, this entire code is executed `match = ((mapcss.regexp_test(mapcss._value_capture(capture_tags, 0, self.re_045a0f34), mapcss._tag_capture(capture_tags, 0, tags, 'name'))) and (mapcss._tag_capture(capture_tags, 1, tags, 'amenity')) and (mapcss._tag_capture(capture_tags, 2, tags, 'amenity') != mapcss._value_const_capture(capture_tags, 2, 'car_pooling', 'car_pooling')) and (not mapcss._tag_capture(capture_tags, 3, tags, 'carpool'))` before it encounters  `and (mapcss.inside(self.father.config.options, 'FR')))` and aborts

To be confirmed with someone from France for i.e. the first type.

p.s. preferably wait merging until all non-France items are 0 again (e.g. #1623 is deployed to fix the original #1618)


frodrigo:
Thank you. Take it any way to go faster.

Famlam:
Thank you, but merging it now probably means you'll have to manually delete the non-french results, right?

frodrigo:
> Thank you, but merging it now probably means you'll have to manually delete the non-french results introduced by #1618, right?

Yes. But it is right in all case.

Since an analyser no more report for a country, results are "lost" in the frontend.

PR #1630, opened 2022-10-29
Separate whitelists for similar and equal values

Famlam:
Currently, there is one whitelist list for both similar values and equal values. 
Therefore, if you want to whitelist a key that can have similar values (such as `opening_hours`), you also whitelist exact duplicates.

In this PR:
- Use separate whitelists for keys that can have equal+similar values, and keys that can only have similar, but not equal values
- Adds a couple new exceptions for keys that can be similar, yet not duplicates (`destination`, `email`, `traffic_sign:[direction]`, `communication:mobile_phone`, `cuisine`, `operator`, `*:date` (like: `survey:date`), `*_name` (like `old_name`, `artist_name`, ...))
- removes the exceptions for `max*` (as to my understanding, those were added for `max*:conditional`, which later got its generic whitelist)
- Gets rid of the 'interesting' try...raise...except approach :). (Note I didn't use `any(map(...))` because `map` would first execute all regexes; however I can use that if desired)
- fixes a typo in the doc

P.s. I kept `ref` in "equal AND similar are allowed", because there's a test case `{"ref":"E 05; E 70; E 05;E 70; E 05;E 70; E 05;E 70; E 05;E 70"},`. I'm however not sure if duplicate values are truly valid?

### DIFF
--- a/plugins/TagFix_DuplicateValue.py
+++ b/plugins/TagFix_DuplicateValue.py
@@ -76,7 +76,6 @@ Ensure the interpretation of the tag does not change when you delete one item.''
             'communication:mobile_phone',
             'cuisine',
             'operator',
-            'source',
         ))
         self.WhitelistSimilarEqualRegex = set(( # Regexes for keys that can have similar and even equal values
             re.compile('seamark:.+:colour'),
--- a/plugins/TagFix_DuplicateValue.py
+++ b/plugins/TagFix_DuplicateValue.py
@@ -37,7 +37,9 @@ similar.'''),
             fix = T_(
 '''Delete one value.'''),
             trap = T_(
-'''In some case all values maybe required.'''))
+'''In some cases all values may be required.
+
+Ensure the interpretation of the tag does not change when you delete one item.'''))
         self.errors[3060]  = self.def_class(item = 3060, level = 3, tags = ['value', 'fix:chair'],
             title = T_('Duplicated values'),
             **doc)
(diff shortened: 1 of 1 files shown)


plugins/TagFix_DuplicateValue.py
```diff
@@ -99,37 +115,45 @@ def levenshtein(self, s1, s2):
...
+                if k in self.WhitelistSimilar:
+                    continue # Key may have similar values
+                if self.anyRegexMatch(self.WhitelistSimilarRegex, k):
+                    continue # Key may have similar values
+
+                vs_long = filter(lambda w: len(w) > 6, vs)
+                for v1,v2 in itertools.combinations(vs_long, 2):
+                    if abs(len(v1)-len(v2)) < 4 and self.levenshtein(v1, v2) < 4:
+                        err.append({"class": 30601, "subclass": stablehash64(k),
+                                    "text": T_("Duplicated similar values {key}={val}", key = k, val = tags[k])})
```
frodrigo, line 155: Yes,  you not changed this, but could be better.
```suggestion
                                    "text": T_("Similar values {v1} and {v2} in {key}={val}", v1 = v1, v2 = v2, key = k, val = tags[k])})
```

frodrigo:
Good idea, thank you.

plugins/TagFix_DuplicateValue.py
Famlam: I like it :)
It may be that the message sometimes becomes too long; in that case we should probably drop the `={val}` at the end, but lets see

PR #1631, opened 2022-10-30
Use wording that suggests it's about counting

### LINKED ISSUE #1569: False-positive for tag 'grades'

Famlam:
Rather than `uncommon_key -> common_key`, write `common_key is more common than uncommon_key, is uncommon_key a typo?`. Using 'more common' implies the analyser is just based on statistics.

(Also: add exception for `sale`, which JOSM uses in their motorcycle dealer preset, but isn't documented anywhere)

Fixes #1569


frodrigo:
Ok. Thank.

PR #1632, opened 2022-10-30
Port check-boundary-polygon to python3

jocelynj:
This script checks mainly if the OSM relation is contained inside the .poly file.

The beginning of the script shows a few failures:
```ested shells at or near point -0.14063139999999999 43.271786200000001
Error(france_aquitaine_pyrenees_atlantiques) boundary not valid (r_id=7450)
Error(france_basse_normandie_manche) poly inside boundary (r_id=7404, poly=http://download.openstreetmap.fr/polygons/europe/france/basse_normandie/manche.poly)
Nested shells at or near point -0.433172 48.863199399999999
Error(france_basse_normandie_orne) boundary not valid (r_id=7419)
Nested shells at or near point 5.4413933999999999 46.851811499999997
Error(france_bourgogne_saone_et_loire) boundary not valid (r_id=7397)
Error(france_bretagne_cotes_d_armor) poly inside boundary (r_id=7398, poly=http://download.openstreetmap.fr/polygons/europe/france/bretagne/cotes_d_armor.poly)
```

### DIFF
--- a/tools/check-boundary-polygon.py
+++ b/tools/check-boundary-polygon.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #-*- coding: utf-8 -*-
 
 ###########################################################################
@@ -21,6 +21,13 @@
 ###########################################################################
 
 import sys
+import logging
+
+logger = logging.getLogger('shapely.geos')
+logging_handler_out = logging.StreamHandler(sys.stdout)
+logger.addHandler(logging_handler_out)
+logger.setLevel(logging.INFO)
+
 from shapely.geometry import MultiPolygon
 
 sys.path.append("..")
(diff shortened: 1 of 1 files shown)


frodrigo:
Ok. Should we also run this in github actions ?

jocelynj:
> Ok. Should we also run this in github actions ?

There might be a bit too many falis at the current time to enable it in github actions, but it might still be worth launching weekly.

frodrigo:
> > Ok. Should we also run this in github actions ?
> 
> There might be a bit too many falis at the current time to enable it in github actions, but it might still be worth launching weekly.

Yes. I was thinking like the weekly relation check. Just one more check like this https://github.com/osm-fr/osmose-backend/blob/master/.github/workflows/poly.yaml#L25-L27

PR #1634, opened 2022-10-31
log -> err

Famlam:
As requested in https://github.com/osm-fr/osmose-backend/pull/1619#issuecomment-1296353313 by @jocelynj 

Now I replaced `logger.log` by `logger.err` for all errors. However, please let me know if you'd like me to either remove the blue color and/or restore `logger.log` for the ones with blue color. Using `logger.err` prefixes the word `error: ` to the output in red, so it looks kind-of funny at the moment when the text after it is blue.
Some examples in the picture below:
image

From a practical point of view, I find red easier to read than blue, but that's just a preference I guess.

### DIFF
--- a/osmose_run.py
+++ b/osmose_run.py
@@ -430,7 +430,7 @@ def main(options):
         logger = OsmoseLog.logger(output, True)
 
     if options.change_init and not options.change:
-        logger.err(logger.log_av_b+"--change must be specified "+logger.log_ap)
+        logger.err("--change must be specified")
         return 1
 
     #=====================================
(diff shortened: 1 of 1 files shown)


jocelynj:
I think it would be better to remove the blue text, and only keep the default red color from `logger.err`.

Famlam:
> I think it would be better to remove the blue text, and only keep the default red color from `logger.err`.

Done 👍

jocelynj:
Thanks!

jocelynj (approved):

Issue #1640, opened 2022-11-05
add `source:maxspeed=AT:urban:40` in TagFix_Maxspeed.py


mcliquid:
**Item**: [3032](https://osmose.openstreetmap.fr/de/issues/open?status=open&item=3032)
**Class**: [303241](https://osmose.openstreetmap.fr/de/issues/open?item=3032&class=303241&country=austria)
**Location**: Austria

In Austria, it is common there to have various speed limits in built-up areas.

Under the town sign, a lower speed is then posted with the addition "Excluding priority roads" or more explicitly "Excluding L***".
In practice, the main thoroughfare in the town then has 50 km/h and all the others 40 or 30 km/h. The sign for this looks like this: https://commons.wikimedia.org/wiki/File:St_Gerold-town_sign-01ASD.jpg or https://commons.wikimedia.org/wiki/File:Nenzing-town_sign-01ASD.jpg.

In Austria, source:maxspeed=AT:urban40 and source:maxspeed=AT:urban:40 have become established for this purpose.

There should be an exception for both variants.

frodrigo:
> Note: I've messed something up. See https://community.openstreetmap.org/t/osmose-meckert-uber-tempo-50-im-ortsverkehr/5113/

Please, can you provide an english summary.

frodrigo:
> Would the solution be as follows? `'at:urban': ['30', '40', '50', '60'],`

We can do this. But should not b ?
```
'at:urban': ['30', '40', '50', '60'], // or just 50 ?
'at:urban40': 40,
'at:urban30': 30,
'at:urban60': 60,
```

frodrigo:
Done.

PR #1644, opened 2022-11-07
Fix 1640: Updated values and rules for AT:urban

mcliquid:
German discussion: https://community.openstreetmap.org/t/osmose-meckert-uber-tempo-50-im-ortsverkehr/5113

Solution from @frodrigo in #1640 

Thanks!


Famlam:
FYI, `at:urban` is 50, which is also the default (see line 29), so no need to specify that one ;)

Famlam:
That's probably because the current live version says the max speed in urban is 30 or 40. (I guess that's related to the conflict warning you see)

https://github.com/osm-fr/osmose-backend/commit/0c7b71007e0cd889e4f530164807bd11b3323173

Famlam:
I'd think so; @frodrigo ?

frodrigo:
If you define an at:urban rule it override the default and replace the value 50.

Famlam:
You'll have to wait until one of the admins has time ;)

Meanwhile, I see "This branch has conflicts that must be resolved". Maybe you can resolve it?

frodrigo:
Also reverting in OSRM https://github.com/Project-OSRM/osrm-backend/pull/6450

PR #1647, opened 2022-11-13
Improve power lines analyser about lines terminations 

flacombe:
Recent power proposals introduced line_management=* and transformer=main for substations supported by power poles.
Current roll out cause false positives on Osmose, like such https://osmose.openstreetmap.fr/fr/issue/f64f3096-a46e-eb8d-7931-38bc9cce76da

Furthermore, older false negatives have to be addressed with a lower threshold about terminators proximity. I propose to change 150m seek distance by 50m.
Such nodes should get line_management=termination to be valid
https://www.openstreetmap.org/node/1365158230

Lone power supports can also be portal, terminal or insulator.

It is expected to improve actual Osmose's alert count on 7040 classes 1, 6 and 7
Particularly in France, where recent surge is caused by transformer=main roll out.
https://osmose.openstreetmap.fr/fr/issues/graph.png?item=7040&class=6&country=france*

### DIFF
--- a/analysers/analyser_osmosis_powerline.py
+++ b/analysers/analyser_osmosis_powerline.py
@@ -24,6 +24,7 @@ from modules.OsmoseTranslation import T_
 from .Analyser_Osmosis import Analyser_Osmosis
 from modules.Stablehash import stablehash64
 
+# Lone power supports
 sql10 = """
 SELECT
     nodes.id,
@@ -40,7 +41,7 @@ FROM
 WHERE
     nodes.tags != ''::hstore AND
     nodes.tags?'power' AND
-    nodes.tags->'power' IN ('pole', 'tower')
+    nodes.tags->'power' IN ('pole', 'tower', 'insulator', 'terminal', 'portal')
 GROUP BY
     nodes.id,
     nodes.geom
@@ -48,6 +49,7 @@ HAVING
     bool_and(ways.id IS NULL)
 """
 
+# Power lines ends with their voltages
 sql20 = """
 CREATE TEMP TABLE line_ends AS
 SELECT DISTINCT ON (ends(ways.nodes))
@@ -69,6 +71,7 @@ sql21 = """
 CREATE INDEX idx_line_ends_id ON line_ends(id)
 """
 
+# Unfinished lines ends
 sql22 = """
 CREATE TEMP TABLE line_ends1 AS
 SELECT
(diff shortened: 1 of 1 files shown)


frodrigo:
It requires to reset false positivies?

flacombe:
Can I see a list of them in France please?

Difficult to me to answer currently, I guess yes

flacombe:
These are not false positive:
* https://osmose.openstreetmap.fr/fr/false-positive/9c3a2627-bf14-00e5-a34b-2a6adbd82418 => should get location:transition=yes
* https://osmose.openstreetmap.fr/fr/false-positive/3e6c0d79-0d93-af67-5eb2-e39cda73722f => Actually unfinished line
* https://osmose.openstreetmap.fr/fr/false-positive/f8616208-d8e6-1fca-9229-45e00ff3fbc8 => The line doesn't exist anymore, but was probably a location transition

So yes, reset false positives please.

analysers/analyser_osmosis_powerline.py
```diff
@@ -94,9 +97,10 @@
     ) AS line_ends
     JOIN nodes ON
         line_ends.id = nodes.id AND
-        NOT (tags?'pole' AND tags->'pole' = 'transition') AND -- deprecated
         NOT (tags?'location:transition' AND tags->'location:transition' = 'yes') AND
-        NOT (tags?'transformer' AND tags->'transformer' in ('distribution', 'minor_distribution')) AND
+        NOT (tags?'transformer' AND tags->'transformer' in ('distribution', 'main')) AND
+        NOT (tags?'substation' AND tags->'substation' in ('minor_distribution')) AND
```
frodrigo, line 102:
```suggestion
        NOT (tags?'substation' AND tags->'substation' = 'minor_distribution') AND
```

frodrigo:
> I propose to change 150m seek distance by 50m.

Think globaly, not just of place with very good data quality. And better try to avoid false positives event if not issues all detected.
After this, if you still think 50m is right, I'm ok with that.

flacombe:
> Think globaly, not just of place with very good data quality. And better try to avoid false positives event if not issues all detected. After this, if you still think 50m is right, I'm ok with that.

To me it's not a matter of source quality but power lines architecture: 150m is wide and you may confuse two independent supports. With or without good aerial imagery.
50m is good with low scale satellite imagery as well.

Lint failed due to apt-get error, could you relaunch it please?

frodrigo:
Ok. Thank. Merged.

PR #1649, opened 2022-11-16
osmosis: Kill background processes on exception

jocelynj:


jocelynj:
@Marc-marc-marc : another change to improve stability of backends.

PR #1650, opened 2022-11-16
clean-old-jobs: Detect error when removing schema, to not make script failing

jocelynj:


jocelynj:
@Marc-marc-marc , @frodrigo : this should fix some issues around failures from this script.

Issue #1651, opened 2022-11-17
VORTAC


jidanni:
https://www.openstreetmap.org/node/5139092102 raises several errors, despite adhering to the relevant wiki pages.

frodrigo:
https://osmose.openstreetmap.fr/en/issue/15179368-8c7d-6059-1c00-1534b98c9fb0

Could be white listed here https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_UpperCase.py#L29

jidanni:
Also
> The object is missing any tag which defines what kind of feature it is. This is unexpected for something with a  name  tag.

> Add a top level tag to state what this feature is. Considered top level tags are (with derived  disused: ,  abandoned:  and  historic:  variants

Famlam:
Are you sure the name is valid? Looks like a descriptive name to me
See https://wiki.openstreetmap.org/wiki/Names section Names are not for descriptions

jidanni:
Well just like Lake Michigan is called Lake Michigan, and not just Michigan, because we wouldn't know if we need to call it Michigan Lake or Lake Michigan if we had to add the word Lake back on...

jidanni:
So if you want to remove the word VORTAC, then how will people distinguish it from a VOR-DME, NDB, etc.?

Anyway, I don't think the government ever used just a town name without the word of the device in the back.

Famlam:
Slightly offtopic, but the wiki gives plenty of examples ;). If the lake is actually called "Lake Michigan", then that's the true name. However, if it's just about differentiating the types of beacon, then there's the tag `beacon:type` already on the object (and more details). Tagging for the renderer/map as you seem to argue is not allowed in OSM, see the same wiki page. 
So if the thing is really called "San Simon VORTAC" then you're right, but if it's just as well known as "the VORTAC beacon near San Simon", it probably shouldn't have a name tag. I don't know, haven't been there. I just see that it's also referred to by many different names like shown here: https://flightaware.com/resources/airport/CFT/ODP/SAN+SIMON+ONE+(OBSTACLE)+(RNAV)/pdf

I've created a PR to fix the second issue you mention

jidanni:
Okay, I'll leave this up to the experts to decide, because I don't even know how to fly an airplane either. Thanks.

frodrigo:
Done for `airmark`.

PR #1653, opened 2022-11-17
Add airmark as main tag

Famlam:
See #1651


Famlam:
It seems the checks / Github actions are ran twice?

PR #1659, opened 2022-11-21
Split Chile in regions

jocelynj:


frodrigo (approved):
Probably the best country to split!

PR #1663, opened 2022-11-30
Remove unneeded whitelists

Famlam:
Credits to @Marc-marc-marc for creating the wiki pages of these 5 keys https://github.com/osm-fr/osmose-backend/pull/1616#issuecomment-1297305341

As the 30 day cache period is expiring today, it should now be safe to remove the whitelists, as they're in the TagInfo file
(Also clarified the `levels` comment)


frodrigo:
Thank you.

Issue #1665, opened 2022-11-30
no issue for missing properties if traffic_sign:forward=* is used


tiptoptom:
I think, this shouldn't be an issue?

https://osmose.openstreetmap.fr/en/issue/8e0e0928-3b03-17c6-6ad7-d8f2f33c8928

Untagged named object

The object is missing any tag which defines what kind of feature it is. This is unexpected for something with a name tag.

![grafik](https://user-images.githubusercontent.com/73886169/204790330-b610cc78-b1e0-4071-be1b-13bb58fd693c.png)

frodrigo:
https://osmose.openstreetmap.fr/en/issue/8e0e0928-3b03-17c6-6ad7-d8f2f33c8928

Untagged named object

The object is missing any tag which defines what kind of feature it is. This is unexpected for something with a `name` tag.

frodrigo:
Done.

PR #1666, opened 2022-11-30
Allow *:forward / *:backward as main tags

Famlam:
See #1665
Allow `[main tag]:forward` and `[main tag]:backward` as main tags


frodrigo:
Thank you.

Issue #1667, opened 2022-12-03
exempt highway=raceway from "One way inaccessible or missing parking or parking entrance" check


emersonveenstra:
`highway=raceway` are often not accessible via clearly defined paths for safety reasons, usually they're fenced in and isolated, so they should be exempted from item 1210. See https://osmose.openstreetmap.fr/en/issue/7d0b9977-058d-6c20-8085-48f8d07505d5 for an example

frodrigo:
Done, Thank @Famlam .

PR #1668, opened 2022-12-03
Only run analyser on real vehicle roads

Famlam:
Suggestions for a more generic fix (rather than just blacklisting raceway) #1667

Add level tags for all common roads in highway table generation
- Adds missing busway/bus_guideway to level=5
- Add level=6 for paths and the like
- Modify floating_islands to require level<=5, so the behavior of the analyser doesn't change by the addition of level=6

Require the level tag to exist for the deadend analyser, i.e. only if it's a real road, the analyser will run (so no longer on raceway)

Still to be tested


Famlam:
Closing this PR for now until 1670 is fixed

PR #1673, opened 2022-12-05
Exclude highway=raceway from highway_deadend

Famlam:
- Exclude `highway=raceway` from the osmosis_highway_deadend analyser as it is usually a closed circuit highway area (with one-directional traffic)
- Adds test case for several common cases (to reduce the chance of regressions when 1670 is dealt with. Note that it obviously doesn't contain the test case of 1670 as otherwise the tests would fail continuously)

See #1667

p.s. the reason for adding line 145 is because there are some raceways connected with service roads. If I just exclude raceways altogether, those service roads are suddenly considered "floating" and pop up as error. (This is part of the test cases)


frodrigo:
Thank you.

Issue #1241, opened 2021-04-03
supervised vs crossing:supervision


nuxper:
Item : 9015 
class : 9015029 
link to sample error : https://osmose.openstreetmap.fr/en/error/088af7ab-9414-1f1a-83cd-01373bb2caed

This analyser reports an error on a node because the key-value "supervised"="no" is used and suggests "crossing:supervision"="no" instead (node https://www.openstreetmap.org/node/2571357555 )

From what I could see:
- the wiki page for supervised exists but I couldn't find a documentation for crossing:supervision : https://wiki.openstreetmap.org/wiki/Key:supervised
- taginfo reports **8 396** use of crossing:supervision in the world ( https://taginfo.openstreetmap.org/keys/crossing%3Asupervision#overview ) versus **232 842** for "supervised" ( https://taginfo.openstreetmap.org/keys/supervised )

I'm not sure whether I should report this here or find the openrailwaymap repository.

frodrigo:
Yes, this rule come from the JOSM openrailwaymap validator. Please contact the project to fix it.
If you no able to do so and still think it as a bad check. We can disable it in Osmose anyway.

https://josm.openstreetmap.de/wiki/Rules

nuxper:
Ok thanks. I'll contact the project and link to this issue

frodrigo:
Issue reported, close here.

Famlam:
The fix is live :)
https://osmose.openstreetmap.fr/nl/issues/graph.png?item=9015&start_date=2022

PR #1676, opened 2022-12-07
Disable disputed rule: supervised=* is deprecated

Famlam:
See https://github.com/OpenRailwayMap/OpenRailwayMap/issues/720 (which got stale)
See https://github.com/osm-fr/osmose-backend/issues/1241

Also adds the blacklist ID to the mapcss files, to simplify debugging purposes


frodrigo:
Ok.
And good idea to output the class id.

PR #1687, opened 2022-12-12
Split Australia's New South Wales state in regions

jocelynj:


frodrigo:
Why do not use a `gen_country` like the other?

jocelynj:
No real reason, but this is aligned with other splits like Flanders or Bayern.

PR #1690, opened 2022-12-14
Config new Norway county layout

frodrigo:
Configure new Norway counties.

Merge of old counties, but also the change of county for one municipally.


jocelynj:
Note that https://en.wikipedia.org/wiki/Counties_of_Norway is saying that counties will change again in 2024 :(

jocelynj:
Do you know which municipality changed from one county to another?
I'm wondering if we can find the old counties boundaries on OSM, and switch to the new ones in 2024.

frodrigo:
https://www.openstreetmap.org/relation/409035

This municipality was merged with an other of an other admin_level=6 while the old and admin_levels was also merged to others.

For we use old admin_level=6 boundaries, except for this merged municipality.

jocelynj:
Instead of changing the diff/extract generation, could we add relation 409035 to region Nordland? This would be aligned with the old split.

(my main issue here is that "Troms og Finnmark" will be split in 2024, so I would prefer to not regenerate the diff/extracts until then)

frodrigo:
Yes, good idea.

PR #1691, opened 2022-12-14
Update config relation id : Taiwan / saint_pierre_et_miquelon

frodrigo:
- Fix relation for Taiwan, bad relation was used
- saint_pierre_et_miquelon, move from coast line to maritime boundary

### DIFF
--- a/osmose_config.py
+++ b/osmose_config.py
@@ -535,7 +535,7 @@ france_com = gen_country(None, country_base='france', download_repo=OSMFR, langu
 
 france_com(["central-america", "saint_barthelemy"], 537967, "FR-BL", proj=2969, phone_code="590", country="saintbarthelemy")
 france_com(["central-america", "saint_martin"], 1891583, "FR-MF", proj=2969, phone_code="590", country="saintmartin")
-france_com(["north-america", "saint_pierre_et_miquelon"], 233377, "FR-PM", proj=32621, phone_code="508", country="saintpierreetmiquelon")
+france_com(["north-america", "saint_pierre_et_miquelon"], 3406826, "FR-PM", proj=32621, phone_code="508", country="saintpierreetmiquelon")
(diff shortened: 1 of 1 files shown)


osmose_config.py
```diff
@@ -1208,7 +1208,7 @@ class gen(default_country):
 default_country("asia", "south_korea", 307756, {"country": "KR", "language": "ko", "proj": 32652}, download_country="south-korea")
 default_country("asia", "syria", 184840, {"country": "SY", "language": "ar", "proj": 32637})
 default_country("asia", "tajikistan", 214626, {"country": "TJ", "language": "tg", "proj": 32642})
-default_country("asia", "taiwan", 3777248, {"country": "TW", "language": ["zh_TW", "en"], "proj": 32651}, download_repo=GEOFABRIK)
+default_country("asia", "taiwan", 449220, {"country": "TW", "language": ["zh_TW", "en"], "proj": 32651}, download_repo=GEOFABRIK)
```
jocelynj, line 1211: Did you see that this relation includes some disputed islands near China's coast, that might be in another region analysed by osmose?

frodrigo: All Taiwan is disputed.

But this costal island are excluded on mainland China https://www.openstreetmap.org/relation/553303#map=7/25.488/119.399

jocelynj: In fact, this extract comes from Geofabrik, and is almost correct: http://download.geofabrik.de/asia/taiwan.html  (some maritime borders are missing)

jocelynj (approved):

osmose_config.py
frodrigo: On Geofabrik poly there is two missing areas on south.
https://www.openstreetmap.org/relation/449220#map=6/16.289/118.059

Issue #1679, opened 2022-12-08
 Item 1210 Class 4 only checks for connection inside admin level 4 borders?


Discostu36:
Many ways in [Appenzell Innerrhoden](https://www.openstreetmap.org/relation/1686666) are marked with Item 1210 Class 4: Small highway group apart from the main network or with insufficient access upstream.

## Probable Cause
There is no `highway=primary` inside this canton, and the validation seems not to check outside the canton. 

## Probable solution
Maybe the check should not be confined to admin boundaries, or the admin level should be changed.

## Example
Way: [193971383](https://www.openstreetmap.org/way/807587822)
Issue: https://osmose.openstreetmap.fr/de/issue/a0294c92-42c4-4ce3-63ee-05be45fefd6c

It is connected to a `highway=primary`:
https://www.openstreetmap.org/way/41212227
https://www.openstreetmap.org/directions?engine=graphhopper_car&route=47.42334%2C9.58764%3B47.41120%2C9.62585#map=14/47.4188/9.6087

Discostu36:
In this comment, it sounds as this was deliberate, but I don't agree:

https://github.com/osm-fr/osmose-backend/issues/1361#issuecomment-1001636658

Having to cross a sub-national boundary to get to a primary highway is usually a non-issue and should not be a reason for this warning to show.

Famlam:
@frodrigo I'm thinking along the line of adding something like
```sql
UNION
SELECT
  highways.linestring,
  highways.id
FROM
  relations AS ptrelation
  JOIN relation_members ON
    relation_members.type = 'W'
    relation_members.relation_id = relations.id
  JOIN highways ON
    highways.id = relation_members.member_id
WHERE
  ptrelation.tags!=''::hstore AND
  ptrelation.tags?'route' AND
  ptrelation.tags->'route' IN ('bus', 'trolleybus', 'tram')
```
to sql10, unless you know a way to properly deal with this border issue?

frodrigo:
> In this comment, it sounds as this was deliberate, but I don't agree:
> 
> [#1361 (comment)](https://github.com/osm-fr/osmose-backend/issues/1361#issuecomment-1001636658)
> 
> Having to cross a sub-national boundary to get to a primary highway is usually a non-issue and should not be a reason for this warning to show.

It is not on purpose, but a technical limitation. Data check is done on extract split by on some boundaries.

> @frodrigo I'm thinking along the line of adding something like

I does not se the goal.

[...]

PR #1700, opened 2022-12-17
Treat extract-border-crossing ways as connected

Famlam:
See #1679

If a road crosses an extract border, consider it connected (i.e. not a floating island)
For the case of #1679 (switzerland_appenzell_innerrhoden), only three ways remain (instead of the many false positives) :
way 378042935
way 418198089
way 753588065

### DIFF
--- a/analysers/analyser_osmosis_highway_floating_islands.py
+++ b/analysers/analyser_osmosis_highway_floating_islands.py
@@ -124,8 +124,6 @@ class Analyser_Osmosis_Highway_Floating_Islands(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
-        if not "proj" in self.config.options:
-          return
         self.classs[4] = self.def_class(item = 1210, level = 1, tags = ['highway'],
             title = T_('Small highway group apart from the main network or with insufficient access upstream'),
             detail = T_(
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_highway_floating_islands.py
```diff
@@ -110,6 +124,8 @@ class Analyser_Osmosis_Highway_Floating_Islands(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
+        if not "proj" in self.config.options:
+          return
```
frodrigo, line 128: Naybe the issue is requiring `highways` without `proj`, but nor directly here ?

Famlam: Well, this is actually because the table `highways` requires `proj` and a.o. opengeofiction doesn't have proj set.
https://github.com/osm-fr/osmose-backend/pull/1554#issuecomment-1245244954

frodrigo: Yes, please, remove from here and open a dedicated issue.

Famlam: I'll remove it. Not going to open an issue for such a minor thing, it's only because you asked me in the past to add it that I added it now too ;)

Famlam:
Note, still running on a broken polygon country per https://github.com/osm-fr/osmose-backend/issues/1679#issuecomment-1342936163, so don't merge yet

analysers/analyser_osmosis_highway_floating_islands.py
frodrigo: Hum, ok, but also bad idea last time. I made the issue #1701.

PR #1702, opened 2022-12-17
Remove duplicate warnings from two sources of deprecated tags

### LINKED ISSUE #1615: Duplicate fix from 2 controls
Currently, several errors ('access=public', 'noexit=no' , and probably more) are detected by 2 controls: 'deprecated value' and 'deprecated tagging'.
Hopefully, both suggest the same fix, but this is confusing and time-consuming for Osmose users.
There ideally should be no duplicate between controls.

Examples: https://www.openstreetmap.org/node/2071563065
https://www.openstreetmap.org/node/3340987280
https://www.openstreetmap.org/node/3340986379
https://www.openstreetmap.org/node/3340989980
https://www.openstreetmap.org/node/3340989982
https://www.openstreetmap.org/way/304914609

Another example:
IMG_20221023_211806
IMG_20221023_211824

Famlam:
Blacklist all entries from https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/deprecated.mapcss that also appear in https://wiki.openstreetmap.org/wiki/Template:Deprecated_features?action=raw (which is read by TagFix_Deprecated). Hence, they produce duplicate warnings.

In case the warnings differed significantly, I left them in.

Fixes #1615

(p.s. since it was a manual review of a veeeery long list, it is possible I may have missed one or two)


frodrigo:
Thank for time it could have request.

It does event make sens to keep deprecated.mapcss ? Should we rather whitelist a part of deprecated.mapcss ?

Famlam:
>It does event make sens to keep deprecated.mapcss ? Should we rather whitelist a part of deprecated.mapcss ?

I understand your point, but I would keep deprecated.mapcss in. It's much easier to blacklist rules when needed, than to check every time if something needs to be whitelist upon updating. (The worst that happens is that you may have a temporary double warning, no big deal IMO).
Maybe we should at some point in the future modify TagFix_Deprecated to load deprecated.mapcss, search for `\[deprecated_tag_as_regex\] ?[,{]` and if so, skip that line.

frodrigo:
Ok, make sense.

jajajaneeneenee:
Could this blacklisting perhaps be done at regular intervals? That would be great. Or is it already done that way?

What I find particularly annoying at the moment (so for some time now): `cycleway=opposite` (and related tag variants) – especially on oneway roads with many sections (= many warnings), closing 2 Osmose warnings each time after changes costs a lot of time (+ nerves) ... And this deprecated tag still exists on many streets. Just 1 warning would really be better.

Famlam:
The "related tag variants" I can blacklist (#2403). For `cycleway=opposite` I cannot do this, for the simple reason that `cycleway:left/right/both=opposite` aren't on the [wiki page with deprecated tags](https://wiki.openstreetmap.org/wiki/Template:Deprecated_features), and they're in the same rule at JOSM: (see [here](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/deprecated.mapcss), lines 2554 to 2557)

However, when you're working in a region with many of such `cycleway=opposite` tags, you can also temporarily disable the JOSM validator rules (or the wiki-based one) in the menu on the left. One or two days later the other ones should disappeared automatically.

Issue #1635, opened 2022-11-01
Detect cycles in relation containment


brawer:
Could Osmose detect cycles in the relation containment graph?

While writing some program that works with OSM relations, I found cycles in the containment graph. For example, [relation/2913585](https://www.openstreetmap.org/relation/2913585) is part of [relation/12762833](https://www.openstreetmap.org/relation/12762833), which is again part of the same relation where we started. If a naive implementation were to traverse this relationship graph, it would run into an infinite loop. This isn’t a super serious problem; checking for cycles is easy enough for data consumers. But it’s definitely a tagging mistake, so I wonder if Osmose could flag it.

In Switzerland, there’s currently three such cycles (as of November 1, 2022):
* [relation/2913585](https://www.openstreetmap.org/relation/2913585) ⟶  {relation/12762833, relation/12762834} ⟶  relation/2913585
* [relation/4012524](https://www.openstreetmap.org/relation/4012524) ⟶  {relation/12762831, relation/12762832} ⟶  relation/4012524
* [relation/9202820](https://www.openstreetmap.org/relation/9202820) ⟶  relation/13987412 ⟶  relation/9202820

Famlam:
Analyser is live:
https://osmose.openstreetmap.fr/nl/map/#source=&item=1160&class=2&zoom=7&lat=45.813&lon=9.206&level=1%2C2

Regarding "In Switzerland, there’s currently three such cycles":
You might want to hide your annexation plans a bit better: 2 of them are still in Italy as of today 😋

brawer:
Awesome, thank you! 😀

PR #1705, opened 2022-12-18
Add analyser to detect cyclic relations

Famlam:
New analyser which detects self-referring relations, either directly (`r1 > r2 > r1`) or via other relations (`r1 > r2 > r3 > r1`).
Restricted to 10 referrals (but can be increased if needed) to avoid potential infinite loops, and limited to relations which have a location (otherwise you'll just get a NULL location).

It doesn't return many results (as self-referring relations are pretty rare; JOSM also doesn't allow adding them) but in nearly all cases it finishes within a second.
[Some test results attached](https://github.com/osm-fr/osmose-backend/files/10254834/cyclic.zip)

#1635

### DIFF
--- a/analysers/analyser_osmosis_relation_cyclic.py
+++ b/analysers/analyser_osmosis_relation_cyclic.py
@@ -48,11 +48,10 @@ WITH RECURSIVE rcte AS (
   FROM
     rcte
     JOIN relation_members AS child ON
-      child.relation_id = rcte.member_id
-    WHERE
+      child.relation_id = rcte.member_id AND
       child.member_type = 'R' AND
       child.relation_id != rcte.start_id AND
-      nloops < {maxloops} -- avoid getting stuck in strange constructions
+      nloops < 10 -- avoid getting stuck in strange constructions
 )
 SELECT DISTINCT ON (start_id)
   start_id,
@@ -61,8 +60,7 @@ SELECT DISTINCT ON (start_id)
 FROM
   rcte
 WHERE
-  start_id = member_id AND
-  nloops < {maxloops}
+  start_id=member_id
 ORDER BY
   start_id,
   LENGTH(path) -- to get the shortest path selected by the DISTINCT ON call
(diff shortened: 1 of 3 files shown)


jocelynj:
That's a very nice idea! And I like the fact that a new test was added :)

analysers/analyser_osmosis_relation_cyclic.py
```diff
@@ -0,0 +1,114 @@
...
+  SELECT
+    rcte.start_id,
+    child.member_id,
+    nloops + 1 AS nloops,
+    CONCAT(rcte.path, ' > r', child.relation_id) AS path
+  FROM
+    rcte
+    JOIN relation_members AS child ON
+      child.relation_id = rcte.member_id
+    WHERE
```
frodrigo, line 52:
```suggestion
      child.relation_id = rcte.member_id AND
```

analysers/analyser_osmosis_relation_cyclic.py
```diff
@@ -0,0 +1,114 @@
...
+)
+SELECT DISTINCT ON (start_id)
+  start_id,
+  ST_AsText(relation_locate(start_id)),
+  path
+FROM
+  rcte
+WHERE
+  start_id = member_id AND
+  nloops < {maxloops}
```
frodrigo, line 65: Always true, doesn't ?

analysers/analyser_osmosis_relation_cyclic.py
Famlam: Just curiosity, it seems either produces the same result (unlike `LEFT JOIN`s), is there a benefit to either of them? I picked this one as it seemed easier to read to me, but will adjust it. Just curious

analysers/analyser_osmosis_relation_cyclic.py
Famlam: Indeed, no need to filter for nloops given the first condition

Famlam:
Not sure what happened with the "Merge branch '...' thing...

analysers/analyser_osmosis_relation_cyclic.py
frodrigo: ON and WHERE are very close, but not the same. Here it just for code readability. I personally prefer ON for condition on joined table, keep condition and table declaration close.

frodrigo:
You need to add the new analyser in the config file.

analysers/analyser_osmosis_relation_cyclic.py
Famlam: Thanks!

Famlam:
> You need to add the new analyser in the config file.

Thanks, done!
(I always run it with `--analyser=...`, hence I didn't notice)

Famlam:
Interesting, apparently there are even quite a lot of self-referring relations, for example: `r2200182 > r2200182`
Anyway, results look ok :)

Issue #1333, opened 2021-09-19
Tags/keys shouldn't be translated


Famlam:
I noticed in some cases tags themselves are translated, i.e. in Dutch:
_"Possible missing traffic_signals:direction tag or crossing on traffic signals"_
Became:
_"Mogelijk ontbreekt traffic_signals:richting tag of kriusing op verkeerslichten"_
(i.e. `traffic_signals:direction` & `crossing` were translated, but shouldn't as they're tag names)

Hence I was thinking: maybe it's an idea to convert all these from (using this example)
```python3
title = T_('Possible missing traffic_signals:direction tag or crossing on traffic signals')
```
To for instance:
```python3
title = T_('Possible missing `traffic_signals:direction` tag or `crossing` on traffic signals')
or
title = T_('Possible missing {0} tag or {1} on traffic signals', 'traffic_signals:direction', 'crossing')
or
title = T_('Possible missing `{0}` tag or `{1}` on traffic signals', 'traffic_signals:direction', 'crossing')
```

Obviously a lot of work and I hope this wouldn't erase good translations?

frodrigo:
Yes, the good practice for now is using backquote. As it may be rendered as markdown.

Famlam:
Would you say it's best to update all the strings now, or only when they're modified in a PR that modifies the code? Or only the ones that are currently incorrectly translated in a language?
(Updating the strings would also cause a lot of the translations to get lost I think)

frodrigo:
You could may a mass update PR. If you want.

If it impacts too much strings, we can try to update these with a script.

jocelynj:
To keep current translations that are on transifex, the script should also update the strings in po/*.po files, so that they can be directly uploaded to transifex.

PR #1706, opened 2022-12-22
Add missing `

Famlam:
Adds `` around text that directly refers to a tag/key/value.
Only did it for `detail/trap/fix` messages, given that `title` messages aren't formatted and there it would thus be of less effect.
(If you would want it on `title` too, especially analyser_osmosis_relation_public_transport and analyser_osmosis_highway_traffic_signal need an update)

@jocelynj Perhaps you can re-run your script of yesterday on these too to update the .po files? :)

#1333


frodrigo:
Humm, maybe I merged too quickly.

Famlam:
Does Transifex automatically extract from Github? If not it shouldn't hurt to run the script afterwards?

jocelynj:
Transifex doesn't download from github: its update is manual.

I've launched the script and updated the translations on commit 6467501d4991bcbf40922d1cb6791313aef21acb.

Issue #1672, opened 2022-12-05
Parking plugin revision


Famlam:
It seems that the parking proposal will be accepted on Dec 8; in that case a big revision of the parking plugin is required.
https://wiki.openstreetmap.org/wiki/Proposed_features/street_parking_revision

Perhaps it would be good to disable this plugin for a couple of days between acceptance of the proposal and updating the plugin, so we can await until the wiki is fully up to date again? @frodrigo ?

Famlam:
Nevermind, it seems the plugin will not error with the new syntax, it nicely slips through.
So we can just wait until the new syntax has some usage

Discostu36:
I think that now as the proposal has been accepted and the tags are used, Osmose should highlight the old tags as deprecated.

frodrigo:
Osmose fetch deprecation from this page, it is automatic.

https://wiki.openstreetmap.org/wiki/Deprecated_features

Famlam:
>Osmose fetch deprecation from this page, it is automatic.

Just curious, will it also fetch these ones? `parking:lane` itself isn't a common tag, only `parking:lane:left/right/both` (same for condition)

frodrigo:
> > Osmose fetch deprecation from this page, it is automatic.
> 
> Just curious, will it also fetch these ones? `parking:lane` itself isn't a common tag, only `parking:lane:left/right/both` (same for condition)

Just `parking:lane=*` as on the wiki.

Famlam:
Great!

Discostu36:
The new tagging scheme already has thousands of uses. Could we re-open this for revision of the parking plugin?

frodrigo:
Result here
http://osmose.openstreetmap.fr/en/issues/open?item=3161&class=31621,31622,31623,31624,31625,31626,31627,31628,31629

Discostu36:
You didn't ask for my opinion, but I'll be so frank to give it nonetheless: I think it should be whitelisted for the reasons that you have given.

frodrigo:
I did not really know this subject. So, as you say.

Famlam:
Hmm, on second thought and a bit more research: there's only 59 such conflicting items (`highway=*` + `amenity=parking` + `parking:[unknownside]`) in Europe (and Europe has the highest numbers of parking:fee, parking:access, ...) so I think I'll leave it for now. The warning from TagRemove_Incompatible has a higher priority anyway.

PR #1710, opened 2022-12-26
Implement new street parking scheme

Famlam:
See #1672

After the acceptance of the proposal, the new parking scheme is rising fast in usage (e.g.: [TagInfo for `parking:both`](https://taginfo.openstreetmap.org/keys/parking%3Aboth#chronology) )

- Preserves old tagging checks (for now) as it's still more common. (Only change is that it's indented by one tab). The fix suggestions have been changed to a 'consider switching scheme'-message
- Reimplements all previous checks with the new scheme
- Includes all subkeys in the 'equal value in `parking:both` and `parking:[side]`'-check, and in the '`parking:*:*` without `parking:*`'-check, such as `parking:both:maxstay`
- Checks for values in the wrong key for `parking:*:restriction`, `parking:*:direction` and `parking:*:orientation`

### DIFF
--- a/plugins/Highway_Parking_Lane.py
+++ b/plugins/Highway_Parking_Lane.py
@@ -90,7 +90,7 @@ be tagged on that object instead.'''),
         self.errors[31621] = self.def_class(item = 3161, level = 2, tags = ['highway', 'parking', 'fix:imagery'],
             title = T_('Bad parking:[side]'),
             detail = T_(
-'''The side was not recognized, expected was either of `left`, `right` or `both`.'''),
+'''The side was not recognized, expected was either `left`, `right` or `both`.'''),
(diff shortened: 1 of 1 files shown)


Famlam:
Note for reviewing: lines 150-192 haven't changed (except for an extra tab as they're in an `if`-block now. Unfortunately Github doesn't display this as such.

frodrigo:
> Note for reviewing: lines 150-192 haven't changed (except for an extra tab as they're in an if-block now). Unfortunately Github doesn't display this as such.

The an option to ignore whitespace change.


Thank you keeping up to date this analyser.

Famlam:
> The an option to ignore whitespace change.

Thank you! I didn't know that option yet

Issue #1646, opened 2022-11-11
Detect too small natural landuse=wood|forest


frodrigo:
Too small natural landuse=wood|forest are rather natural=tree or other thing.

![image](https://user-images.githubusercontent.com/1785486/201343891-0418c558-653d-4e82-ac95-b8a511c09548.png)
https://www.openstreetmap.org/way/129880985/history

Famlam:
Seems we also got #684 fixed with this (well, the non-dubious cases ;) )

frodrigo:
Not sure why we got no results 
http://osmose.openstreetmap.fr/en/issues/open?item=3100&useDevItem=all

frodrigo:
> Not sure why we got no results http://osmose.openstreetmap.fr/en/issues/open?item=3100&useDevItem=all

http://osmose.openstreetmap.fr/en/issues/open?item=3100&useDevItem=true

useDevItem=all looks broken

Famlam:
I think it is safe to make this item public, based on the results I see.

frodrigo:
Thank. Enabled.

PR #1718, opened 2022-12-29
Add analyser: too small polygon

Famlam:
Implements #1646

Detects small natural=wood / landuse=forest. The current cutoff of 20m2 is arbitrary, might need fine-tuning.
Implemented this way to allow addition of other objects in the future without needing new SQL queries. For example, I can imagine [`landuse=village_green`](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dvillage_green#Incorrect_use) to be added at some point.

Test case included. Results for 4 countries [attached](https://github.com/osm-fr/osmose-backend/files/10320542/output.zip), some have quite large amount of results.

The biggest caveat I can predict is that technically a single tree could also be considered to fulfill the requirement of "tree-covered area".


frodrigo:
Note. There is already a small farm analyser. Maybe interesting to merge the analysers.

Famlam:
> Note. There is already a small farm analyser. Maybe interesting to merge the analysers.

Ah, hadn't spotted this one. Funny that the code is so similar nevertheless. Any reason why it is in the 'disabled' folder?

frodrigo:
> > Note. There is already a small farm analyser. Maybe interesting to merge the analysers.
> 
> Ah, hadn't spotted this one. Funny that the code is so similar nevertheless. Any reason why it is in the 'disabled' folder?

Not, sure, I think it is because of too many false -positives, and it was in 2015.

I just found this in the history : https://github.com/osm-fr/osmose-backend/commit/9c266fd6fae22e56c8eb09978278ee4a61d7137a#diff-130af7a1da77e75cbdf5aadddf22a7f7f241f342549a8b83f0fca8afd3c2d066R626

Famlam:
I'll play around a bit with the size, to see when it's no longer returning false positives. Besides Cuba, it was previously disabled in NL too, and I happen to be familiar there ;). Later this week...
https://github.com/osm-fr/osmose-backend/commit/e351d4d5bfc5ba939d1479de1dacac0d5bf1624c

Famlam:
I've added "the good parts" of the mini_farm analyser, as well as checks for `landuse=farmland` (not `landuse=farm`, as that's deprecated) and `landuse=village_green`.

The original mini_farm analyser used 5000m2 as a cutoff. This did cause a lot of issues.
From about 800m2 the results start to become ok, with still many false positives though. At 600m2 it's really ok.
I currently picked 120m2 as that seems to be where there's almost no false positives anymore. Nonetheless, there's also a lot of detections missed then.

 **Please pick your preference:**
1. set it to 800 and have many false positives, but also many valid detections
2. set it to 600 and have some false positives, but also many valid detections
3. keep it at 120 and have almost no false positives, but also just a few valid detections

I do tend towards 600. 

Also we might have to at some point add a filter for countries where landuse is imported automatically from a gouvernement source, such as Slovenia.

frodrigo:
Maybe there is secondary tags that can help on avoiding false positives ?

Famlam:
> Maybe there is secondary tags that can help on avoiding false positives ?

If I run the analyser with 600m2 for farmland, I get (using `--country=cuba --country=netherlands_gelderland --country=usa_iowa --country=slovenia --country=france_nord_pas_de_calais_nord --country=jersey`):

85549x `source=*`
85183x `source:date=*`
85240x import-specific tags (`raba:id=*`)
83694x `landuse=farmland`
2416x `landuse=forest`
140x `natural=wood`
99x various other tags

Hence, filtering on sources seems to be the only viable solution for small farmlands.

-----

When limiting `landuse=farmland` to 120m2, I still get 10510 hits for that tag. Note that the majority of those is from Slovenia, it could appear at the end that that's just a very special country in this respect. But this we would see in the "useDevItem" phase.

-----

In terms of size distribution (area vs. occurrence grouped by 10m2), there's quite a big step around 100, but besides that there seems to be no clear stepping point for small farms:
image

If I compare the overall farm size distribution in one country (netherlands_gelderland), there also doesn't seem to be another clear sweet spot, it's pretty equally distributed; in fact, there's not even a clear lowest size (first graph: grouped by 500m2, second: grouped by 25m2). Hence, the cutoff will always be somewhat arbitrary, as the tiniest ones are clear errors.
image
image

For a different country:
image
image

-----

I think it's fair to say that this analyser will never be without false positives. It is just a good first filter to find likely mistakes though, as long as we ensure that the number of items that have to be marked as false positives by the user is so low that it's not considered as 'flooding the map with balloons'

frodrigo:
Good, thank you.

Excluding Slovenia or other areas can be done.

frodrigo:
Do you think we can compute automatically the step number from the current extract (and use default value for small extracts) ?

Famlam:
> Do you think we can compute automatically the step number from the current extract (and use default value for small extracts) ?

Do you mean: calculating the average size of a farm per country and using that to get to a "minimum area" value? I think this would be fairly error prone as countries like the US (much open area) will probably have some very big farmlands, but those do not automatically make smaller farmlands "bad". 
I think the only thing we can do is consider what could be a viable farmland size. Using potatoes as an example, 1m2 is definitely not viable (that's just enough potatoes for your own family per year I guess - I read 3.5kg potatoes/m2 on a random [site](https://makkelijkemoestuin.nl/en/knowledge-base/growing-potatoes)). No chance that someone will commercially sell a single bag of potatoes per month, but starting around 100m2 (or 350kg potatoes) I guess it can be enough to sell in a small village market. (Of course, a single farm can have multiple farmlands, for different crops)

frodrigo:
I was more think of computing the big step we can see on your first graph. So we does not have to disable area like Slovenia.
But you probably right, same size for all area could be.

Famlam:
I do fear that this big step may be Slovenia-specific though. There is no "natural" reason for such a steep step that I can think of (unless law forbids farms smaller than that size)

frodrigo:
Ok, I misunderstanding the first graph.
So, yes, all static values. 
But we can add a condition to have a different value for Slovenia.

Famlam:
It's easier to just lower the number from 120 to 100 then ;).

frodrigo:
Can you reuse the old famrland item 3100, please?

Famlam:
Sure, this does not affect it from being under useDevItem first, right?

Famlam:
By the way, I got another idea to identify many bad farmlands. Will provide the patch later in a separate PR (different analyser)

PR #1719, opened 2022-12-29
Ignore `dvalue_nolink=1`, require exact matches

Famlam:
The entry for `highway=incline_steep` is now:
```
{{Deprecated features/item|lang={{{lang|}}}|date=2008-06-12
|dkey=highway|dvalue=incline_steep|dvalue_nolink=1
|suggestion={{Tag|incline}}
|3}}
```
(See: [wiki edit history](https://wiki.openstreetmap.org/w/index.php?title=Template%3ADeprecated_features&type=revision&diff=2456444&oldid=2455457) )
Hence I suspect the assertionErrors originate from `dvalue_nolink` being used instead of `dvalue`.


frodrigo:
Right. Thank for fixing this.

PR #1723, opened 2023-01-01
Warn about farmland/farmyard mixup

Famlam:
Finds farmlands with significant buildings inside itself, which are very likely farmyards instead. This way, many bigger bad farmlands are still found even if the polygon_small analyser (https://github.com/osm-fr/osmose-backend/pull/1718) has a low treshold

Depending on which description belongs to the group of 3100, the item (and class) may have to be changed to that number.
I named it building_in_polygon so its easy to add more similar checks to the same file; can rename it to building_in_farmland if desired though.

### DIFF
--- a/analysers/analyser_osmosis_building_in_polygon.py
+++ b/analysers/analyser_osmosis_building_in_polygon.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Osmose Project 2023                                        ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from modules.OsmoseTranslation import T_
+from .Analyser_Osmosis import Analyser_Osmosis
+
+sql10 = """
+SELECT
+  ways.id,
+  buildings.id,
+  -- Use the building location to point out the error, even though the landuse should be changed
+  ST_AsText(way_locate(buildings.linestring)),
+  buildings.tags->'building'
+FROM
+  buildings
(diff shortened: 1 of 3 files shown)


analysers/analyser_osmosis_building_in_polygon.py
```diff
@@ -0,0 +1,91 @@
...
+  ST_AsText(way_locate(buildings.linestring)),
+  buildings.tags->'building'
+FROM
+  buildings
+  JOIN ways ON
+    ways.is_polygon AND
+    ways.tags != ''::hstore AND
+    ways.tags?'landuse' AND
+    ways.tags->'landuse' = 'farmland' AND
+    ST_Contains(ST_Transform(ST_MakePolygon(ways.linestring), {proj}), buildings.polygon_proj)
```
frodrigo, line 40: Add condition on un-projeted geom to use index
```
way.linestring && building.linestring
```

frodrigo:
Good idea! 

Maybe we can also check for vineyard and orchard.

Large greenhose should be on `landuse=greenhouse_horticulture`.

Famlam:
All done, thanks!
Not entirely sure about a single greenhouse on an area dedicated to farmland in general (e.g. mixed greenhouses and non-greenhouses). I thought `greenhouse_horticulture` was only for dedicated greenhouse landspace. But I removed the exception for now, we can always re-add it if there's many issues with this. It can only lead to more detailed mapping I guess, just like `vineyard` is more detailed than `farmland`.

frodrigo:
Thank, merged.

Famlam:
Maybe good that I never check whether the id of the landuse *equals* the id of the building. Someone in the Phillipines tags all rice fields as building=house + landuse=farmland 🙃 
https://www.openstreetmap.org/way/1100793688#map=17/15.32115/119.99650

Other than that, the results look good to me (safe to be released to the general public). Do you want to wait for https://github.com/osm-fr/osmose-backend/pull/1724 to be rolled out? If not, for me it would be fine to remove the useDevItem flag

frodrigo:
It is released now.

Famlam:
Based on the results I checked, I think it can be released :)
(i.e. without useDevItem)

frodrigo:
As I say, It is already released.

Famlam:
That's odd, because it is empty here: https://osmose.openstreetmap.fr/nl/issues/open?item=1310 , while https://osmose.openstreetmap.fr/nl/issues/open?item=1310&useDevItem=true shows all results. 

(I thought your first message meant that the patch of #1724 was released)

frodrigo:
I was mistaken with item=3100.

Ok, but for now we cannot enable it. The main frontend host server crashed and we are on a backup.

Famlam:
Ah, sorry to hear that! Shall I remind in - say - two weeks or so? Or is more time needed?

frodrigo:
Ok, now it is public.

Famlam:
I guess you don't grow crops on said roof? Then I would say that it, if it's really in all 360 degrees surrounded by crop-growing land (=farmland) could become a multipolygon yes. 

Although I guess just making the building itself the "inner" would also be fine if the crops reach up to the walls

PR #1730, opened 2023-01-07
Release constraint on upper case Brand, NSI not reliable on Upper/Lower case

frodrigo:
Name_UpperCase plugin raise false positives because it only consider brand in uppercase on NSI.

But NSI is not reliable on upper/lower case.

So, treat NSI without respecting case.

cc @malteger

### DIFF
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -41,11 +41,14 @@ class Name_UpperCase(Plugin):
             detail = T_(
 '''This feature is tagged with a name which contains a fully uppercase word (or words).
  This is not expected for the majority of named features.'''),
+            fix = T_(
+'''acronyms must be set in `short_name` tag.'''),
             trap = T_(
 '''While uncommon, it is possible for a name to have uppercase words.
  This is particularly the case for corporate/branded locations as well as acronyms.''')
         )
-        self.UpperTitleCase = re.compile(r".*[\p{Lu}\p{Lt}]{5,}")
+        self.Letter = re.compile(r"^\p{Letter}{5,}$")
+        self.UpperTitleCase = re.compile(r".*[\p{Uppercase_Letter}\p{Titlecase_Letter}]{5,}")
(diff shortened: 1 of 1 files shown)


plugins/Name_UpperCase.py
```diff
@@ -41,11 +41,14 @@ def init(self, logger):
             detail = T_(
 '''This feature is tagged with a name which contains a fully uppercase word (or words).
  This is not expected for the majority of named features.'''),
+            fix = T_(
+'''acronyms must be set in `short_name` tag.'''),
```
Famlam, line 45: I think acronyms should start with a capital `A`?

Also an optional suggestion to make it a bit more specific:
"Acronyms should be set in the `short_name` tag, unless the full name is never used in reality except for historic purposes".
As an example, a well-known general purpose store in The Netherlands is HEMA. Officially this means Hollandsche Eenheidsprijzen Maatschappij Amsterdam, but no-one in The Netherlands knows this full name (I had to look it up on Wikipedia as well to write this example); the former acronym is now the real name.

malteger:
To me this PR makes the check useless, as every word from every brand which might trigger the check (i.e. is 5 characters or more) is added to the whitelist. I just did some checking, and at least for German there are also generic words added (and the whitelist increases from 38 to 1756 entries). The result would be that no words which appear in brands will be reported by this check, even if everyone would agree that the uppercase version is clearly wrong.

In my understanding the NSI has the official name for brands in the name field to be used in the name field in OSM, and for me that includes the case. But after reading https://github.com/osmlab/name-suggestion-index/issues/5510 I think that the NSI is probably a bad choice for populating the whitelist in retrospect.

Famlam:
Hmm, good point. If it's that bad with capitalization, then I am also concerned that TagFix_Brand is a problem as it would propose to add incorrect entries.

@frodrigo do you have like statistics of how often the letter casing is wrong in NSI? If it is not too often, it may be better to fix NSI

frodrigo:
Ok, right. It convince me also.

Issue #1733, opened 2023-01-13
waste_disposal capacity in m3 should be a valid decimal value


carlosarc87:
Invalid capacity value
"capacity" value X is not an integer

item: 3091
class: 30912

[Tag:amenity=waste_disposal](https://wiki.openstreetmap.org/wiki/Tag:amenity=waste_disposal): 
[capacity](https://wiki.openstreetmap.org/wiki/Key:capacity)=* - to specify the capacity in m³.

Examples:
http://osmose.openstreetmap.fr/es/issue/a812063f-771a-38c7-4a4f-5fd1b3f4986a
http://osmose.openstreetmap.fr/es/issues/open?source=425432&item=3091&class=30912

Regards

Famlam:
Indeed. It should be excluded in https://github.com/osm-fr/osmose-backend/blob/master/plugins/Capacity.py#L53 as it can be non-integer

carlosarc87:
Thank you so much, it seems to be working.
I am sorry, but I forgot to include amenity=recycling, that has the same problem.
I hope you can add it when it is possible.
Regards

Famlam:
Hmm, `capacity` is used 7k times with that key, but not documented on the wiki. 
Could you update the wiki please? https://wiki.openstreetmap.org/wiki/Tag:amenity%3Drecycling#Useful_combinations

carlosarc87:
The wiki was updated by other user a few minutes ago, regards

Famlam:
Thanks! I'll let it sit for a couple of days in case someone disagrees with the wiki change and then update the code :)

Famlam:
Not sure what I messed up in the first two attempts, but third PR seems to be ok :). Sorry for the PR-spam :)

carlosarc87:
I have checked that it is fixed, thank you so much.
Regards

PR #1734, opened 2023-01-13
Exclude amenity=waste_disposal from integer capacity value check

Famlam:
The combination is documented since 2017, and used about 10.6k times
See #1733


Famlam:
(The test failure seems unrelated to this PR)

PR #1737, opened 2023-01-14
Update URLs of TagWatchFrViPofm.py

### LINKED ISSUE #1736: Update or replace TagwatchCleaner
The wiki page used in https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagWatchFrViPofm.py has been moved:
https://wiki.openstreetmap.org/w/index.php?title=User:FrViPofm/TagwatchCleaner&action=history
(Hence the tests are failing)

Is this list still being used? (And if so, isn't it better to have it as mapcss file, as far as it's not yet covered by i.e. JOSM?)

Famlam:
1. Update the URLs
2. Get the URL out of the translation in case it needs to be updated again

Untested, just assuming the URL change is all that needs to be done 0:)
Fixes #1736


frodrigo:
Thank you.

PR #1738, opened 2023-01-14
Check item_map for removed classes

Famlam:
- Update mapcss files; blacklisted one that is also covered by another test
- Add a test to find which mapcss rules have disappeared (meaning these need to be cleared at the frontend). Only works for hardcoded items. Also added a test for missing 'hardcoded classes', but I didn't do this for now.


frodrigo:
Merged manually. I merged the two last commits.
Thank you.

PR #1740, opened 2023-01-16
mapcss parsing: string with int comparison triggers RuleAbort

Famlam:
Have this as an example (minimal) mapcss rule:
```css
*[x!=0] {
  assertMatch: "node x=\"0.5 osmose\"";
  throwWarning: "x";
}
```

Compile the python code and run the test. The match fails due to `modules.mapcss_lib.RuleAbort` being raised by `to_n` - called by `__ne__` - as it tries to convert the nonconvertible string `0.5 osmose` to a float:
https://github.com/osm-fr/osmose-backend/blob/fe158f2c45eac2d0a210cd22a8d0f0d9af68087e/mapcss/mapcss_lib.py#L158-L165
(The error is caught by the `try`/`except` in the mapcss2osmose-generated .py file)

For comparison: JOSM only uses numerical comparisons if the types match: https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/gui/mappaint/mapcss/Functions.java#L722

Unfortunately there was no assertMatch in any mapcss file that revealed this behaviour. I've added one to the Dutch rules now (as this one contains a rule that did trigger the error, thereby causing false negatives)

p.s.: please note that the mapcss update was necessary as there's also a difference in how JOSM and Osmose treat `[/regex/!~/regex/]` rules, and fixing this bug revealed that difference. Since this is undocumented behaviour, I decided to just modify the Dutch mapcss so that both give the same result. Unfortunately I made a little mistake there, hence the double mapcss update.

### DIFF
--- a/plugins/tests/test_mapcss_parsing_evalutation.validator.mapcss
+++ b/plugins/tests/test_mapcss_parsing_evalutation.validator.mapcss
@@ -1,0 +1,162 @@
+/* Disclaimer: 
+This document only serves to contain test cases to evaluate
+the parsing and evaluation of mapcss files, and is therefore
+not supposed to be used in JOSM for validation of OSM data */
+
+
+node[x=0] {
+  throwWarning: "test #1740";
+  assertMatch: "node x=0";
+  assertNoMatch: "node x=1";
+  assertNoMatch: "node x=1.0";
+  assertNoMatch: "node x=Osmose";
+  -osmoseTags: list("fix:survey");
+  -osmoseItemClassLevel: "4030/40301:0/2";
+}
+node[x!=0] {
+  throwWarning: "test #1740";
+  assertMatch: "node x=1";
+  assertMatch: "node x=1.0";
+  assertMatch: "node x=Osmose";
+  assertNoMatch: "node x=0";
+}
+
+
+way[x~=C1] {
+  throwWarning: "test #1610";
+  assertMatch: "way x=C1";
+  assertMatch: "way x=C00;C1;C22";
+  assertNoMatch: "way x=C12";
+}
+
+
+way:righthandtraffic[x=y][z?] {
+  throwWarning: tr("test #1603 - {0}{1}", "{1.tag}", "{2.tag}");
+  -osmoseAssertMatchWithContext:list("way x=y z=yes", "inside=NL");
+}
+way[x=y][z?]:righthandtraffic {
+  throwWarning: tr("test #1603 - {0}{1}", "{0.tag}", "{1.tag}");
+  -osmoseAssertMatchWithContext:list("way x=y z=yes", "inside=NL");
+}
+
+
+*[parking][amenity!~/^(parking|motorcycle_parking)$/] {
+  throwWarning: tr("test #994 - {0}{1}", "{0.key}", "{1.tag}");
+  assertMatch: "node parking=yes amenity=osmose";
+  assertNoMatch: "node parking=yes amenity=parking";
+}
+
+
+*[a][!c],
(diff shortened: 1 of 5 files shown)


jocelynj:
It would be nice to add unit tests to mapcss_lib.py to catch this kind of issue. Is is possible?

Famlam:
I kind-of figured that having an assert in "my" Dutch rules now also counts as a unit test, but yeah, maybe good to make a fake mapcss file with difficult cases (as the Dutch mapcss file may change in the future). Will do so once this patch lands, if that's fine with you?

frodrigo:
> hence the double mapcss update.

Note, you can edit history in local and push force the branch to change the history.

frodrigo:
> I kind-of figured that having an assert in "my" Dutch rules now also counts as a unit test, but yeah, maybe good to make a fake mapcss file with difficult cases (as the Dutch mapcss file may change in the future). Will do so once this patch lands, if that's fine with you?

May be just add test to `mapcss_lib.py`, for now, there is no test on it :-/.

Famlam:
> > hence the double mapcss update.
> 
> Note, you can edit history in local and push force the branch to change the history.

Will try, one second

Famlam:
> > I kind-of figured that having an assert in "my" Dutch rules now also counts as a unit test, but yeah, maybe good to make a fake mapcss file with difficult cases (as the Dutch mapcss file may change in the future). Will do so once this patch lands, if that's fine with you?
> 
> May be just add test to `mapcss_lib.py`, for now, there is no test on it :-/.

Ok. Probably on Friday or Saturday though, bit busy coming days so then this PR has to wait, sorry 0:)
Note that I also added `assertMatch:"way highway=residential traffic_sign=\"NL:C02\" oneway:bicycle=no oneway=no"` to the Dutch rules which would've triggered this issue prior to the patch, due to `oneway=no` (from assertMatch) vs `[oneway!=-1]` (from the rules), so there actually is a test on it now.

Do you mind if I create a generic test for mapcss rules? Even though some tests are "off-topic" to this PR? Because if we start testing, I'd prefer to also add tests for other "previous" issues of the mapcss parsing.

frodrigo:
> Do you mind if I create a generic test for mapcss rules? Even though some tests are "off-topic" to this PR? Because if we start testing, I'd prefer to also add tests for other "previous" issues of the mapcss parsing.

Sure, tests are very welcome!

Famlam:
Had some unexpected free time, so I implemented some tests :)
Figured this way made more sense than adding it to mapcss_lib, as this way the entire process is being tested, from compiling to parsing the rule to the specific functions involved. (And with few exceptions, almost everything can be tested with the asserts of mapcss). 
(Found the bug #1742 this way too; already added it to the file, but commented out for now until its fixed)

plugins/tests/test_mapcss_parsing_evalutation.validator.mapcss
```diff
@@ -0,0 +1,162 @@
+/* Disclaimer: 
+This document only serves to contain test cases to evaluate
+the parsing and evaluation of mapcss files, and is therefore
+not supposed to be used in JOSM for validation of OSM data */
+
+
+node[x=0] {
+  throwWarning: "test #1740";
+	assertMatch: "node x=0";
```
frodrigo, line 9: Can you fix the indentation of this file please.

Famlam: Oops, didn't notice it. Fixed now

frodrigo:
Nice. Thank you to do it.

PR #1741, opened 2023-01-17
Stationnement vélo Strasbourg

le-jun:
Hi, making my first steps onto osmose backend with bicycle parking data provided by Strasbourg. Adapted the analyser file from the Paris one (which itself is based upon the Bordeaux one).

I’d like to bring some attention on the fact that provided data are under the [Base Nationale du Stationnement Cyclable](https://transport.data.gouv.fr/datasets/stationnements-cyclables-issus-dopenstreetmap) which aims to ease the conversion between OSM and local data. I didn’t find any ready-made template to adapt from which is why I tried to do so myself, left as many fields as possible (before commenting them out) so that may become a template for future uses.

### DIFF
--- a/analysers/merge_bicycle_parking_FR_strasbourg.py
+++ b/analysers/merge_bicycle_parking_FR_strasbourg.py
@@ -64,9 +64,7 @@ tag_mapping1 = {
 }
 
 tag_mapping2 = {
-    "ref": (
-        lambda res["id_local"]
-    ),
+    "ref": "id_local",
 #    "access": (
 #        acces      
 #    ),
@@ -82,19 +80,11 @@ tag_mapping2 = {
     "lit": (
         lambda res: "yes" if res["lumiere"] == "vrai" else None
     ),
-    "website": (
-        lambda res["url_info"]
-    ),
-    "start_date": (
-        lambda res["d_service"]
-    ),
-    "owner": (
-        lambda res["proprietaire"]
-    ),
+    "website": "url_info",
+    "start_date": "d_service",
+    "owner": "proprietaire",
     "operator": (
         lambda res: "Eurométropole de Strasbourg" if res["gestionnaire"] == "eurometropole" else None
     ),
-    "note": (
-        lambda res["commentaires"]
-    )
+    "note": "commentaires"
 }
--- a/analysers/merge_bicycle_parking_FR_strasbourg.py
+++ b/analysers/merge_bicycle_parking_FR_strasbourg.py
@@ -23,21 +23,50 @@
 from modules.OsmoseTranslation import T_
 from .Analyser_Merge import Analyser_Merge_Point, SourceOpenDataSoft, CSV, Load_XY, Conflate, Select, Mapping
 
-tag_mapping = {
-    "ref": (
-        lambda res["id_local"]
-    ),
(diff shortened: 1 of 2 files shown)


analysers/merge_bicycle_parking_FR_strasbourg.py
```diff
@@ -0,0 +1,103 @@
...
+###########################################################################
+
+from modules.OsmoseTranslation import T_
+from .Analyser_Merge import Analyser_Merge_Point, SourceOpenDataSoft, CSV, Load_XY, Conflate, Select, Mapping
+
+STANDS_TYPES = {
+    "Arceau autre": "stands",
+    "Arceau St-Germain": "wide_stands",
+    "Epingle vélo": "stands",
+    "Sans": "floor",
```
frodrigo, line 30: Maybe `None` rather than "floor"?

analysers/merge_bicycle_parking_FR_strasbourg.py
```diff
@@ -0,0 +1,97 @@
...
+            ),
+            Conflate(
+                select = Select(
+                    types = ["nodes", "ways"],
+                    tags = {"amenity": "bicycle_parking"}),
+                conflationDistance = 20,
+                mapping = Mapping(
+                    static1 = {"amenity": "bicycle_parking"},
+                    static2 = {"source": self.source},
+                    mapping1 = tag_mapping )))
```
frodrigo, line 97: https://github.com/osm-fr/osmose-backend/blob/master/doc/4-Merge.md#mapping

I guess not all tags should go to `mapping1`. When there is change in mapping1 it popup again on Osmose-frontend. But ignored when tag change is on `mapping2`. The same for static1 and static2.

It is important to avoid flooding the contributors with boring issues.

le-jun: I get what you mean, I didn’t thought about it at all and simply copied what’s been done and added what seemed right. Which tags do you think would better fit into `mapping2`?

le-jun: Following further reading of the doc, I can’t see how any of the tags I put into `mapping1` could be some kind of "metadata". In my opinion they’re all necessary and any change made would mean some kind of important modification to be edited.

frodrigo: Personally, I will limit mapping1 to ref, capacity* (and bicycle_parking). It is already what other merge contents.

le-jun: Fine by me!

frodrigo:
Thank you.
Merged manually :
- squash commits
- fix code layout
- fix lat/lon string extract and order
- read file encoding UTF-8 with BOM
- fix source code filename

le-jun:
Hi, it’s been a week already and it seems the [changes still aren’t live](https://osmose.openstreetmap.fr/en/issues/open?item=8150). Is it normal or did we screwed up somewhere?

frodrigo:
Deploy is done manually. We does not make deploy since the merge of this PR.
We are all voluntary and some time it can take time.
cc @jocelynj

frodrigo:
It is now deployed.

PR #1744, opened 2023-01-19
Exclude amenity=recycling too

Famlam:
See https://github.com/osm-fr/osmose-backend/issues/1733#issuecomment-1384529624
Fixes the issue again :)


jocelynj:
@Famlam : just a remark: instead of doing a lot of merge from master to a pull-request, you can instead rebase on master. This would give a better git history.

Famlam:
Thanks! I don't really know what went wrong (it says "committed 3 months ago", while the branch is only a few hours old), but I closed this PR and created a clean one instead. Will try force pushing next time. (Too bad the web interface only allows merging from master)

PR #1749, opened 2023-01-24
Fix test due to Wiki update

Famlam:
[Someone removed `trafic_calming` from the wiki](https://wiki.openstreetmap.org/w/index.php?title=Tagging_Mistakes&type=revision&diff=2468534&oldid=2468507) (probably because it doesn't occur in OSM anymore, but there's no change reason given), hence replacing it by a similar item that does still exist, so the tests don't fail anymore.

p.s. the other failing test (Construction.py) I didn't fix yet: `not_correct_dates = ["22/01/2023",`, do you just mean "not expired dates"? Because the format seems to be equally incorrect as `"22/01/2012"` the line above


frodrigo:
Thank you. I also fixed the date issue.

PR #1751, opened 2023-01-30
Reset fixme false positives every year

Famlam:
It is impossible for a `fixme` tag to be a false positive, as the tag literally means something has to be fixed. However, still a large number of fixme issues are marked as false positives: https://osmose.openstreetmap.fr/nl/issues/false-positive?item=4061&source=&class=40610,40611&username=&bbox=

Being marked as such very likely means the "fixme-origin" wasn't resolved by the time someone reviewed it. Hence, similar to the ['finished construction' plugin](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Construction.py#L98), reset the false positives every year

### DIFF
--- a/plugins/TagRemove_Fixme.py
+++ b/plugins/TagRemove_Fixme.py
@@ -21,6 +21,7 @@
 
 from modules.OsmoseTranslation import T_
 from plugins.Plugin import Plugin
+from datetime import date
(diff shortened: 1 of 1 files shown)


frodrigo:
Nice idea!

PR #1756, opened 2023-02-06
Fix postcode plugin on extracts on partial country extracts

Famlam:
The TagFix_Postcode analyser did not provide any result for partial extracts if they had a different country ID. For example, the provinces of The Netherlands ("NL-GR" instead of "NL" etc). (Fixed in the first commit)

Since there's also a postcode rule in the Dutch mapcss file, I added that one to the blacklist, to avoid duplicate warnings. (Commit 2 + 3)


frodrigo:
Nice. Thank you.

Issue #1757, opened 2023-02-09
Validate `check_dates`, `source`, `note` and `fixme` related tags


Dimitar5555:
`check_dates`, `source`, `note` and `fixme` are prefix tags, not suffix tags.

Any instances of `*:check_dates`, `*:source`, `*:note` and `*:fixme` should be flagged (the proper tags are `check_date:*`, `source:*`, `note:*`, `fixme:*`)
`surface_survey` should be `check_date:surface`
`smoothness:date` should be `check_date:smoothness`

jajajaneeneenee:
A note: there is an ongoing proposal [Proposed features/:note suffix](https://wiki.openstreetmap.org/wiki/Proposed_features/:note_suffix) which argues to change `note` to a suffix with some reasons (some language suffixes for `note` can cause conflicts/ambiguities in some rare cases if it's used as a prefix).

And the German wiki page for `note` even recommends using it as a suffix rather than a prefix! And the [English wiki page for `note`](https://wiki.openstreetmap.org/wiki/Key:note) for example doesn't say a word about it (the same with `fixme`). Some kind of confusing ...

Another note: I saw the tag "parking:note" serveral times in my region which causes already an Osmose warning because it's not a valid/common suffix like left|right|both. Seems to make sense.

Dimitar5555:
> A note: there is an ongoing proposal [Proposed features/:note suffix](https://wiki.openstreetmap.org/wiki/Proposed_features/:note_suffix) which argues to change `note` to a suffix with some reasons (some language suffixes for `note` can cause conflicts/ambiguities in some rare cases if it's used as a prefix).

The "ongoing" proposal seems to be inactive for the past 7 months and most people who commented on it seem to disagree with it.

> And the German wiki page for `note` even recommends using it as a suffix rather than a prefix! And the [English wiki page for `note`](https://wiki.openstreetmap.org/wiki/Key:note) for example doesn't say a word about it (the same with `fixme`). Some kind of confusing ...

It looks like this is a mistake since the text says that is should be a prefix. There are about 2.1k `highway:note` in Germany which is why I've created a topic on their community forum (https://community.openstreetmap.org/t/highway-note-in-germany-highway-note-in-deutschland/8776)

jajajaneeneenee:
> The "ongoing" proposal seems to be inactive for the past 7 months and most people who commented on it seem to disagree with it.
 
[...]

PR #1758, opened 2023-02-09
More "fixme"s to be warned about

Famlam:
Inspired by #1757, include the fixme prefix and suffix in the 'fixme' check. (Also include cases like `fixme2`). The suffix and fixme2 may not be official schemes, but their meaning and purpose is clear enough IMO to warrant displaying their contents as a warning on the map.

Several thousands of cases worldwide: https://taginfo.openstreetmap.org/search?q=fixme

(Note: this does not fix 1757)


frodrigo:
Good idea. Thank you.

PR #1763, opened 2023-02-16
Warn about amenity on/in roof

### LINKED ISSUE #1761: amenity on rooftops
I initially [proposed it to JOSM](https://josm.openstreetmap.de/ticket/22556) and it's still under consideration of a throwOther rule, but throwOthers are ignored in Osmose. Hence I was wondering what you think about adding this rule to Osmoses own rules?

```css
*[building=roof][amenity] {
  throwWarning: tr("{0} together with {1}, usually {1} is located underneath the {2}", "{0.tag}", "{1.tag}", "{0.value}");
  assertMatch: "way building=roof amenity=fuel";
  -osmoseItemClassLevel: "--------to-be-determined, probably item 3032--------";
}
```

Rationale: amenities are (almost) never located **on** a roof, usually beneath. So the proper tagging would be:
`building=roof + layer=1` for the roof 
`amenity=*` on a new node (or if the shape is known, on an area) underneath the roof area

Famlam:
Fixes #1761

Add rule to warn about `amenity` (such as fuel stations) together with `building=roof`.
I excluded amenity=shelter as I can see how that could be a valid combination, if the shelter is literally a roof only. (Wouldn't be my preferred tagging though)

### DIFF
--- a/plugins/TagFix_MultipleTag2.py
+++ b/plugins/TagFix_MultipleTag2.py
@@ -1,0 +1,96 @@
+#-*- coding: utf-8 -*-
+import modules.mapcss_lib as mapcss
+import regex as re # noqa
+
+from plugins.Plugin import with_options # noqa
+from plugins.PluginMapCSS import PluginMapCSS
+
+
+class TagFix_MultipleTag2(PluginMapCSS):
+
+
+
+    def init(self, logger):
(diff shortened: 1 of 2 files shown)


frodrigo:
Thank

PR #1764, opened 2023-02-17
Initial commit of long crossing check

emersonveenstra:
This is a new analyser that looks for ways with `footway=crossing` or `cycleway=crossing` that are over 200 meters long. This happens when people change the tags without splitting the way over the actual crossing part. I haven't found any false positives in the equivalent overpass queries:
```
way["footway"="crossing"](if: length() > 200)({{bbox}});
(._;>;);
out;
```
```
way["cycleway"="crossing"](if: length() > 200)({{bbox}});
(._;>;);
out;
```
This isn't ready to be merged, I need a little guidance on what to make the item number and level, and what I should do for tests

### DIFF
--- a/analysers/analyser_osmosis_highway_long_crossing.py
+++ b/analysers/analyser_osmosis_highway_long_crossing.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyright Osmose Contributors 2023                                   ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from modules.OsmoseTranslation import T_
+from .Analyser_Osmosis import Analyser_Osmosis
+
+sql10 = """
+SELECT
+    id,
+    ST_AsText(way_locate(linestring))
+FROM
+    {0}ways
+WHERE
+    tags != ''::hstore AND
+    (
+        (tags?'footway' AND tags->'footway' = 'crossing') OR
+        (tags?'cycleway' AND tags->'cycleway' = 'crossing')
+    ) AND
+    ST_Length(linestring::geography) > 200
+"""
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_highway_long_crossing.py
```diff
@@ -0,0 +1,56 @@
...
+
+from modules.OsmoseTranslation import T_
+from .Analyser_Osmosis import Analyser_Osmosis
+
+sql10 = """
+SELECT
+    id,
+    ST_AsText(way_locate(linestring))
+FROM
+    {0}ways
```
Famlam, line 31: I think this is limited to highways, right? Then table `{0}highways` is probably more efficient?

Note you'll have to import `highways` and `touched_highways`, see for example:
https://github.com/osm-fr/osmose-backend/blob/9fa8a7d195db0425871072b3a9c8b52ecbdadc2b/analysers/analyser_osmosis_highway_bad_intersection.py#L88-L89

Famlam:
Nice idea :)
Maybe also interesting to include a waterway crossing ([ford=yes or ford=stepping_stones](https://wiki.openstreetmap.org/wiki/Key:ford)). I recently found one nearby my place :)

---

> I need a little guidance on what to make the item number and level, and what I should do for tests

Regarding the tests, you can maybe use 
https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_building_in_polygon.py#L80-L97
where you replace "osmosis_building_in_polygon" by the analyser name. (Note that I used proj=23032 there, which corresponds to i.e. [NL and Italy](https://github.com/osm-fr/osmose-backend/blob/master/osmose_config.py#L1546))

Then using for example JOSM you can make a .osm file with test cases, for example one `highway=footway + footway=crossing` of length 50m and one of 250m. Edit the .osm (=XML) file to add `timestamp='2014-03-31T22:00:00Z' version='1'` to every node/way, and to get positive ID numbers. Example file: https://github.com/osm-fr/osmose-backend/blob/master/tests/osmosis_building_in_polygon.osm

If the "bad" crossing would for example have ID 1234, then use `self.check_err(cl="CLASSNUMBER", elems=[("way", "1234")])`

analysers/analyser_osmosis_highway_long_crossing.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+class Analyser_Osmosis_Highway_Long_Crossing(Analyser_Osmosis):
+
+    def __init__(self, config, logger = None):
+        Analyser_Osmosis.__init__(self, config, logger)
+        self.classs_change[1] = self.def_class(item = 7800, level = 2, tags = ['tag', 'highway', 'fix:survey'],
+            title = T_('Long Crossing'),
+            detail = T_(
+'''The crossing way is much longer than expected'''))
```
Famlam, line 48: Maybe also add a `fix` message, describing to split the way when the crossing is over?
Also, don't forget to add a '.' at the end of the `detail` message :)

frodrigo:
@emersonveenstra still interested to work on this?

emersonveenstra:
@frodrigo i probably won't have the time for a while, feel free to take it or give it to someone else

Famlam:
I might be able to take it over if you prefer, in a week or two

frodrigo:
Closed in favour of https://github.com/osm-fr/osmose-backend/pull/1992

Issue #1598, opened 2022-10-12
Osmose mapcss tags extension


Famlam:
Spin-off from [issue 1567](https://github.com/osm-fr/osmose-backend/issues/1567#issuecomment-1273849693)

Nice-to-have mapcss keys would be:
- `-osmoseDetails` (same as setting `details` in a regular python plugin)
- `-osmoseTrap` (same as setting `trap` in a regular python plugin)
- `-osmoseFix` (same as setting `fix` in a regular python plugin)
- `-osmoseResource` (same as setting `resource` in a regular python plugin)

- `-osmoseOnlyForCountry` (same as setting `only_for = [x]` in a regular python plugin, to limit the plugin to certain countries)
- `-osmoseNotForCountry` (same as setting `not_for = [x]` in a regular python plugin, to disable the plugin in certain countries)

frodrigo:
>     * `-osmoseOnlyForCountry` (same as setting `only_for = [x]` in a regular python plugin, to limit the plugin to certain countries)
> 
>     * `-osmoseNotForCountry` (same as setting `not_for = [x]` in a regular python plugin, to disable the plugin in certain countries)

This two can be done using this already existing mapccs code `[inside("FR")] `

Famlam:
> > ```
> > * `-osmoseOnlyForCountry` (same as setting `only_for = [x]` in a regular python plugin, to limit the plugin to certain countries)
> > 
> > * `-osmoseNotForCountry` (same as setting `not_for = [x]` in a regular python plugin, to disable the plugin in certain countries)
> > ```
> 
> This two can be done using this already existing mapccs code `[inside("FR")] `

That only works on a selector level

frodrigo:
> > > ```
> > > * `-osmoseOnlyForCountry` (same as setting `only_for = [x]` in a regular python plugin, to limit the plugin to certain countries)
> > > 
> > > * `-osmoseNotForCountry` (same as setting `not_for = [x]` in a regular python plugin, to disable the plugin in certain countries)
> > > ```
> > 
> > 
> > This two can be done using this already existing mapccs code `[inside("FR")] `
> 
> That only works on a selector level

But, we do not report county in doc https://github.com/osm-fr/osmose-backend/blob/master/doc/0-Index.md#classes-of-issues

Famlam:
I'll await #1782 to go live (to see if there's anything else to be adjusted), then add `-osmoseExample`. For some reason I missed that one.

PR #1765, opened 2023-02-17
Convert 'simple' TagFix_Multiple to mapcss

Famlam:
See https://github.com/osm-fr/osmose-backend/issues/1761#issuecomment-1434571780 : 

### Convert simple rules from TagFix_Multiple to mapcss rules ### 
(which can be loaded in JOSM and area probably also easier to maintain) 

Exceptions:
I did not touch the following simple ones:

- Because I kind-of disagree with the rule 😉 :
```py
        if tags.get("barrier") == "fence" and "fence_type" not in tags and "material" in tags:
            err.append({"class": 303210})
```

- Because it seems that we're missing a lot of exceptions, compared to [JOSM's very similar rule](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss#L496)
```py
if (tags.get("tracktype") or tags.get("lanes")) and not tags.get("highway") and not tags.get("disused:highway") and not tags.get("abandoned:highway") and not tags.get("construction:highway") and not tags.get("proposed:highway") and not tags.get("planned:highway") and not tags.get("leisure") == "track":
            err.append({"class": 20803})
```

- And a couple because the python code has `detail`, `fix` or `trap` text, which would be lost if I go to mapcss

Additionally I made the following modifications:
- dropped `buildingpart` from the fix suggestions as it's considered a tagging mistake on the wiki 
- removed tag `highway` from the `amenity=recycling` rules
- Slightly modified the text of the two indoor rules to make sure it's clear we're talking about key names, not words that should be translated

### DIFF
--- a/plugins/TagFix_MultipleTag.py
+++ b/plugins/TagFix_MultipleTag.py
@@ -105,9 +105,6 @@ For further detail, see [the wiki](https://wiki.openstreetmap.org/wiki/Key:acces
 This is almost never the case and more specific tags should be used instead. \
 For further detail, see [the wiki](https://wiki.openstreetmap.org/wiki/Key:access#Transport_mode_restrictions).'''))
 
-        self.errors[32302] = self.def_class(item = 3230, level = 2, tags = ['highway', 'fix:chair'],
-            title = T_('Suspicious name for a container'))
-
         self.driving_side_right = not (self.father.config.options.get("driving_side") == "left")
         self.driving_direction = "anticlockwise" if self.driving_side_right else "clockwise"
         name_parent = []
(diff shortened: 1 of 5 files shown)


frodrigo:
> And a couple because the python code has detail, fix or trap text, which would be lost if I go to mapcss

We may extend MapCSS to also contains this, like we do for test https://github.com/osm-fr/osmose-backend/blob/master/doc/2-PluginMapCSS.md#tests

Famlam:
> > And a couple because the python code has detail, fix or trap text, which would be lost if I go to mapcss
> 
> We may extend MapCSS to also contains this, like we do for test https://github.com/osm-fr/osmose-backend/blob/master/doc/2-PluginMapCSS.md#tests

Yes, see #1598 ;)

PR #1767, opened 2023-02-17
Remove duplicate error

Famlam:
The following rule from JOSM [relation.mapcss](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/relation.mapcss):
 ```css
relation[!type] {
  throwError: tr("relation without type");
  assertMatch: "relation name=Foo";
  assertNoMatch: "relation type=route name=Foo";
}
```

Is exactly duplicated by class 21102:
https://github.com/osm-fr/osmose-backend/blob/9fa8a7d195db0425871072b3a9c8b52ecbdadc2b/plugins/TagFix_MultipleTag.py#L226-L227

In principle, I could remove class 21102 (same level, same scope), but I find the description of Osmose clearer than that of JOSM:
https://github.com/osm-fr/osmose-backend/blob/9fa8a7d195db0425871072b3a9c8b52ecbdadc2b/plugins/TagFix_MultipleTag.py#L70-L73
Hence I chose for blacklisting the JOSM rule instead


frodrigo:
Ok, thank.

PR #1771, opened 2023-03-02
Exclude parking=rooftop

Famlam:
See https://github.com/osm-fr/osmose-backend/issues/1761#issuecomment-1437609869
Debatable case, but if the roof is nothing but parking deck I could understand the tagging.
Went for the easiest solution (i.e. not checking `amenity=parking`) since there's only 77 matches worldwide


frodrigo:
Agree, better to avoid false positives.

PR #1773, opened 2023-03-06
Incorrect indent for 2+nd `set` declaration

Famlam:
While working on issue 1766 (without success) I noticed that, when multiple `set` statements are used, every `set` except the first gets a wrong number of spaces. Using the following example rule:

```css
way {
  set a;
  set b;
  throwWarning: "test";
  assertMatch: "way";
}
```
the following python code was generated:
```py
        # way
        if True:
            match = False
            if not match:
                capture_tags = {}
                try: match = (())
                except mapcss.RuleAbort: pass
            if match:
                # set a
                # set b
                # throwWarning:"x"
                # assertMatch:"way"
                set_a = True
            set_b = True
                err.append({'class': 1, 'subclass': 1979148453, 'text': {'en': 'x'}})
```
(Note that `set_b = True` is incorrectly indented in the generated code)

There are no actual rules that presently use multiple `set`s, but since this was an easy fix, I figured I could just as well fix it :)


frodrigo:
Thank. Good catch!

PR #1782, opened 2023-03-11
Add extra info to 'probably only bottles, not all glass' warning

### LINKED ISSUE #1776: Clarification for 32301 (`Probably only for bottles, not any type of glass`)
For the class ID 32301 `Probably only for bottles, not any type of glass`:
- It should mention `glass material`, otherwise a non-native English speaker can interpret `type of glass` as `type of glass object`, to mean objects of any shape, without thinking about the material.
- It should point to the table entry for `recycling:glass` in the wiki article for [`amenity=recycling`](https://wiki.openstreetmap.org/wiki/Tag:amenity=recycling), because
  - most people do not know that different materials exist, so this warning can come out as weird / surprising;
  - to make an effective, well-informed survey, one needs to know the story explained there.

For example, someone recently proposed in the OSM Italy chat to dismiss all warnings in his town as false positives, because he could only think of soda-lime examples when he asked for info at the municipality. When he was pointed to the wiki's explanation, he said `I would have never found that reference`, which confirms that this warning is incomplete.

Famlam:
Fixes #1776


frodrigo:
Thank you

Issue #1766, opened 2023-02-17
Rule skipped if `set` used for different type (node/way/relation)


Famlam:
I tried to update the mapcss files in preparation for another PR, and saw the following line:
`# Skip selector using undeclared class TunnelCulvertWithoutWaterway`
in the py file for https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss

It came from this rule:
`# *[tunnel][!highway][!area:highway][!railway][!waterway][!piste:type][type!=tunnel][public_transport!=platform][route!=ferry][man_made!=pipeline][man_made!=goods_conveyor][man_made!=wildlife_crossing][man_made!=tunnel][power!=cable]!.TunnelCulvertWithoutWaterway`

`TunnelCulvertWithoutWaterway` is however declared, just not for `node`/`relation`:
```css
way[tunnel=culvert][man_made!=tunnel][!waterway] {
  throwWarning: tr("{0} without {1} or {2}", "{0.tag}", "{1.tag}", "{2.key}"); 
  group: tr("suspicious tag combination"); 
  set TunnelCulvertWithoutWaterway; 
}
```

As it is only set for `way` rules, the mapcss2osmose script now discards any `node`/`relation` rule that refers to it via `!.TunnelCulvertWithoutWaterway`

(The discarding was done in case it was declared in an "unparsable" rule)

Famlam:
In JOSM, the following rule matches everything, even though I have never declared any `abcdef`
```
*!.abcdef {
  throwWarning: "x";
}
```

So indeed, `.something` is always false (and `!.something` is always true) until you've encountered a `set something` line.

In fact, the current implementation in Osmose for unparsable rules (which JOSM doesn't have as they define the rules) has another related flaw, as it only checks if a `set` has been declared once. However, if you have (extreme example) the following:

```
node[x=y] {
  set abc; /* matches almost nothing, but is parsable by Osmose */
}
way > * {
  set abc; /* matches any node of a way, but can't be parsed by Osmose */
}
node:unconnected {
  set abc; /* matches any node not part of a way, but can't be parsed by Osmose */
}
node!.abc {
  throwWarning: "X";
}
```
Then Osmose will throw a warning for all nodes, except nodes with `x=y`, while JOSM won't do anything.

Probably the best is not to filter out rules that have an undeclared `set`, but only to filter out rules where the set was declared in a rule that could not be parsed.

frodrigo:
[...]

PR #1785, opened 2023-03-13
Convert 'simple' TagFix_Multiple to mapcss (part 2)

Famlam:
Continuation of #1765
Now including the rules that were previously impossible without losing details (as we didn't have `-osmoseDetail` and such back then)
Also disables 2 JOSM rules that duplicated class 1050

Rules I skipped were:
- Class [20803](https://github.com/Famlam/osmose-backend/blob/7d7ca8996886364880515d50e910b5bcb5ebfe94/plugins/TagFix_MultipleTag.py#L130-L131) (We are missing a lot of exceptions if compared to [JOSMs near-identical rule](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss#L500); maybe this rule should be deleted, what do you think?)
- Class 303210 (I disagree with the rule)
- Classes 1050, 32200 and 32201 because mapcss doesn't support multiline strings (and we shouldn't add that because otherwise people can't load the rules in JOSM anymore)
- Class 21101 as it's not possible to do this in mapcss

### DIFF
--- a/mapcss/item_map.py
+++ b/mapcss/item_map.py
@@ -156,6 +156,7 @@ item_map = \
                             '{0} on a relation without {1}': 9001003},
                   'item': 9001,
                   'prefix': 'Josm_',
+                  'subclass_blacklist': [643796350, 317760248],
                   'tags': ['tag'],
                   'url': 'https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss?format=txt',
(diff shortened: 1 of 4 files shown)


Famlam:
p.s. just thinking: in order to be able to update combinations.mapcss from JOSM before #1766 is fixed, we could also just add that rule (without the class) to this file. What do you think? (I'll await this PR to be merged first)

frodrigo:
Merged. Thank you.

Issue #1726, opened 2023-01-07
Landuse intersection with railways and major highway


frodrigo:
Add a new analyzer to detect intersection or "touch" between:
* railways or major highways (down to secondary or tertiary), ignoring bridges and tunnels,
* and, landuse like forest or farmland.

Famlam:
I like it :)
Note that i.e. landuse=grassland or landuse=railway or landuse=industrial (and probably several others) can overlap with railways, and the latter also with highways, so it should probably be a blacklist-based method.

PR #1788, opened 2023-03-16
Add analyser for intersecting ways for transport and incompatible landuses

Famlam:
Detects (major) highways/active railways/big permanent waterways/plane-carrying aeroways intersecting with landuses (including `natural`) that are incompatible. For example, a runway intersecting with a forest (which typically doesn't end well) or a motorway in the water (without bridge or tunnel).
See #1726

Ignores 'snapped' ways (in some countries this seems to be the common method: to snap a highway to the nearby landuse) and very minor overlaps (<1m, to avoid overloading the map with balloons)

Includes tests.

(Note that I disabled `ways.tags?'landuse'` as this would result in a very slow analysis for some countries as the SQL parser made a poor choise)

### DIFF
--- a/analysers/analyser_osmosis_polygon_intersects.py
+++ b/analysers/analyser_osmosis_polygon_intersects.py
@@ -76,61 +76,150 @@ WHERE
 """
 
 sql11 = """
+CREATE TEMP TABLE landusage AS
 SELECT
-  mobility_ways.id,
-  ways.id,
-  ST_AsText(ST_Centroid(ST_Intersection(ST_MakePolygon(ways.linestring), mobility_ways.linestring))),
-  mobility_ways.type,
-  CONCAT(mobility_ways.type, '=', mobility_ways.tags->mobility_ways.type),
+  id,
+  linestring,
+  ST_MakePolygon(linestring) AS poly,
   CASE
-    WHEN ways.tags?'landuse' THEN CONCAT('landuse=', ways.tags->'landuse')
-    WHEN ways.tags?'natural' THEN CONCAT('natural=', ways.tags->'natural')
-  END
+    WHEN tags?'landuse' THEN 'landuse'
+    WHEN tags?'natural' THEN 'natural'
+  END AS landusekey,
+  CASE
+    WHEN tags?'landuse' THEN tags->'landuse'
+    WHEN tags?'natural' THEN tags->'natural'
+  END AS landusevalue,
+  tags,
+  FALSE AS is_multipolygon
+FROM
+  ways
+WHERE
+  is_polygon AND
+  tags != ''::hstore AND
+  (tags?'natural' OR tags?'landuse') AND
+  -- Leave minipolygons for analyser osmosis_polygon_small
+  ST_Area(ST_MakePolygon(ST_Transform(linestring, {proj}))) >= 20
+UNION ALL
+SELECT
+  ways.id,
+  ways.linestring,
+  -- Better would be to parse inner and outer segments and create an SQL multipolygon
+  -- For now, just use a 0.0000000001m buffer zone around the linestring
+  ST_Buffer(ways.linestring, 0.0000000001, 'quad_segs=1 endcap=flat join=bevel') AS poly,
+  CASE
+    WHEN r.tags?'landuse' THEN 'landuse'
+    WHEN r.tags?'natural' THEN 'natural'
+  END AS landusekey,
+  CASE
+    WHEN r.tags?'landuse' THEN r.tags->'landuse'
+    WHEN r.tags?'natural' THEN r.tags->'natural'
+  END AS landusevalue,
+  r.tags,
+  TRUE AS is_multipolygon
+FROM
+  relation_members
+  LEFT JOIN relations AS r ON
+    relation_id = r.id
+  JOIN ways ON
+    member_role = 'outer' AND
+    member_id = ways.id AND
+    member_type = 'W'
+WHERE
+  r.tags->'type' = 'multipolygon' AND
+  (r.tags?'natural' OR r.tags?'landuse')
+"""
+
+sql12 = """
+CREATE INDEX idx_mobilityway_linestring ON mobility_ways USING GIST(linestring);
+CREATE INDEX idx_landusage_linestring ON landusage USING GIST(linestring);
+CREATE INDEX idx_landusage_poly ON landusage USING GIST(poly);
+"""
+
+sql13 = """
+CREATE TEMP TABLE intersections AS
+SELECT
+  mobility_ways.id AS mobility_way_id,
+  mobility_ways.linestring AS mobility_way_linestring,
+  mobility_ways.type AS mobility_way_type,
+  landusage.id AS landusage_way_id,
+  landusage.poly AS landusage_poly,
+  landusage.landusekey,
+  landusage.landusevalue,
+  CONCAT(mobility_ways.type, '=', mobility_ways.tags->mobility_ways.type) AS tag_way,
+  CONCAT(landusage.landusekey, '=', landusage.landusevalue) AS tag_polygon
 FROM
   mobility_ways
-  JOIN ways ON
-    ways.is_polygon AND
-    ways.tags != ''::hstore AND
+  JOIN landusage ON
     (
+      NOT landusage.tags?'layer' AND NOT mobility_ways.tags?'layer' OR
       (
-        --ways.tags?'landuse' (...) sometimes causes bad route
-        (
-          ways.tags?'landuse' AND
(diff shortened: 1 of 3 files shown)


Famlam:
[Some results](https://github.com/osm-fr/osmose-backend/files/10996205/Some.results.zip)

frodrigo:
Thank for doing this.

Do all the kind of objects in only one query make it complexe. Do you thing
make one query by kind could make it more simple and maintenable?

I not checked yet, but I guess better perf may be achieved using index.

I think multipolygone could be supported, at least only the boundary.

Famlam:
Thank you!

> Thank for doing this. Do all the kind of objects in only one query make it complexe. Do you thing make one query by kind could make it more simple and maintenable? 

I'll split out the gathering of the landuses/natural (and soon multipolygons). Thanks!

> I not checked yet, but I guess better perf may be achieved using index. 

It's currently using the index on `linestring` in sql11 (sql10 is very fast already). Not sure which index to add?

Famlam:
@frodrigo I might need some advice here...

The current implementation for multipolygons puts a small buffer around every outer way. This is pretty simple, but for some reason, `ST_Touches` does not work anymore (also not if I use `side=left`/`side=right` depending on `ST_IsPolygonCW`) rather than the current implementation). Hence, cases like 
https://www.openstreetmap.org/way/332922967#map=18/49.23106/-2.15475 vs. https://www.openstreetmap.org/way/167981432 pop up

Additionally, there are cases like
https://www.openstreetmap.org/way/230084881#map=19/49.24584/-2.13132
where a few nodes do not connect (node 1854163936, 60964703) causing `ST_Touches` to fail and the way is too 'wobbly' to have my `ST_Distance` requirement filter it out

The second issue I can live with; would possibly warrant a trap message, but besides that, the warning is 'valid enough' for me

For the first issue, I tried to implement full multipolygon support, but (even though I currently fail to fetch the `inner` ways and thus not even work yet...) I know already that I won't be able to parse inner/outer ways that are comprised of non-ring linestrings that jointly form a ring. So it will never be perfect (and the inner part is prone to false positives).
<details>
The attempt to capture multipolygon geoms. Doesn't work, inner_ways is somehow always empty.

```
CREATE TEMP TABLE multipolygons AS
SELECT
  outer_way.id AS outer_wid,
  outer_way.linestring AS outer_linestring,
  CASE WHEN inner_ways.linestring IS NOT NULL THEN
    ST_MakePolygon(outer_way.linestring, array_agg(inner_ways.linestring))
  ELSE
    ST_MakePolygon(outer_way.linestring)
  END AS geom,
  r.tags AS tags
FROM
  relations AS r
  JOIN relation_members ON
    relation_members.relation_id = r.id AND
    relation_members.member_type = 'W'
  LEFT JOIN ways AS inner_ways ON
    member_role = 'inner' AND
    member_id = inner_ways.id AND
    ST_IsValid(inner_ways.linestring) AND
    ST_IsRing(inner_ways.linestring)
  JOIN ways AS outer_way ON
    member_role = 'outer' AND
    member_id = outer_way.id AND
    ST_IsValid(outer_way.linestring) AND
    ST_IsRing(outer_way.linestring)
WHERE
  r.tags->'type' = 'multipolygon' AND
  (
    inner_ways IS NULL OR
    ST_Within(inner_ways.linestring, ST_MakePolygon(outer_way.linestring))
  )
GROUP BY
  outer_way.id,
  outer_way.linestring,
  inner_ways.linestring,
  r.tags
```
</details>

Hence I'm wondering about your opinion: would you
- have suggestions on how to make multipolygon work with the countries where the way is connected to the landuse (either by getting ST_Touches to work with ST_Buffer, or by getting full multipolygon support)
- OR would you suggest we just disable this analyser for the countries where they connect the way to the landuse to avoid false positives? (Or maybe just disable multipolygon support for those)

analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -0,0 +1,316 @@
...
+  (tags?'natural' OR tags?'landuse') AND
+  -- Leave minipolygons for analyser osmosis_polygon_small
+  ST_Area(ST_MakePolygon(ST_Transform(linestring, {proj}))) >= 20
+UNION ALL
+SELECT
+  ways.id,
+  ways.linestring,
+  -- Better would be to parse inner and outer segments and create an SQL multipolygon
+  -- For now, just use a 0.0000000001m buffer zone around the linestring
+  ST_Buffer(ways.linestring, 0.0000000001, 'quad_segs=1 endcap=flat join=bevel') AS poly,
```
frodrigo, line 108: That in degrees, not in meters.

analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -0,0 +1,316 @@
...
+    landusage.linestring && mobility_ways.linestring AND
+    -- Only find ways where the highway/railway/waterway/aeroway actually enters the
+    -- polygon. Exclude cases where the polygon is tied to the way without going inside.
+    ST_Intersects(landusage.poly, mobility_ways.linestring) AND
+    NOT ST_Touches(landusage.poly, mobility_ways.linestring) AND
+    (
+      -- Ignore very minor overlaps. The distance between the center of the overlap and one of the ways must be at least 1m
+      -- Can't do this for multipolygons (without parsing all inner/outer segments) hence disabling this simplification for those
+      landusage.is_multipolygon OR
+      ST_Distance(ST_Centroid(ST_Transform(ST_Intersection(landusage.poly, mobility_ways.linestring), {proj})), ST_Transform(mobility_ways.linestring, {proj})) > 1 OR
```
frodrigo, line 169: The intersection could be a multi, and the centroid not a reliable value.

frodrigo:
If multipolygons are too complicated as first step we can ignore it. It was never done, but we can also build a clear `multiplygones` table like we build an `highways` one.

Abaout landuse glued to major ways. It is a bad practice. So, maybe just let start by raisin this everywhere.

analysers/analyser_osmosis_polygon_intersects.py
Famlam: I assume you're referring to the 0.0000000001m? If I read the docs correctly, it should be the radius, not the degrees, and thus be a distance unit, not a degrees value?

> For geometry, the distance is specified in the units of the Spatial Reference System of the geometry. For geography, the distance is specified in meters.

https://postgis.net/docs/ST_Buffer.html
(Although my assumption of meter is also wrong as I'm not passing it through ST_Transform; it just had to be very small)

analysers/analyser_osmosis_polygon_intersects.py
Famlam: You're right!
I couldn't think of a better, simple solution to detect how far a linestring penetrates into an area*. Hence, I've removed this "ignore small overlaps" code. They should be fixed anyway; they're just a bit more... spammy ;) (I only had this block to reduce the number of issues to empathize the most important ones with big overlaps.)

* The intersection result of linestring + polygon is also a linestring, so anything area-based will fail

analysers/analyser_osmosis_polygon_intersects.py
Famlam: I've removed the bad comment about meters :)

Famlam:
> If multipolygons are too complicated as first step we can ignore it. It was never done, but we can also build a clear multiplygones table like we build an highways one.

That sounds like a good idea: most polygon-based analysers ignore multipolygons at the moment, but there are plenty of analysers that could benefit from this. (But I'll keep it out of the scope of this PR then, and stick to the simple "ST_buffer on outer ways" workaround here).

> Abaout landuse glued to major ways. It is a bad practice. So, maybe just let start by raisin this everywhere.

I personally agree, but I know there's also plenty of people who strongly disagree, so for now I prefer to keep it on the safe side and not report these here. If we ever add an analyser for this, I think it should be a dedicated one, so the results can be disabled by the people and countries who don't like it, without them marking everything as a false positive. (P.s.: I managed to fix my issue with ways glued to multipolygons by using ST_crosses & ST_contains instead of ST_intersects and ST_touches, so now multipolygons behave like polygons as far as it involves crossings vs. touching...)

[results.zip](https://github.com/osm-fr/osmose-backend/files/11256053/results.zip)

analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -0,0 +1,322 @@
...
+  (r.tags?'natural' OR r.tags?'landuse')
+"""
+
+sql12 = """
+CREATE INDEX idx_mobilityway_linestring ON mobility_ways USING GIST(linestring);
+CREATE INDEX idx_landusage_linestring ON landusage USING GIST(linestring);
+"""
+
+sql13 = """
+-- Collect high|rail|water|aeroways either entering or fully inside landuse/natural geometries
```
frodrigo, line 141:
```suggestion
-- Collect highway|rail|water|aeroways either entering or fully inside landuse/natural geometries
```

frodrigo:
Thank you. Let's start with this version.

PR #1792, opened 2023-03-22
Workaround 1766

Famlam:
I don't expect I'll be able to fix the underlying issue of #1766 on a short notice. Hence I figured it would be good to temporary add the 'buggy' rule to the local mapcss rules (without the class that fails of course) so that we can also update combinations.mapcss without losing this rule (which a few weeks ago matched nearly 2k items worldwide).

This PR:
- Adds the rule that is incorrectly filtered out to Osmoses own rules
- Updates the mapcss files (including combinations.mapcss)

Does **not** fix 1766.
By updating combinations.mapcss, it does close issue 1658

(Of course, if anyone else expects to be able to fix 1766 soon, this PR is obsolete)


frodrigo:
Ok, on the idea.

We may also black list the rule from JOSM to avoid future duplicate ?

For now I have no free time to dig into this issue.

Famlam:
> We may also black list the rule from JOSM to avoid future duplicate ?

This will not work, because every time that JOSM changes something on the rule (any of the selectors, not just the buggy one) the subclass (and thus the ID to blacklist) will change, causing duplicate detections to appear. Whereas with the current approach, the `!.class` rule will never work in the JOSM file, so there will never be a duplication.

frodrigo:
You are right!

Issue #1790, opened 2023-03-21
Validation rule traffic sign: Unmapped road bump (Germany)


sun-geo:
I'm happy about general traffic sign recognition validation.

I noticed the task (8300/34) seems to be not suitable in Germany in general, because there doesn't exist such a traffic sign at all.

item 	8300   -   traffic sign 
class 	34     -   Unmapped road bump
[osmose link for item 8300 and class 34](https://osmose.openstreetmap.fr/de/issue/625e2ddd-ab9c-d547-aadc-98a9651432f9)
`traffic_calming=bump`[wiki](https://wiki.openstreetmap.org/wiki/Tag%3Atraffic_calming%3Dbump) / `hazard=bump`[wiki](https://wiki.openstreetmap.org/wiki/Tag:hazard%3Dbump) / sign: ![Vienna_Convention_road_sign_Aa-7b-V1 svg](https://user-images.githubusercontent.com/47290585/226734632-df5eb799-8801-406c-8a6f-6825345a4d37.png) [bump sign commons wikimedia](https://commons.wikimedia.org/wiki/File:Vienna_Convention_road_sign_Aa-7b-V1.svg) / [bump sign Category commons wikimedia](https://commons.wikimedia.org/wiki/Category:Diagrams_of_bump_warning_road_signs)

Therefore maybe there are almost all issue marker false positive cases, because of other wrong recognized signs.
The automatic traffic sign recognition is mostly wrong - instead of the "Unmapped road bump" sign there are other signs recognized, eg. sign: ![Vienna_Convention_road_sign_Aa-7a-V1 svg](https://user-images.githubusercontent.com/47290585/226736364-ed387d70-54fb-403b-b6d1-d3d057ab1aa5.png) [uneven road sign commons wikimedia](https://commons.wikimedia.org/wiki/File:Vienna_Convention_road_sign_Aa-7a-V1.svg) / [uneven road sign Category commons wikimedia](https://commons.wikimedia.org/wiki/Category:Diagrams_of_uneven_surface_warning_road_signs)

which is a different tag in OSM, see `hazard=damaged_road`[wiki](https://wiki.openstreetmap.org/wiki/Tag:hazard%3Ddamaged_road)
[...]

PR #1794, opened 2023-03-25
Disable traffic sign for countries where it does not exist

Famlam:
Disable bump detections for more countries where it does not exist
See https://github.com/osm-fr/osmose-backend/issues/1790#issuecomment-1483757846


frodrigo:
Thank you

PR #1795, opened 2023-03-26
Add analyser for unexpected barriers on highway

Famlam:
Class 3: detects any barrier on an important (`tertiary_link` or above) road. Such a barrier is likely wrong (either the road is not so important as claimed, or access restrictions (night-time closures or so) are missing, or the barrier is supposed to be on a connecting side road only.
Example: `barrier=kerb` on a primary highway. It should be either `highway=crossing` + `kerb=X` (no `barrier=*`), or the `barrier=kerb` tag should be on the connecting footway (as the kerb is not actually on the primary highway itself).
Example: https://www.openstreetmap.org/node/5798745719

Class 4: ~any barrier on a connection between a residential/unclassified road and a minor road (cycleway, path, service, ...). Very likely such a barrier is actually supposed to be located on the minor road only.~
EDIT, updated after feedback: any crossing with more than 2 minor outgoing ways, having a barrier on it. Very likely the barrier is only supposed to be on one of the ways.
Example: a cycle barrier on a crossing of multiple ways [not end nodes]... I wouldn't even know how it would look like to block cyclists from more than two directions.
Example: https://www.openstreetmap.org/node/1814866994#map=18/57.97252/19.22840&layers=N

<details>

This analyser is basically the SQL implementation of [JOSM's mapcss rule](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss):

```css
/* #20742 - No warning about barrier with inappropriate access tags on highway */
way[highway=~/^(motorway|motorway_link|trunk|trunk_link|primary|primary_link|secondary|secondary_link|tertiary|tertiary_link)$/][!access][!vehicle] > node[barrier][barrier!~/entrance|border_control|height_restrictor|toll_booth/][!access][!motor_vehicle][!vehicle] {
  throwWarning: tr("{0} without access tags such as {1}, {2}, or {3}.", "{0.tag}", "{2.key}", "{3.key}", "{4.key}");
  group: tr("suspicious barrier");
}

way[highway=~/^(footway|path|bridleway|cycleway|service)$/] > node[barrier]:connection {
  set barrierSmallRoadConnection;
}

way[highway=~/^(unclassified|residential)$/] >[index = 1] node.barrierSmallRoadConnection,
way[highway=~/^(unclassified|residential)$/] >[index = -1] node.barrierSmallRoadConnection {
  set barrierAllowedAtConnection;
}

/* 20742; warnings for major roads set by other rule, also issue 20742 */
way[highway=~/^(unclassified|residential)$/] > node[barrier][barrier!=bollard][!access][!access:conditional][!vehicle][!vehicle:conditional][!motor_vehicle][!motor_vehicle:conditional].barrierSmallRoadConnection!.barrierAllowedAtConnection {
  throwWarning: tr("Suspicious {0} on a connection of a small highway with a larger highway", "{0.tag}");
  set hasWarningForBarrierOnWay;
  group: tr("suspicious barrier");
}
```
</details>

[results.zip](https://github.com/osm-fr/osmose-backend/files/11072676/results.zip)

### DIFF
--- a/analysers/analyser_osmosis_highway_area_access.py
+++ b/analysers/analyser_osmosis_highway_area_access.py
@@ -91,6 +91,58 @@ WHERE
     NOT barrier.waytags->'{vehicle}' IN ('no', 'use_sidepath', 'unknown')
 """
 
+sql31 = """
+SELECT
+  highways.id,
+  barrier.id,
+  ST_AsText(barrier.geom)
+FROM
+  highways
+  JOIN nodes AS barrier ON
+    barrier.tags != ''::hstore AND
+    barrier.id = ANY(highways.nodes) AND
+    --barrier.tags?'barrier' AND -- commented to prevent slow route
+    barrier.tags->'barrier' NOT IN ('border_control', 'cattle_grid', 'entrance', 'height_restrictor', 'toll_booth') AND
+    NOT (barrier.tags?'access' OR barrier.tags?'access:conditional') AND
+    NOT (barrier.tags?'vehicle' OR barrier.tags?'vehicle:conditional') AND
+    NOT (barrier.tags?'motor_vehicle' OR barrier.tags?'motor_vehicle:conditional')
+WHERE
+  highways.level <= 3 AND
+  NOT (highways.tags?'access' OR highways.tags?'access:conditional') AND
+  NOT (highways.tags?'vehicle' OR highways.tags?'vehicle:conditional') AND
+  NOT (highways.tags?'motor_vehicle' OR highways.tags?'motor_vehicle:conditional')
+"""
+
+
+sql41 = """
+SELECT
+  ways.id,
+  barrier.id,
+  ST_AsText(barrier.geom)
+FROM
+  highways AS ways
+  JOIN way_nodes ON
+    way_nodes.way_id = ways.id
+  JOIN nodes AS barrier ON
+    barrier.tags != ''::hstore AND
+    barrier.id = way_nodes.node_id AND
+    --barrier.id = ANY(ways.nodes) will cause a very slow planning (focused on the X.tags?'Y') on some countries (e.g. japan_chubu)
+    barrier.tags?'barrier' AND
+    barrier.tags->'barrier' NOT IN ('bollard', 'border_control', 'cattle_grid', 'entrance', 'height_restrictor', 'toll_booth') AND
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_highway_area_access.py
```diff
@@ -91,6 +91,58 @@
...
+
+sql41 = """
+SELECT
+  ways.id,
+  barrier.id,
+  ST_AsText(barrier.geom)
+FROM
+  highways AS ways
+  JOIN way_nodes ON
+    way_nodes.way_id = ways.id
```
frodrigo, line 125: Usually join on linestring index is faster ways.linestring && minor_highway.linestring. Have you try it ?

Famlam: Good one, I forgot that linestrings also intersect if they just share 1 node (=crossing). Will try later this week!

Famlam:
> What do you think to check for barrier at intersection, event on minor road. An intersection look an odd place for barrier.

I agree :). 
I've implemented the feedback and the check for intersections on any road (excluding areas because of many dubious cases with highway=pedestrian as area).

Regarding the city gate, I don't think it's necessary to implement a dedicated analyser for this yet (in this PR). If they're on a major way, they'll be found by class 3 anyway, unless they have appropriate access tags. For minor roads, I don't think I can distinguish between i.e. historic city gates that are actually closed at times and city gates that are incorrectly tagged.

frodrigo:
Look good to me. Thank you.

praszuk:
It causes a lot of false positives with `barrier=kerb` (node) on crossings where 2 ways are connected e.g.:

Places where cycleway and footway merging into one way
https://www.openstreetmap.org/node/8332937889
https://www.openstreetmap.org/node/6300333817
https://www.openstreetmap.org/node/7283476949

or on simplified crossing where 2 way crossings are connected to 1 kerb node (e.g. for unmarked crossings)

Other (`barrier=gate`):
https://www.openstreetmap.org/node/7732100284

It probably shouldn't be treated as error too.

Marc-marc-marc:
> Other (`barrier=gate`):
> https://www.openstreetmap.org/node/7732100284

forr me it's a tagging errror
coming from https://www.openstreetmap.org/way/1072754579 to https://www.openstreetmap.org/way/1072754577 you don't cross at the gate, you cross before or after the gate, so the shared node for those ways isn't the gate

PR #1801, opened 2023-04-02
Add analyser: mismatching parking access

Famlam:
- Detects parkings with public access (`access=yes, permissive, ...` or `access` not set), where all ways leading to it have limited access (such as `private`)
- Adds tests
- ~(Also adds the `&&` intersection check to the existing sql)~

I got the idea after spotting my own mistake here:
https://www.openstreetmap.org/way/1127681810

The road towards the parking is access=private, but the parking itself is set as if it's a public (and free) one, making it stand out when I search for a free parking spot.

Some [results attached](https://github.com/osm-fr/osmose-backend/files/11133361/parking.zip)

### DIFF
--- a/analysers/analyser_osmosis_parking_highway.py
+++ b/analysers/analyser_osmosis_parking_highway.py
@@ -27,7 +27,9 @@ sql10 = """
 CREATE TEMP TABLE park_highway AS
 SELECT
   id,
-  linestring
+  linestring,
+  highway,
+  tags
 FROM
   highways
 WHERE
@@ -46,7 +48,8 @@ SELECT
   pr.tags->'park_ride' != 'no'
 FROM
   ways AS pr
-  LEFT JOIN park_highway as highway ON
+  LEFT JOIN park_highway AS highway ON
+    pr.linestring && highway.linestring AND
     ST_Intersects(pr.linestring, highway.linestring)
 WHERE
   ST_NPoints(pr.linestring) >= 2 AND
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_parking_highway.py
```diff
@@ -46,7 +48,8 @@
   pr.tags->'park_ride' != 'no'
 FROM
   ways AS pr
-  LEFT JOIN park_highway as highway ON
+  LEFT JOIN park_highway AS highway ON
+    pr.linestring && highway.linestring AND
```
frodrigo, line 52: This change is not required. `ST_Intersects` already use the index.

https://postgis.net/docs/ST_Intersects.html

analysers/analyser_osmosis_parking_highway.py
```diff
@@ -57,6 +60,47 @@
...
+  parking.id,
+  private_parking_way.id,
+  ST_AsText(ST_Centroid(parking.linestring)),
+  private_parking_way.tags->'access',
+  parking.tags->'access'
+FROM
+  ways AS parking
+  JOIN park_highway AS private_parking_way ON
+    parking.linestring && private_parking_way.linestring AND
+    ST_Intersects(parking.linestring, private_parking_way.linestring) AND
```
frodrigo, line 76: Same, `&&`, not required.

analysers/analyser_osmosis_parking_highway.py
```diff
@@ -57,6 +60,47 @@
...
+    private_parking_way.highway = 'service' AND
+    private_parking_way.tags?'access' AND
+    private_parking_way.tags->'access' IN ('private', 'permit', 'delivery', 'customers')
```
frodrigo, line 79: Add `private_parking_way.tags != ''::hstore AND` to be able to use the index (more a good practice to add it every where, even if may not required here).

frodrigo:
> Detects parkings with public access (access=yes, permissive, ... or access not set), where all ways leading to it have limited access (such as private)

As I understand the query, it is not what you implemented.

Using ST_Intersects also use highways inside the parking area. Probably better to use only external (part of) ways.

I think the first join can be avoided with something like this
```sql
SELECT
FROM
  ways AS parking
  JOIN park_highway // public and private access
GROUP BY parking.id
HAVING array_agg(park_highway->access) not only includes private access
```

analysers/analyser_osmosis_parking_highway.py
Famlam: I don't fully understand with what you mean by 'to be able to use the index'. (It'll always be true, since I'm using a subset of the table `highways`)

frodrigo: Right!

Famlam:
Done :)
New results attached: [results.zip](https://github.com/osm-fr/osmose-backend/files/11150460/results.zip)

analysers/analyser_osmosis_parking_highway.py
```diff
@@ -57,6 +59,40 @@
...
+    (
+      parking.tags->'access' NOT IN ('private', 'permit', 'delivery', 'customers', 'no') AND
+      parking.tags->'access' NOT LIKE '%;%'
+    )
+  )
+GROUP BY
+  parking.id,
+  parking.tags
+HAVING
+  ARRAY_AGG(parking_way.tags->'access') <@ array['private', 'permit', 'delivery', 'customers']
```
frodrigo, line 93:
```suggestion
  array_agg(parking_way.tags->'access') <@ array['private', 'permit', 'delivery', 'customers']
```

It is a function

Famlam: Done :).
I left the ST_* ones intact, as they're used throughout the analysers in various writing methods (`ST_*`, `St_*` and `st_*` all occur).
(`MIN` and `min` are both used in the various analysers; `SUM` function is always used as uppercase, but indeed, array_agg is always used lowercase)

analysers/analyser_osmosis_parking_highway.py
```diff
@@ -27,7 +27,9 @@
 CREATE TEMP TABLE park_highway AS
 SELECT
   id,
-  linestring
+  linestring,
+  highway,
+  tags
```
frodrigo, line 32:
```suggestion
  linestring,
  tags->'access' AS access
```

analysers/analyser_osmosis_parking_highway.py
```diff
@@ -57,6 +59,40 @@
   highway.id IS NULL
 """
 
+
+sql20 = """
+SELECT
+  parking.id,
+  max(parking_way.id),
```
frodrigo, line 66: If you wish you can list all access using array_agg with `any_full` or `array_full` in place of `way_full`.

Famlam:
Thanks! It will look a bit crowded on rare cases like this one (https://www.openstreetmap.org/way/176306209, having 17 access ways) but in other cases it will probably avoid confusion on why only one access way is highlighted.

I've also made sure to output all access values now, even though I haven't encountered a real-life case where one parking has dedicated access ways for i.e. customers and employees.

analysers/analyser_osmosis_parking_highway.py
Famlam: Good catch, indeed no longer needed

frodrigo:
Looks good. Thank you !

analysers/analyser_osmosis_parking_highway.py
Famlam: I noticed any_full isn't documented, so I made a separate PR for this :)

PR #1814, opened 2023-04-17
Remove previous PyKOpeningHours to make sure to update it

jocelynj:


jocelynj:
@Marc-marc-marc : what do you think of this change?  I've seen that pykopeninghours is not recompiled even if the requirements.txt file has been updated.

frodrigo:
It will recompile it each time, doesn't it? And it is not so fast.

jocelynj:
No, it will recompile only if `requirements.txt` is younger than `/tmp/requirements.txt.$USER.done`.

frodrigo (approved):

PR #1815, opened 2023-04-17
Fix TagWatchFrViPofm after wiki updates

Famlam:
Renamed earlier today by Tigerfell https://wiki.openstreetmap.org/w/index.php?title=Tagging_mistakes&action=history

Also Chapelle was removed from the wiki ( https://wiki.openstreetmap.org/w/index.php?title=Tagging_mistakes&type=revision&diff=2505513&oldid=2505107 )


frodrigo:
Ok. Thank you to look at that.

Issue #1813, opened 2023-04-16
This indoor feature is not reachable 


Marc-marc-marc:
item | 1300
class | 2
a indoor room https://www.openstreetmap.org/way/1159640032
the door https://www.openstreetmap.org/node/4888497424
connected to the "external" network https://www.openstreetmap.org/way/1163077113
something missing in osm or bug ?

Famlam:
It seems highway=path isn't included here:
https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_indoor.py#L116

Not sure if that was intentional, @nlehuby ?

nlehuby:
I would rather use footway in that case, but yes, path could be added here.

Marc-marc-marc:
thanks for the improvement (long debate on whether or not a highway=* without a blue pedestrian sign is a footway asnothing prohibits me from cycling to enter some shops especially bike-related shops :) )

it is not only highway=path that is missing but also some values for 4 wheel-wide vehicles.
e.g. the room in a garage is accessible if it is connected with a highway=service (even if at the other end it may exist a pedestrian exit, even without the internal pedestrian path, it is already routable, including by foot using the highway=service if allowed (=for ex for the private garage next a house)

frodrigo:
Deployed

PR #1816, opened 2023-04-18
Allow highway=path to indoor features

Famlam:
See #1813 
Although usually `highway=footway` seems more appropriate, it doesn't seem wrong to have `highway=path` (the wiki of highway=path also indicates there's a lot of debate on when something is truely a footway)

Additionally, the warning message (`This indoor feature is not reachable`) is very confusing for such situations, as an indoor feature connected by a path is actually connected.

Hence lets tolerate `highway=path`


frodrigo:
Yes, better avoid false positives.

PR #1826, opened 2023-04-25
Add warning when analyser ran while not in config

Famlam:
To make it more easy to spot when I'm running an analyser (with `--analyser=*` as I always do) without adding it to the config properly (see #1825), add a warning message.
(A "normal" test didn't seem like a good idea to me, since sometimes analysers are disabled on purpose, for example merge_street_objects at the moment, but this would pop up as a useful warning that you're running something via `--analyser=*` that wouldn't run normally)

Feel free to reject this idea, it was just a thought to make Osmose more Famlam-proof 😉

### DIFF
--- a/osmose_config.py
+++ b/osmose_config.py
@@ -2026,6 +2026,21 @@ class TestPluginCommon(unittest.TestCase):
         duplicate_polygon_ids = list(map(lambda a: a[0], filter(lambda c: c[1] >= 2, Counter(polygon_ids).items())))
         assert len(duplicate_polygon_ids) == 0, "Duplicate relation IDs: {}".format(sorted(duplicate_polygon_ids))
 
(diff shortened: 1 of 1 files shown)


jocelynj:
It might be possible to add a test to osmose_config.py, checking that all analysers on all available countries has a corresponding analyser python file.

I'm not sure that your proposition is really useful, and it wouldn't protect against another issue, which would be a test added to osmose_config.py, but forgetting to merge the analyser file.

Famlam:
@jocelynj much better suggestion, implemented :)

jocelynj (approved):
This looks good, thanks!

Famlam:
I guess "approved" means "feel free to merge"?

frodrigo:
Yes, please, merge it.

Issue #1827, opened 2023-04-25
Load of Osmose MapCSS in JOSM


frodrigo:
We try to list the Osmose MapCSS for JOSM.
But the load of Osmose MapCSS fails on JOSM

1. Ensure Osmose MapCSS are loadable by JOSM.
2. Can/should we package all Osmose MapCSS in one Zip?
3.  Include translation from Osmose.

Famlam:
> Ensure Osmose MapCSS are loadable by JOSM.


With the PR in #1829 every rule can be loaded in JOSM. The issues were:
- `'` is a special character in JOSM. You can't use it for string quotations, and if you want to use it in a string, you'll have to use `''` to get a single apostrophe (we don't parse this: we just output two aposthrophes). Or replace it by i.e. `.)
- `[key1!=tag(key2)]` doesn't work in JOSM. (And I'm actually surprised it does in Osmose as it looks to me as if you're comparing the string "key1" to the value of tag "key2", instead of the value of key1 with the value of key2). Anyway, the `!=*` operator works, and otherwise `tag("key1")!=tag("key2")` would've worked too
- In autofixes, JOSM first executes all functions (like `replace`, before it parses placeholders like `{0.value}`. Hence, the hypothetical `fixAdd: concat("x=", replace("{0.value}", "value", "key"));` on `way a=b` would give `x=a` as output, rather than `x=b`.

(I encountered all of these before 😉, either in my Dutch filters or my local mapcss file)

frodrigo:
> You can't use it for string quotations

Should we remove simple quoted string from grammar ? https://github.com/osm-fr/osmose-backend/blob/master/mapcss/MapCSS.g4#L108

> you'll have to use '' to get a single apostrophe (we don't parse this: we just output two aposthrophes)

We can fix this, but probably not in the grammar.

> [key1!=tag(key2)] doesn't work in JOSM. (And I'm actually surprised it does in Osmose as it looks to me as if you're comparing the string "key1" to the value of tag "key2", instead of the value of key1 with the value of key2). Anyway, the !=* operator works, and otherwise 

Ok, it generates the same python code.

> In autofixes, JOSM first executes all functions [...]

Ok.

Famlam:
> Should we remove simple quoted string from grammar ? https://github.com/osm-fr/osmose-backend/blob/master/mapcss/MapCSS.g4#L108

I considered it as an additional feature we support :). 

> We can fix this, but probably not in the grammar.

[...]

PR #1829, opened 2023-04-26
Mapcss 2 josm

Famlam:
Solves the first point of #1827. See the issue for the explanation.

Please double-check the autofixes for Farsi; my computer has some difficulties with writing Arabic/Farsi characters.


frodrigo:
Thank you.

Issue #1833, opened 2023-04-29
Report miss usage of landuse=village_green


frodrigo:
In many countries landuse=village_green is used just to get green area on the map. Eg in roundabout center.

https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dvillage_green

Idea to stop false positive:
- Inside roundabouts or between two roads
- Too small
- Too many `landuse=village_green` in the same area.
- ST_MaximumInscribedCircle

Famlam:
I attemped this already in https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_polygon_small.py#L93, but with a very conservative value.

I guess bumping it up to 500m2 (22x22m square or 25m diameter round) should be no problem? Maybe even 1000m2 (31x31m square or 36 diameter).

frodrigo:
The excessive usage of `landuse=village_green` could come from bad translation / preset in JOSM.

Famlam:
I discovered why the existing validator didn't work anymore... apparently a ` snuck in somewhere without me noticing...
But since that analyser only checks for area (so a long, stretched piece of land would go undetected) it is still good to add an extra check with `ST_MaximumInscribedCircle`

PR #1835, opened 2023-04-29
Fix village_green & increase minimum to 500m2

Famlam:
The idea with `ST_MaximumInscribedCircle` in #1833 is still a good idea to implement, so this does not fix that issue, but for now, at least fix the existing validator. Apparently a ` snuck in without me noticing...
Also increase the minimum a bit to 500m2.

(500m2 = 22x22m square or 25m round)


frodrigo:
This is a ready to merge PR ?

Famlam:
Yes (unless you find 500 too much). It gives 268 matches in Slovenia and less on the other 11 countries I ran it on.
Mind that modules/downloader may get stuck again while running the tests

frodrigo:
Ok. Thank you.

PR #1843, opened 2023-05-02
fix old tag type -> genus

Marc-marc-marc:


Famlam:
I suspect (based on occurances later in the code) that it was referring to https://wiki.openstreetmap.org/wiki/Key:leaf_type instead of genus?

Marc-marc-marc:
there are actually 4 tags now: genus species leaf_type leaf_cycle
the analysis was specialised in detecting species values in French, I don't think leaf_type leaf_cycle/leaf_type is concerned (i see no species in french looking at https://taginfo.openstreetmap.fr/keys/leaf_type#values)

I hesitated to put all 4 so as not to make a too long message and I chose genus to replace type as it is the other tag likely to have a value in French.
I'm not sure if this is a good idea, it might be an opportunity to expand the test or to add another analysis for the detection of name=* type=* in English or split it in 2 :) fix the type->genus in the text and another PR to extend it

Famlam:
~(I guess the limited allowed values for `leaf_type` and `leaf_cycle` could be in TagFix_BadValue as hardcoded values, not restricted to FR, but that'd indeed be a different PR)~

Ok, you're right, most of the bad `leaf_type` values are `arbre` which apparently only means `tree` according to Google Translate. And changing the plugin at lines 94+ to replace the deprecated tag by the active one would result in people putting `species=arbre` with the current phrasing. The plugin definitely does need a bit of improvement, since now for example `denotation=garden` also results in `Put the right data in the correct tag, in this case species.` as fix message, which would cause an existing `species` to be overwritten with `garden` (which to my knowledge isn't a species nor a genus). Example: https://osmose.openstreetmap.fr/en/issue/6981c1e7-1ace-98a9-09d5-4a418b90e642

PR #1845, opened 2023-05-03
Exclude mini-intersections

### LINKED ISSUE #1836: Implement a new "remove minor overlap" method
>> https://github.com/osm-fr/osmose-backend/pull/1788#pullrequestreview-1386809497 (I couldn't think of a better solution than the flawed method I had there, but this code was meant to remove minor overlaps)

>Right! But in this case we have to ST_Dump the intersection

Famlam:
Fixes #1836

(Note for reviewing: many lines are just indention changes, so switching off showing whitespace changes is recommended) 

For some bigger extracts:
usa_iowa: 157->130 results
slovenia: 543->388 results
netherlands_gelderland: 222->127 results
japan_chubu: 7025->6479 results
germany_saarland: 268->248 results

No significant differences in runtime.

Also whitelists `landuse=quarry`, which often contains railways

### DIFF
--- a/analysers/analyser_osmosis_polygon_intersects.py
+++ b/analysers/analyser_osmosis_polygon_intersects.py
@@ -217,11 +217,9 @@ WHERE
       mobility_way_type != 'highway'
     )
   )
-GROUP BY
+ORDER BY
   mobility_way_id,
   landusage_way_id,
-  landusage_poly,
-  mobility_way_linestring,
   mobility_way_type,
   tag_way,
   tag_polygon
--- a/analysers/analyser_osmosis_polygon_intersects.py
+++ b/analysers/analyser_osmosis_polygon_intersects.py
@@ -135,41 +135,67 @@ sql13 = """
 -- Collect highway|railway|aeroways either entering or fully inside landuse/natural geometries
 CREATE TEMP TABLE intersecting_geometries AS
 SELECT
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -135,41 +135,67 @@
...
+      ST_Distance(ST_Centroid(intersection_part_geom), ST_Transform(mobility_way_linestring, {proj})) > 1 OR
+      ST_Distance(ST_Centroid(intersection_part_geom), ST_Transform(landusage_linestring, {proj})) > 1
     )
+  )
 """
 
 sql14 = """
 -- Intersections with railway, aeroway or highway
 SELECT
+  DISTINCT ON (mobility_way_id, landusage_way_id)
```
frodrigo, line 198: Usually is `DISTINCT ON` + `ORDER BY`, or `GROUP BY`. Not a mix of `DISTINCT ON` + `GROUP BY`.

Famlam: Good point. I'll change it to `GROUP BY`.
(I can also use a regular `DISTINCT` and no `ORDER BY`, but this would give a double output for linestrings used in two multipolygon relations with different tags as outer. (Which, admittedly, is rare))

frodrigo:
Thank you.

PR #1847, opened 2023-05-05
Follow-up natural=tree + type

Famlam:
Follow-up for after #1843 
- Adds a warning that `type` on `natural=tree` is deprecated (except for palm). About 3500 hits worldwide.
- Removes the above warning from the FR-only analyser. The keys that would be valid if `leaf_type` was used, no longer exist in the database, all remaining type keys are broken.
- Adds a warning for leaf_type and leaf_cycle having an undocumented value (approx 7300 worldwide)

Still have to test, may be completely broken :)


Famlam:
@Marc-marc-marc what do you think of this follow-up to get rid of the old `type` tag?

Marc-marc-marc:
> @Marc-marc-marc what do you think of this follow-up to get rid of the old `type` tag?

not tested but nicie improvement
for palm -> taxon=Arecaceae taxon:en=Palm which I have previously documented on https://wiki.openstreetmap.org/wiki/Key:wood because both key have similar issues

Marc-marc-marc:
add "fix:chair", "deprecated" to osmoseTags: list ?

Famlam:
> for palm -> taxon=Arecaceae taxon:en=Palm which I have previously documented on https://wiki.openstreetmap.org/wiki/Key:wood because both keys have similar issues

Hmm, I'm a bit uncomfortable to add a rule that proposes these tags now, for the following reasons:
- I have little to no knowledge on palms or taxonomy, but it's not really clear to me what the key `taxon` is supposed to contain (the "more loose" species?)
- In addition, the key (and namespace) [taxon](https://wiki.openstreetmap.org/wiki/Key:taxon) seems to challenge `genus` and `species`:
> 'The use of taxon prevents the proliferation of various tags to describe biological components of mapped objects (e.g., name:biological=* genus=* species=*)'

However, taxon is `in use` whereas the other two are `approved` (rather than i.e. deprecated). Also, taxon is much less used than the other two with no clear sign of going to overtake them anytime soon: it's main increase is two big steps in the past. Hence it feels uncomfortable to suggest one key rather than the other without a clear directive (like deprecation) for any of them. Especially since:
- the current wiki of [Key:type](https://wiki.openstreetmap.org/wiki/Key:type) demands a `genus` or `species` tag before `type=palm` may go, not a `taxon` tag, so I would contradict the wiki by forcing another tag (and it's also not listed as the proposed fix for the bad tagging on [leaf_type#Possible_tagging_mistakes](https://wiki.openstreetmap.org/wiki/Key%3Aleaf_type#Possible_tagging_mistakes)
- Furthermore, the wiki of [Key:type](https://wiki.openstreetmap.org/wiki/Key:type) clearly warns against removing `type=palm` without adding other tags. It is logical to assume the same holds for `leaf_type=palm`, even though that's not listed explicitly as bad tagging. However, not listing `palm` in TagFix_BadValue is likely to result in at least some people just removing the bad key.
- Lastly, a bad argument, but it was also whitelisted in the previous implementation, blacklisting it now would cause another 32k (type=palm) or 10k (leaf_type=palm) to suddenly appear

Hence I prefer to stay on the safe side with this, for now at least

---

> add "fix:chair", "deprecated" to osmoseTags: list ?

Done ;)

Famlam:
@frodrigo the PR is ready for review, but it may be good to wait with merging this PR until #1853 is merged as it'll possibly get a merge conflict (but solving merge conflicts in the translations is more difficult than just regenerating the mapcss here)

frodrigo:
@Marc-marc-marc actual changes looks ok for you ?

Marc-marc-marc:
> > for palm -> taxon=Arecaceae taxon:en=Palm which I have previously documented on https://wiki.openstreetmap.org/wiki/Key:wood because both keys have similar issues
> 
> Hmm, I'm a bit uncomfortable to add a rule that proposes these tags now, for the following reasons:
> 
>     * I have little to no knowledge on palms or taxonomy, but it's not really clear to me what the key `taxon` is supposed to contain (the "more loose" species?)

the wiki page for this key could indeed be improved.
taxon is the generic term describing all categories (of a plant in our case), including genus and species

I see 2 possible uses for this key:
- you don't know if "foo" is a genus or a species -> you put in taxon, that has the merit of not being erroneous
- you want to add information that is neither a genus nor a species -> taxon (this is the case we are interested in: palm is a family (the category that groups several genus) https://en.wikipedia.org/wiki/Arecaceae

I will work on rewording the wiki page and post on tagging for proofreading.
if it's whilelisted for now, no worries about keeping it whitelisted for now, the most important thing is not to propose an erroneous fix that would only propagate a new error.

Marc-marc-marc:
> @Marc-marc-marc current changes looks ok for you ?

yes

PR #1820, opened 2023-04-23
Add check for shop/office/craft too close to a way

Famlam:
Implements #1819

Adds a check for `shop`, `craft` or `office` entries as nodes that are too close to a way. (Buildings tagged with shop/office/craft overlapping with highways are already caught).

Implementation as close as possible to tree & power object too close to way, with a primary difference that I include the buffer zone in the `&&` check (otherwise a node close to an end point of a curved way could get lost)

---

Unrelated, but it may be useful to rename table `highway` (`sql00`) to `highway_railway`. I confused myself a couple of times with the generic table `highways`, but table `highway` includes some railways too (but is missing [`light_rail`](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dlight_rail) I think). Also I suspect line 482 should be `self.create_view_touched("water", "W")` instead of `..., "N")`. And `elevator` is whitelisted twice in sql00 and sql10. But this can all be a future PR.

---

Some results in the details link below:
<details>

**Slovenia**:
```
<error class="12">
<node id="3438934568" lat="46.566022" lon="15.943065" version="2" user="">
<tag k="name" v="Zavarovalnica Maribor" />
<tag k="office" v="insurance" />
<tag k="addr:city" v="Cerkvenjak" />
<tag k="addr:street" v="Cerkvenjak" />
<tag k="addr:postcode" v="2236" />
</node>
<way id="336870964" version="6" user="">
<tag k="ref" v="734" />
<tag k="name" v="Cerkvenjak" />
<tag k="highway" v="tertiary" />
<tag k="surface" v="asphalt" />
<tag k="maxspeed" v="30" />
</way>
<location lat="46.566022" lon="15.943065" />
</error>
<error class="12">
<node id="5860680485" lat="46.2329506" lon="15.6376553" version="1" user="">
<tag k="name" v="Legro" />
<tag k="shop" v="shoes" />
</node>
<way id="847741302" version="1" user="">
<tag k="name" v="Strma cesta" />
<tag k="highway" v="residential" />
<tag k="surface" v="asphalt" />
</way>
<location lat="46.2329506" lon="15.6376553" />
</error>
```

**Japan_Chubu**:
```
<error class="12">
<node id="828448475" lat="34.9747325" lon="138.3834197" version="2" user="higa4">
<tag k="name" v="きものの成田屋 (Naritaya)" />
<tag k="shop" v="clothes" />
</node>
<way id="69080689" version="8" user="pon_de_lions">
<tag k="oneway" v="yes" />
[...]

PR #1848, opened 2023-05-05
Add extra tests

Famlam:
As requested in https://github.com/osm-fr/osmose-backend/pull/1820#issuecomment-1536663565 by Jocelyn

Also remove unneeded visible='true' (which JOSM adds and I normally remove, but forgot last time apparently)


frodrigo:
Thank you.

PR #1849, opened 2023-05-05
Fix postcode test

Famlam:
The test was failing due to wiki update https://en.wikipedia.org/w/index.php?title=List_of_postal_codes&diff=1153347765&oldid=1151168639

This switches to a different country with the same scheme as previously listed for the US

### DIFF
--- a/plugins/TagFix_Postcode.py
+++ b/plugins/TagFix_Postcode.py
@@ -214,6 +214,17 @@ class Test(TestPluginCommon):
         assert not a.node(None, {"addr:postcode":"HM 02"})
         assert a.node(None, {"addr:postcode":"plopplop"})
 
+    def test_SA(self):
+        a = TagFix_Postcode(None)
+        class _config:
+            options = {"country": "SA", "project": "openstreetmap"}
+        class father:
+            config = _config()
+        a.father = father()
+        a.init(None)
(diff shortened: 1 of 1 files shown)


frodrigo:
Better switch last US test to `addr:postcode` tag, doesn't ?

Famlam:
Depends on what you want to have tested. I thought it would make sense to test the same code evaluation as before, rather than stick to the country?

frodrigo:
With the list of existing tests I think it as about countries, but we can make both ^^.

Famlam:
Haha :)

I'm fine with doing so, however...
I suspect the user actually broke the wiki page (in which case we'll see many errors soon in that country I guess). For the US, he changed
`| NNNNN, NNNNN-NNNN` to
`| NNNN` + `|NNNNN-NNNN`, whereas his comment seems to suggest it should be  
`| NNNNN` + `|NNNNN-NNNN`

So I'm a bit hesitant to add a test for US now...
(And I'm not a local or any other way familiar with the US postal code system, so I won't edit the wiki myself)

frodrigo:
I fixed the US postcode to five N on the wikipage.

Famlam:
Now it has:
1. a test with the new US format 
2. a test for a different country so the old format is still being tested.

frodrigo:
Thank you.

PR #1851, opened 2023-05-06
Imporve perf of Analyser osmosis highway vs building.py

frodrigo:
Optimize the more time consuming analyzer.

First commit : ensure not duplicate issue reported. Required for the second commit.
Second commit : split the highway to hit less objects on index (especially building).

After the first commit, on tested extract (Gironde) the result is the same.
After second commit there is few more reported issues because multiples intersection on same highway x waterway including one issue intersection and one not issue intersection are not reported. So the split may report previously hidden issues.

Time before commits : 9m42.061s, 9m26.890s, 8m41.663s, 9m48.615s
Time after first commit : 9m48.978s, 9m48.820s, 8m56.001s, 9m50.080s
Time after second commit: 7m15.892s, 7m25.198s, 7m3.211s

### DIFF
--- a/analysers/analyser_osmosis_highway_vs_building.py
+++ b/analysers/analyser_osmosis_highway_vs_building.py
@@ -26,10 +26,12 @@ from .Analyser_Osmosis import Analyser_Osmosis
 
 sql00 = """
 CREATE TEMP TABLE {0}highway AS
+WITH ways AS (
 SELECT
     id,
     linestring,
     ST_Transform(linestring, {1}) AS linestring_proj,
+    ceil(ST_Length(ST_Transform(linestring, {1})) / 500)::integer AS split_n,
     nodes,
     tags->'highway' AS highway,
     coalesce(tags->'level', '0') AS level,
@@ -41,7 +43,7 @@ SELECT
     END AS layer,
     tags ?| ARRAY['ford', 'flood_prone'] AS onwater
 FROM
-    {0}ways AS ways
+    {0}ways
 WHERE
     tags != ''::hstore AND
     ((
@@ -55,6 +57,33 @@ WHERE
     NOT tags?'area:highway' AND
     array_length(nodes, 1) <= 100 AND -- Large ways have too big bbox
     ST_NPoints(linestring) > 1
+)
+SELECT
+    id,
+    linestring_part AS linestring,
+    linestring_proj_part AS linestring_proj,
+    nodes,
+    highway,
+    level,
+    layer,
+    onwater
+FROM
+    ways
+    CROSS JOIN LATERAL (
+        SELECT
+            ST_LineSubstring(
+                linestring,
+                i::float / split_n,
+                (i + 1)::float / split_n
+            ) AS linestring_part,
+            ST_LineSubstring(
+                linestring_proj,
+                i::float / split_n,
+                (i + 1)::float / split_n
+            ) AS linestring_proj_part
+        FROM
+            generate_series(0, split_n - 1) AS t(i)
+    ) AS t
 """
 
 sql01 = """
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_highway_vs_building.py
```diff
@@ -202,7 +236,7 @@
 """
 
 sql30 = """
-SELECT
+SELECT DISTINCT ON (highway.id, tree.id)
```
Famlam, line 239: As tree is a point object, it may be enough to give a single warning (e.g. `DISTINCT ON (tree.id)`), even if two highways are nearby, right? 
If two highways are too nearby, it is very likely that it is the tree (and thus not the highways) is positioned wrong. And solving it for one will then also solve (or modify) the other warning. Just a suggestion, feel free to ignore or postpone to a future PR.

analysers/analyser_osmosis_highway_vs_building.py
```diff
@@ -214,10 +248,13 @@
...
         tree.geom && highway.linestring AND
         ST_Intersects(ST_Buffer(tree.geom::geography, 0.25)::geometry, highway.linestring)
+ORDER BY
+    highway.id,
+    tree.id
 """
 
 sql31 = """
-SELECT
+SELECT DISTINCT ON (highway.id, power.id)
```
Famlam, line 257: Same as line 239

analysers/analyser_osmosis_highway_vs_building.py
```diff
@@ -26,10 +26,14 @@
...
 sql00 = """
 CREATE TEMP TABLE {0}highway AS
+WITH ways AS (
 SELECT
     id,
     linestring,
     ST_Transform(linestring, {1}) AS linestring_proj,
+    ST_Length(ST_Transform(linestring, {1})) AS length,
+    ceil(ST_Length(ST_Transform(linestring, {1})) / 500)::integer AS split_n,
+    ST_Length(ST_Transform(linestring, {1})) / ceil(ST_Length(ST_Transform(linestring, {1})) / 500)::integer AS split_length,
```
Famlam, line 36: `length` and `split_length` are only used in the combination `split_length / length`. However, given how they are defined, `split_length / length` = `1 / split_n` (i.e. the fraction of the linestring length).

Maybe it is clearer to omit `length` and `split_length` and use
`1/ceil(ST_Length(ST_Transform(linestring, {1})) / 500) AS split_length_fraction,` instead
(or `1 / split_n` in the code below instead of `split_length / length`)

Famlam (approved):
Looks good to me (although, admittedly, I'm not familiar with `CROSS JOIN LATERAL`, so I just trust you on that ;) ).
Some minor suggestions, one mathematically and two just conceptually, but they won't cause issues if they're not resolved.

analysers/analyser_osmosis_highway_vs_building.py
frodrigo: Right, fixed.

analysers/analyser_osmosis_highway_vs_building.py
frodrigo: Yes.

Famlam:
Looks good to me :)

jocelynj (approved):
This looks good, especially as it passes the new tests :)

Famlam:
@frodrigo a bit late, but just thinking: is this line still necessary now that this PR is in?
`array_length(nodes, 1) <= 100 AND -- Large ways have too big bbox`
https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_osmosis_highway_vs_building.py#L58

frodrigo:
Yes. I forgotten to remove it.

Famlam:
No problem, I'll include it in the PR I'm currently working on :)

Issue #1850, opened 2023-05-05
Avoid translating `level` (by making more clear its about the tag)


Famlam:
I always have a sneak peak at translation updates to check for odd Dutch translations 😇 and spotted that [`level` was translated in one](https://github.com/osm-fr/osmose-backend/commit/7330225b8cfae568e14b9fbc5a754ee51a77578b#diff-44a066469e7c75b3cdbbe18743cae1045124dcb86d20c65269ba4c323b04f61cR3774-R3780), at a point it's referring to the key `level` rather than the word `level`.

I updated it in the local translations minutes ago, but to avoid it from happening again, we should probably change:
https://github.com/osm-fr/osmose-backend/blob/952f18a491108ff484f1a7734fa57b7c930769a0/plugins/TagFix_MultipleTag2.validator.mapcss#L114-L122

to:

```css
way[waterway][level] {
    throwWarning: tr("Waterway with `level`");
    assertMatch: "way waterway=stream level=-1";
    fixChangeKey: "level=>layer";
    -osmoseItemClassLevel: "3032/30327:0/2";
    -osmoseTags: list("tag", "fix:chair");
    -osmoseDetail: tr("The tag `level` should be used for buildings, shops, amenities, etc.");
    -osmoseTrap: tr("Remove `level` and check if `layer` is needed instead.");
}
```

And then run jocelyns [magic script again](https://github.com/osm-fr/osmose-frontend/pull/451#issuecomment-1493352086). Since it's an [old string from 2020](https://github.com/osm-fr/osmose-backend/commit/5aa153813318e6a5e9a3bbee75ca9bca8fb28f1a#diff-7d0c24935629f7d41ceffecdde7926e44eb4c77848421be2de510a1b9ba531c4R38) quite some translations have it, of which most seem to be ok.

Unfortunately my local copy of Osmose is a bit "dirty" at the moment 😇 so filing it for later.

Famlam:
@jocelynj Just to be sure, can I actually do this or would it require updating the translations directly before and directly after to avoid loss of recently translated strings? (In which case I'd just prepare the python code and ask you to do it)?

frodrigo:
As I understand, the issue is only when there is lot of changes in the string. When there is only small changes it keep the link between source an translated strings, but flag it as  "review required"

jocelynj:
The script doesn't work that well with strings from mapcss, and I had to tweak it.

[...]

PR #1853, opened 2023-05-06
Quotes around `level`

Famlam:
See #1850

@jocelynj I hope this is what you meant when you said:
>What might be a better way for simple change like this one is to modify the .po files on the pull-request, and I can then push to transifex if I'm informed.


jocelynj:
Yes, it is what I meant. I see that some translated strings will need to be changed afterwards: we can try this change, and check if transifex marks these strings as "review required".

Famlam:
> I see that some translated strings will need to be changed afterwards: we can try this change, and check if transifex marks these strings as "review required".

I'm not sure if you mean I should change something, or that you mean you or Frodrigo will merge & apply it to Transifex?

frodrigo:
@Famlam plase, can you rebase this branch.

jocelynj:
@Famlam : I've pushed the updated translations to transifex - hopefully, it worked as expected.

PR #1854, opened 2023-05-06
Add country analyzer name in table loaded from dynamic merge

frodrigo:
Add country analyzer name in table loaded from dynamic merge to avoid conflict on cached tables


jocelynj:
Could it increase the storage needed on backends?

It might if an opendata source is shared between several regions, and we don't filter by area on import to the database.

frodrigo:
> Could it increase the storage needed on backends?

Theoretically yes.

> It might if an opendata source is shared between several regions, and we don't filter by area on import to the database.

But in practice, all data source based on GDAL are filtered by polygon before load. And other country wide datasets are run on osm110.

I think there will be no storage increasement.

A next improvement could be to base all loaders on GDAL.

PR #1857, opened 2023-05-07
Exclude pedestrian from tree-too-close

### LINKED ISSUE #1852: `highway=pedestrian` area too close to tree
https://osmose.openstreetmap.fr/nl/issue/01a7493e-5f49-e1d1-3e24-de4e0b3d1c6f
https://www.openstreetmap.org/way/202048871

Probably we should exclude `highway=pedestrian` from sql30 in https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_highway_vs_building.py ("Tree and highway too close").

If imagery doesn't allow to see the exact "cutout" for the tree trunk in the pavement, I would tag it the same as is done here in case of an `highway=pedestrian` mapped as area.
(Although I prefer having `area=yes/no` on closed `highway=pedestrian`, it is not mandatory to my knowledge and it seems to be interpret as `area=yes` by default)

EDIT: better link

Famlam:
Fixes #1852 

Allow trees very close to the edge of `highway=pedestrian`.

Alternative would be something like:
```sql
(highway.highway != 'pedestrian' OR nodes[1] != nodes[array_length(nodes, 1)]) AND
```
(or adding `is_polygon` in table highway)

### DIFF
--- a/analysers/analyser_osmosis_highway_vs_building.py
+++ b/analysers/analyser_osmosis_highway_vs_building.py
@@ -243,7 +243,7 @@ FROM
     JOIN {1}highway AS highway ON
         highway.level = '0' AND
         highway.layer = '0' AND
-        highway.highway NOT IN ('footway', 'path', 'track') AND
+        highway.highway NOT IN ('footway', 'path', 'track', 'pedestrian') AND
         tree.geom && highway.linestring AND
         ST_Intersects(ST_Buffer(tree.geom::geography, 0.25)::geometry, highway.linestring)
 ORDER BY


frodrigo:
There is already `is_area`, it is not what you need?

Famlam:
No, the `area` key is optional (and not used here), if omitted it defaults to yes for `pedestrian` in most editors/renderers.

(Obviously I'm in favor of having the area key present, i.e. a bit south of my example in the issue, you can find one where it should have area=no)

frodrigo:
I am for adding `is_area` in the `highway` table and getting closer to reuse the common `highways` table.

And we can directly set `is_area=true` for `pedestrian` by default (and the same for `highways`).

Famlam:
> I am for adding is_area in the highway table and getting closer to reuse the common highways table.

As table `highway` (not `highways`) filters out areas
https://github.com/osm-fr/osmose-backend/blob/9fbe359f915dc7be024e091aa7cdc9e13c4a208f/analysers/analyser_osmosis_highway_vs_building.py#L56
adding `is_area` would be confusing, since all explicit areas have been removed from the table. I can either add `is_pedestrian_area` (and only use that in sql30), or - for consistency - filter out `highway=pedestrian` with `is_polygon` in sql00 already?

> And we can directly set is_area=true for pedestrian by default (and the same for highways).

👍

frodrigo:
> filter out highway=pedestrian with is_polygon in sql00 already?

Yes, but also with `is_area`.

Issue #1818, opened 2023-04-19
Create common table for multipolygons


Famlam:
As mentioned in https://github.com/osm-fr/osmose-backend/pull/1788#issuecomment-1510281766 it would be useful to have a table containing all multipolygons, similar to i.e. the tables 'highways' and 'buildings'. Parsing multipolygons isn't trivial:
- Outer:
  - Can be a closed linestring OR
  - Can consist of several connected linestrings forming a closed loop
  - 1 or more
- Inner:
  - Can be a closed linestring
  - Can consist of several connected linestrings forming a closed loop
  - 0 or more

[Example of a complicated multipolygon for demonstration purpose only (not a real-life example)](https://github.com/osm-fr/osmose-backend/files/11277000/difficult_mp.zip)

The primary use case would be that every current analyser that checks things (most notably intersections) on polygons could also support multipolygons using a common table.

frodrigo:
Note 1. We already have a validator. This should not be for detecting issue, but only for valid MP.
Note 2. I think `ST_MakePolygon` done most of the work

Famlam:
Regarding note 1: agree, I should've made this more clear
Regarding note 2: correct for the simple cases, but unfortunately it doesn't seem to support multiple outers or composite (non-closed) linestrings forming a closed loop. (And I personally have no clue how to handle the latter)

frodrigo:
I try an implementation.

The naive query is, but it work only with valid data in OSM
```sql
CREATE TEMP TABLE multipolygon AS
SELECT
    id,
    ST_BuildArea(linestrings) AS poly
FROM (
    SELECT
        relations.id,
        ST_LineMerge(ST_Collect(ways.linestring)) AS linestrings
    FROM
        relations
        JOIN relation_members ON
            relation_members.relation_id = relations.id AND
            relation_members.member_type = 'W' AND
            relation_members.member_role IN ('outer', 'inner')
        JOIN ways ON
            ways.id = relation_members.member_id
    WHERE
        relations.tags->'type' = 'multipolygon'
    GROUP BY
        relations.id
) AS t
WHERE
    ST_IsValid(linestrings)
;
```

I add validation before runnin the same idea.
```sql
CREATE TEMP TABLE multipolygon_parts AS
SELECT
    relations.id,
    relation_members.member_role AS role,
    (ST_Dump(ST_LineMerge(ST_Collect(ways.linestring)))).geom AS linestring
FROM
    relations
    JOIN relation_members ON
[...]

PR #1859, opened 2023-05-08
Add support of requires_tables multipolygons #1818

frodrigo:
Implement a new common required table for multipolygons. See #1818.

Use it in a first analyser as demo analyser_osmosis_polygon_small.

TODO:
- [x] Update the doc
- [x] Add warning, contains only closed and borders inside polygon
- [x] Use `LEFT JOIN` to fail on missing members
- [x] Remove version, user_id, tstamp, changeset_id

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -107,6 +107,92 @@ WHERE
 ANALYZE {0}.highway_ends;
 """
 
+    # Multipolygons table is not complete. It does not contain multipolygons across extract border or invalid ones.
+    sql_create_multipolygons = """
+DO $$
+DECLARE
+    mp RECORD;
+BEGIN
+    DROP TABLE IF EXISTS {0}.multipolygons;
+    CREATE UNLOGGED TABLE {0}.multipolygons (
+        id bigint,
+        tags hstore,
+        poly geometry(Geometry, 4326) NOT NULL,
+        poly_proj geometry(Geometry, {1}) NOT NULL,
+        is_valid boolean NOT NULL
+    );
+
+    FOR mp in (
+        SELECT
+            relations.id,
+            relations.tags,
+            ST_LineMerge(ST_Collect(ways.linestring)) AS linestrings
+        FROM
+            relations
+            JOIN relation_members ON
+                relation_members.relation_id = relations.id AND
+                relation_members.member_type = 'W' AND
+                relation_members.member_role IN ('outer', 'inner')
+            JOIN ways ON
+                ways.id = relation_members.member_id
+        WHERE
+            relations.tags->'type' = 'multipolygon'
+        GROUP BY
+            relations.id
+    ) LOOP
+        BEGIN
+            IF ST_BuildArea(mp.linestrings) IS NOT NULL AND NOT ST_IsEmpty(ST_BuildArea(mp.linestrings)) THEN
+                WITH
+                unary AS (
+                    SELECT
+                        id, tags,
+                        (ST_Dump(poly)).geom AS poly
+                    FROM (VALUES (
+                        mp.id,
+                        mp.tags,
+                        ST_BuildArea(mp.linestrings)
+                    )) AS t(id, tags, poly)
+                ),
+                simplified AS (
+                    SELECT
+                        id, tags,
+                        ST_BuildArea(ST_Collect(
+                            ST_ExteriorRing(poly),
+                            (SELECT ST_Union(ST_InteriorRingN(poly, n)) FROM generate_series(1, ST_NumInteriorRings(poly)) AS t(n))
+                        )) AS poly
+                    FROM
+                        unary
+                ),
+                multi AS (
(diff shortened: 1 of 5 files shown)


Famlam:
Looks great!
I build some difficult test cases (will propose them later, probably good to have; not specifically because this analyser is complicated but since it would show errors if the table is modified later) and they all pass.
I'll have a thorough look later this week, but a first question that arises in my mind is: why are these columns in the table: `version, user_id, tstamp, changeset_id`? (They're not in the other similar tables like `highways`)

analysers/Analyser_Osmosis.py
```diff
@@ -105,6 +105,93 @@ class Analyser_Osmosis(Analyser):
...
+            relations.*,
+            ST_LineMerge(ST_Collect(ways.linestring)) AS linestrings
+        FROM
+            relations
+            JOIN relation_members ON
+                relation_members.relation_id = relations.id AND
+                relation_members.member_type = 'W' AND
+                relation_members.member_role IN ('outer', 'inner')
+            JOIN ways ON
+                ways.id = relation_members.member_id
```
Famlam, line 134: If I understand https://github.com/osm-fr/osmose-backend/issues/1818#issuecomment-1522291312 correctly, the multipolygon will be incomplete at boundaries. This could cause false positives for area-based queries (especially with multiple outers) and similar kind of queries (i.e. if we use is_valid in analyser_osmosis_relation_multipolygon to detect bad ones).
The current implementation will not abort/fail if a member way is missing it seems, it just pretends that a non-downloaded way doesn't exist.

Is it possible (and not too slow) to use a `LEFT JOIN` instead of a regular `JOIN` with ways?
Then the `SELECT` could contain `COUNT(relation_id) = COUNT(ways.id) AS is_complete` (which could be added as a row to the final table)

That way, analysers where borders may cause issues can simply query on `multipolygon.is_complete`. 
Otherwise, each analyser where this is relevant would need to go through the following (dummy) code to filter such relations out:
```sql
SELECT
  mp.*
FROM
  multipolygons AS mp
  JOIN relation_members ON
    relation_members.relation_id = mp.id AND
    relation_members.member_type = 'W'
  LEFT JOIN ways ON
    ways.id = member_id
HAVING
    COUNT(mp.id) = COUNT(ways.id)
```

analysers/Analyser_Osmosis.py
```diff
@@ -105,6 +105,93 @@ class Analyser_Osmosis(Analyser):
...
+            relations.tags->'type' = 'multipolygon'
+        GROUP BY
+            relations.id
+    ) LOOP
+        BEGIN
+            IF ST_BuildArea(mp.linestrings) IS NOT NULL AND NOT ST_IsEmpty(ST_BuildArea(mp.linestrings)) THEN
+                WITH
+                unary AS (
+                    SELECT
+                        id, version, user_id, tstamp, changeset_id, tags,
```
Famlam, line 145: I'm not sure why `version, user_id, tstamp, changeset_id` need to be included? They I don't expect they'll be used in any analysers (and are also not present in `highways`, `buildings`, ...)?

Unless they're needed for `create_view_touched` for when an inner or outer way is changed (or a single node of an inner or outer way is moved, thereby causing i.e. the poly (and i.e. its area) to change, or ...)

analysers/Analyser_Osmosis.py
frodrigo: > Is it possible (and not too slow) to use a LEFT JOIN instead of a regular JOIN with ways?

Right added in the todo list.

> Then the SELECT could contain COUNT(member_id) = COUNT(ways.id) AS is_complete (which could be added as a row to the final table)

I prefer to not include incomplete MP.

analysers/Analyser_Osmosis.py
frodrigo: > I'm not sure why version, user_id, tstamp, changeset_id need to be included? I don't expect they'll be used in any analysers (and are also not present in highways, buildings, ...)?

To be removed.

frodrigo:
@Famlam lot of good feedback, thank you. I have no free time to work on this in the upcoming week. If you wish, you can work on this on a new branch under `osm-fr/osmose-backend`, so latter we could both work on it.

Famlam:
> @Famlam lot of good feedback, thank you. I have no free time to work on this in the upcoming week. If you wish, you can work on this on a new branch under `osm-fr/osmose-backend`, so latter we could both work on it.

Welcome :)
I've implemented it in #1861
I'm just not sure what you mean with `Add warning, contains only closed and borders inside polygon`

frodrigo:
> I'm just not sure what you mean with Add warning, contains only closed and borders inside polygon

I add it.

I change the having close to a simplest one.

Famlam (approved):
Looks good to me :)
(I haven't tested the diff mode though, but I assume it'll include changes in inner and outer members as it's basically the same as a regular relation diff)

frodrigo:
Let's merge it!

PR #1865, opened 2023-05-14
Treat `highway=pedestrian` polygons as areas

Famlam:
Part 1 of 2 of https://github.com/osm-fr/osmose-backend/issues/1852#issuecomment-1541835203

Treat closed `highway=pedestrian` polygons as areas by default, unless `area=no` is tagged.
This matches the behavior with at least Carto, iD and JOSM. (Note however that there is no documentation for this: the wiki also recommends adding `area`)

Best is to have an area tag present, of course, but it's not always the case.
Currently, worldwide, 229080 closed `highway=pedestrian`, of which 9620 without `area` tag (4%)

Confirmed with the Barcelona extract (which contains a few) that such ways do not get strange detections due to this PR

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -45,7 +45,10 @@ SELECT
     tags->'highway' LIKE '%_link' AS is_link,
     (tags?'junction' AND tags->'junction' = 'roundabout') AS is_roundabout,
     (tags?'oneway' AND tags->'oneway' IN ('yes', 'true', '1', '-1')) AS is_oneway,
-    (tags?'area' AND tags->'area' != 'no') AS is_area,
+    CASE
+      WHEN tags?'area' THEN tags->'area' != 'no'
+      ELSE (is_polygon AND tags->'highway' = 'pedestrian')
+    END AS is_area,
     tags->'highway' IN ('planned', 'proposed', 'construction') AS is_construction,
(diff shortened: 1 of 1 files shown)


analysers/Analyser_Osmosis.py
```diff
@@ -45,7 +45,10 @@ class Analyser_Osmosis(Analyser):
     tags->'highway' LIKE '%_link' AS is_link,
     (tags?'junction' AND tags->'junction' = 'roundabout') AS is_roundabout,
     (tags?'oneway' AND tags->'oneway' IN ('yes', 'true', '1', '-1')) AS is_oneway,
-    (tags?'area' AND tags->'area' != 'no') AS is_area,
+    CASE
+      WHEN tags?'area' THEN tags->'area' != 'no'
+      ELSE (is_polygon AND tags->'highway' = 'pedestrian')
```
frodrigo, line 50: Unlike the original, this can return NULL.
Maybe:
```suggestion
      ELSE (is_polygon AND tags?'highway' AND tags->'highway' = 'pedestrian')
```

Famlam: Everything without the tag `highway` is already filtered out in the WHERE statement, are you sure this is necessary to do it here in the SELECT again?

frodrigo: Right!

Issue #1830, opened 2023-04-26
Allow `unexpected char` for names in NSI


Famlam:
Plugin [Name_Script ](https://github.com/osm-fr/osmose-backend/blob/master/plugins/Name_Script.py) (item 5070, class 50701) reports when a name contains an character that's not common in the charset of the country. For example, in the case of [n5782442218](https://www.openstreetmap.org/node/5782442218), I get
> "name"="Søstrene Grene" onverwachte karakter "ø" (LATIN SMALL LETTER O WITH STROKE, 0x00f8)

(I marked it as false positive)

This shop however [is also in the NameSuggestionIndex](https://nsi.guide/index.html?t=brands&k=shop&v=interior_decoration&tt=S%C3%B8strene%20) with the same spelling. 
It would be good to whitelist names that are also in the NSI. We already do this for uppercase names:
https://github.com/osm-fr/osmose-backend/blob/07b1a016851186e87515e245cd4ce1d6d34f817b/plugins/Name_UpperCase.py#L59-L84

Since the code to do so would be (nearly) a duplicate, maybe it's an idea to move these NSI-parsing lines to a separate modules file, so that we can just call the same functions?

frodrigo:
Deployed.

Famlam:
I suspect that updating the plugin version doesn't clear the cached external resource (here: the NSI), correct?
Otherwise something is wrong: Søstrene Grene does still appear https://osmose.openstreetmap.fr/nl/issue/bc938e46-cf3f-c2aa-6c86-13cbc4d448da

frodrigo:
Yes, cache and code update are not related.

Famlam:
Now the issue is fixed properly, the error is gone ;)

PR #1867, opened 2023-05-17
Whitelist names in NSI from plugin Name_Script

Famlam:
See #1830

This PR
- Moves common functions to a common file in subfolder /modules/
  - The filter for uppercase names was of course left in the Name_Uppercase plugin
  - Copyright I copied from both existing files to the new file
- Uses the NSI to whitelist names (only the non-localized variant, e.g. `name` but not `name:nl`) that contain "uncommon" characters

Note: failing test is unrelated to this PR

### DIFF
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -22,8 +22,7 @@
 from modules.OsmoseTranslation import T_
 from plugins.Plugin import Plugin
 import regex as re
-from modules.downloader import urlread
-import json
+from plugins.modules.name_suggestion_index import whitelist_from_nsi
 
 # Whitelist of allowed capitals by country code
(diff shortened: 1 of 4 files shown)


plugins/Name_UpperCase.py
```diff
@@ -22,8 +22,7 @@
 from modules.OsmoseTranslation import T_
 from plugins.Plugin import Plugin
 import regex as re
-from modules.downloader import urlread
-import json
+from plugins.modules.name_suggestion_index import _whitelist_from_nsi
```
frodrigo, line 25: Remove the "_" on starts, it means private function and it is no more.

Famlam: Done, thanks!

frodrigo:
Thank you

PR #1870, opened 2023-05-19
Convert osmosis_polygon_intersects to use table multipolygons

Famlam:
- Converts the multipolygon workaround in polygon_intersects to full multipolygon support
- Adds `any_id` as the 'non-_full' equivalent of `any_full`
  - (suggestions for better names? I didn't want to overwrite the python `any` function. If the name is fine, I'll update the doc)
- Updates the tests to use the relation id, and adds two small tests

The results seem fine. The increase (detections inside multipolygons) or decrease (small incursions of outer ways are now filtered) of detections:
| extract | before | after | after switch to buffer |
| --- | --- | --- | --- |
| belgium_wallonia_german_community | 36 | 57 | 57 |
| jersey  | 3 | 3 | 3 |
| switzerland_basel_stadt  | 14 | 14 | 14 |
| sweden_gotland  | 1 | 1 | 1 |
| netherlands_gelderland  | 127 | 125 | 87 |
| usa_iowa  | 128 | 124 | 124 |
| slovenia  | 378 | 343 | 292 |
| germany_saarland  | 248 | 316 | 314 |
| japan_chubu  | 6569 | 10379 | 10333 |
| venezuela | 118 | 124 | 122 |

### DIFF
--- a/analysers/analyser_osmosis_polygon_intersects.py
+++ b/analysers/analyser_osmosis_polygon_intersects.py
@@ -76,8 +76,8 @@ sql11 = """
 CREATE TEMP TABLE landusage AS
 SELECT
   'W' || id AS id,
-  ST_Transform(linestring, {proj}) AS linestring_proj,
   ST_MakePolygon(linestring) AS poly,
+  ST_Transform(ST_MakePolygon(linestring), {proj}) AS poly_proj,
   CASE
     WHEN tags?'landuse' THEN 'landuse'
     WHEN tags?'natural' THEN 'natural'
@@ -97,33 +97,23 @@ WHERE
   ST_Area(ST_MakePolygon(ST_Transform(linestring, {proj}))) >= 20
 UNION ALL
 SELECT
-  'R' || mp.id AS id,
-  ST_Transform(ST_LineMerge(ST_Collect(ways.linestring)), {proj}) AS linestring_proj, -- usually multilinestring
-  mp.poly,
+  'R' || id AS id,
+  poly,
+  poly_proj,
   CASE
-    WHEN mp.tags?'landuse' THEN 'landuse'
-    WHEN mp.tags?'natural' THEN 'natural'
+    WHEN tags?'landuse' THEN 'landuse'
+    WHEN tags?'natural' THEN 'natural'
   END AS landusekey,
   CASE
-    WHEN mp.tags?'landuse' THEN mp.tags->'landuse'
-    WHEN mp.tags?'natural' THEN mp.tags->'natural'
+    WHEN tags?'landuse' THEN tags->'landuse'
+    WHEN tags?'natural' THEN tags->'natural'
   END AS landusevalue,
-  mp.tags
+  tags
 FROM
-  multipolygons AS mp
-  JOIN relation_members ON
-    relation_id = mp.id
-  JOIN ways ON
-    member_role IN ('outer', 'inner') AND
-    member_id = ways.id AND
-    member_type = 'W'
+  multipolygons
 WHERE
-  mp.is_valid AND
-  (mp.tags?'natural' OR mp.tags?'landuse')
-GROUP BY
-  mp.id,
-  mp.poly,
-  mp.tags
+  is_valid AND
+  (tags?'natural' OR tags?'landuse')
 """
 
 sql12 = """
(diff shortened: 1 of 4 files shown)


frodrigo:
> (suggestions for better names? I didn't want to overwrite the python any function. If the name is fine, I'll update the doc)

It is ok for me.

Nevertheless, can you please also update the doc.

Famlam:
Done :)

analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -98,37 +97,38 @@
...
-  r.tags,
-  TRUE AS is_multipolygon
+  mp.tags
 FROM
-  relation_members
-  LEFT JOIN relations AS r ON
-    relation_id = r.id
+  multipolygons AS mp
+  JOIN relation_members ON
+    relation_id = mp.id
```
frodrigo, line 115: No sure why we still need relation members?

Famlam: This particular analyser filters out minor intersections too. I do not have a good way to find (and filter) all minor intersections based on the geometry only.
I'm not referring to perpendicular incursions (such as the one on the left in the example picture) as they can easily be filtered by length, but are relatively rare. Instead I'm referring to sloped, shallow, incursions, such as the one on top (which constitutes the majority of the small incursions). They can have a very long length, yet only penetrate half a meter at most. This latter type is filtered based on the proximity to the borders of the relation instead. And to get the borders, I need the linestrings. See line 182 and 183 in the code

image

frodrigo: Ok, but now we have MP as valid polygon, we can just use a negative ST_Buffer to adress both cases, cannot ?

Famlam: Good point, hadn't thought of that method.

Famlam:
I also seem to have another issue. 
The "old" analyser - without multipolygon - took 82 seconds to run on netherlands_gelderland (using osm110)
The new analyser seems to get stuck on sql13 for a reaaaaally long time.

Nothing strange in the EXPLAIN:
```
['Nested Loop  (cost=0.15..304970.38 rows=2818 width=264)']
['  ->  Seq Scan on landusage  (cost=0.00..47037.88 rows=353688 width=192)']
['  ->  Index Scan using idx_mobilityway_linestring on mobility_ways  (cost=0.15..0.72 rows=1 width=104)']
['        Index Cond: (landusage.poly && linestring)']
["        Filter: ((((NOT (landusage.tags ? 'layer'::text)) AND (NOT (tags ? 'layer'::text))) OR ((landusage.tags ? 'layer'::text) AND (tags ? 'layer'::text) AND ((landusage.tags -> 'layer'::text) = (tags -> 'layer'::text)))) AND (((st_buffer(landusage.poly_proj, '-2'::double precision) && st_transform(linestring, 23032)) AND _st_crosses(st_buffer(landusage.poly_proj, '-2'::double precision), st_transform(linestring, 23032))) OR ((st_buffer(landusage.poly_proj, '-2'::double precision) ~ st_transform(linestring, 23032)) AND _st_contains(st_buffer(landusage.poly_proj, '-2'::double precision), st_transform(linestring, 23032)))))"]
```

Running EXPLAIN ANALYSE now...

frodrigo:
MP are big and probably hit lot of object. Intersection of large polygon can also be better with split : ST_Subdivide.

Famlam:
Thanks, I'll try implementing that!

Removing the buffer calculation from this query (i.e. moving ST_Buffer from sql13 to sql11) also saves a lot (852 to 123 seconds, see details) and only increases the runtime of sql11 by 44 seconds), but probably better results can be achieved with your suggestion. But good to know for i.e. osmosis_highway_vs_building and osmosis_relation_public_transport that it's probably a bad idea to nest ST_Intersects and ST_Buffer calls.
```
['Nested Loop  (cost=0.15..313792.50 rows=2818 width=264) (actual time=29.167..122717.934 rows=4247 loops=1)']
['  ->  Seq Scan on landusage  (cost=0.00..59392.88 rows=353688 width=192) (actual time=0.080..1866.846 rows=353688 loops=1)']
['  ->  Index Scan using idx_mobilityway_linestring on mobility_ways  (cost=0.15..0.71 rows=1 width=104) (actual time=0.334..0.340 rows=0 loops=353688)']
['        Index Cond: (landusage.poly && linestring)']
["        Filter: ((((NOT (landusage.tags ? 'layer'::text)) AND (NOT (tags ? 'layer'::text))) OR ((landusage.tags ? 'layer'::text) AND (tags ? 'layer'::text) AND ((landusage.tags -> 'layer'::text) = (tags -> 'layer'::text)))) AND (((landusage.poly_proj_buffer && st_transform(linestring, 23032)) AND _st_crosses(landusage.poly_proj_buffer, st_transform(linestring, 23032))) OR ((landusage.poly_proj_buffer ~ st_transform(linestring, 23032)) AND _st_contains(landusage.poly_proj_buffer, st_transform(linestring, 23032)))))"]
['        Rows Removed by Filter: 1']
['Planning Time: 1.282 ms']
['Execution Time: 122719.594 ms']
```

frodrigo:
> which means breaking down only the really big multipolygons

Yes make sens.

I agree, test are not required.

Famlam:
> I agree, test are not required.

Tests for ST_Subdivide that would require manual interaction to be tested removed 👍

frodrigo:
Thank you

PR #1877, opened 2023-06-01
Fix whitelist with NSI for Name_Script 

Famlam:
Fixes https://github.com/osm-fr/osmose-backend/issues/1830 also for names with spaces.

- Don't split by spaces anymore in the generic function; only do this in Name_UpperCase (fixes whitelisting names with spaces in Name_Script)
- Fix countries with a `-` in Name_Script
- Add a test, this time with language and country set

In addition:
- Allow partial matches in Name_Script (in case a brand has a local suffix)

### DIFF
--- a/plugins/Name_Script.py
+++ b/plugins/Name_Script.py
@@ -120,7 +120,7 @@ appropriate.'''),
 
         self.whitelist_names = set()
         if country:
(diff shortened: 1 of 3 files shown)


frodrigo:
Look good, thank ypyou.

PR #1878, opened 2023-06-01
Add support for `area` mapcss selectors

Famlam:
Add support for `area` selectors in mapcss by converting them into `way[area!=no]` and `relation[type=multipolygon]` rules.

Additionally:
- Updates the mapcss rules
- Removes `:tagged`, which if present in a rule always errors (since it's not defined anywhere else)
- Uses the module stablehash rather than duplicating the function
- Filters rules with anything else than `*`, `node`, `way`, `relation` (after `area` conversion)
- Fixes breaking when `:closed`/`:closed2` is used with a relation or node (or `area` after conversion, which is the use case here)

**Sources**:
- [`area` == `relation[type=multipolygon]` & `way`](https://github.com/JOSM/josm/blob/ed783461e7043ddf1ef587aeb3879209fcbb6e1f/src/org/openstreetmap/josm/gui/mappaint/mapcss/Selector.java#L796-L809)
- [`way` must be `area!=no`](https://github.com/JOSM/josm/blob/ed783461e7043ddf1ef587aeb3879209fcbb6e1f/src/org/openstreetmap/josm/gui/mappaint/mapcss/MapCSSStyleIndex.java#L145)
- [`:closed`/`:closed2` are false for all nodes and relations, except multipolygons](https://github.com/JOSM/josm/blob/c7ea6f33fa2ae3e929b8e96a042206b2bf64a184/src/org/openstreetmap/josm/gui/mappaint/mapcss/ConditionFactory.java#L718)


frodrigo:
If there is commits not directly linked to support `area` feature, please can you move it to an other PR to make this one simpler, please ?

Famlam:
> If there is commits not directly linked to support `area` feature, please can you move it to an other PR to make this one simpler, please ?

I've split off two of them into separate PRs while I continue testing this one. Will later rebase it as they're literally the same as the first two commits here.
Implementing `area` without `:closed` for relations will fail, see [this line](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss#L692), so they have to be linked; same for the test file. I hope it's fine that I only put them in separate commits?

There will also be another commit when I update the new mapcss files (after I finish testing)

Famlam:
Nevermind, managed to mess it up in Github, suddenly it lists 10 commits instead of 3 leftover commits. Will create a new PR as soon as I know whether I can have the 3 related commits combined. 

(Let's hope no external mapcss rules get updated in the meantime, because then it would become difficult to see what is caused by what)

frodrigo:
Thank you. Just rebase this branch now.

Famlam:
Rebasing left 10 commits, but reverting them left 1 mysterious commit. Guess I'll just create a new PR as soon as I've finished testing it on some 'real life' extracts (to make sure the placeholders still work)

Famlam:
Created a new PR without weird commits :)

PR #1883, opened 2023-06-02
Add support for `area` mapcss selectors

Famlam:
Add support for `area` selectors in mapcss by converting them into `way[area!=no]` and `relation[type=multipolygon]` rules.

Also fixes breaking when `:closed`/`:closed2` is used with a relation or node (or `area` after conversion, which is the use case here, see [this rule](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss#L692))

**Sources**:
- [`area` == `relation[type=multipolygon]` & `way`](https://github.com/JOSM/josm/blob/ed783461e7043ddf1ef587aeb3879209fcbb6e1f/src/org/openstreetmap/josm/gui/mappaint/mapcss/Selector.java#L796-L809)
- [`way` must be `area!=no`](https://github.com/JOSM/josm/blob/ed783461e7043ddf1ef587aeb3879209fcbb6e1f/src/org/openstreetmap/josm/gui/mappaint/mapcss/MapCSSStyleIndex.java#L145)
- [`:closed`/`:closed2` are false for all nodes and relations, except multipolygons](https://github.com/JOSM/josm/blob/c7ea6f33fa2ae3e929b8e96a042206b2bf64a184/src/org/openstreetmap/josm/gui/mappaint/mapcss/ConditionFactory.java#L718)

I did not observe any Python errors or bad results after running it on several countries. Only issue 1882, but that's unrelated to the changes here; it just popped up as the rule involved is no longer ignored but the error is elsewhere in the mapcss code.


mapcss/mapcss2osmose.py
```diff
@@ -71,6 +72,22 @@ def simple_selector_pseudo_class(t, c):
...
+def convert_area_selectors(t, c):
+    """
+    type = rule
+    Convert area* rules to way[area!=no]* + relation[type=multipolygon]*
+    """
+    areaselectors = list(filter(lambda selector: selector['simple_selectors'][0]['type_selector'] == 'area', t['selectors']))
+    for s in areaselectors:
+        relSelector = deepcopy(s)
+        relSelector['simple_selectors'][0]['type_selector'] = 'relation'
+        relSelector['simple_selectors'][0]['predicates'].append({'type': 'booleanExpression', 'operator': '=', 'operands': [{'type': 'valueExpression', 'operator': None, 'operands': [{'type': 'primaryExpression', 'derefered': False, 'value': {'type': 'osmtag', 'value': 'type'}}]}, {'type': 'valueExpression', 'operator': None, 'operands': [{'type': 'primaryExpression', 'derefered': False, 'value': {'type': 'osmtag', 'value': 'multipolygon'}}]}], 'selector_index': selector_index_map['arearule']}) # This is the output of an [type=multipolygon] selector with a mock selector_index
```
frodrigo, line 84: Hard coding this may make it hard to maintain ?

mapcss/mapcss2osmose.py
```diff
@@ -71,6 +72,22 @@ def simple_selector_pseudo_class(t, c):
     t['pseudo_class'] = list(filter(lambda p: not (p['not_class'] and p['pseudo_class'] in ('completely_downloaded', 'in-downloaded-area')), t['pseudo_class']))
     return t
 
+def convert_area_selectors(t, c):
+    """
+    type = rule
+    Convert area* rules to way[area!=no]* + relation[type=multipolygon]*
+    """
+    areaselectors = list(filter(lambda selector: selector['simple_selectors'][0]['type_selector'] == 'area', t['selectors']))
+    for s in areaselectors:
```
frodrigo, line 81: Does it works with multiples `area` ?

mapcss/mapcss2osmose.py
Famlam: I fully agree, but I could not find a way to soft-code it. Preferably one would just want `somefunction("[type=multipolygon]")` (or even extract it from `way[type=multipolygon] { throwWarning: "";}`)
The parser requires a ANTLR filestream, which is only defined for text files; even using an StringIO failed.

mapcss/mapcss2osmose.py
Famlam: I'll add a test

mapcss/mapcss2osmose.py
Famlam: (If you can think of a way, also the similar hardcoded rule of pseudo_class_righthandtraffic could be rewritten)

Famlam: Have an idea :)

Famlam: It works :)
(Let's do pseudo_class_righthandtraffic in a different PR)

Famlam:
For some reason the force pull from upstream/dev (to include the modifications from the other PR) has inverted the commit sequence for the addition of the mapcss test rules and the commit that regenerates the py files.
However, despite this, the regenerated py file does include the tests

frodrigo:
@Famlam it is ready for you ? You can move out some commits to other PR ?

Famlam:
It's ready to be reviewed, yes.
All of it is directly related to implementing `area` rules, so I thought it should all be in the same PR?

Of course you can ignore commit "Regenerating mapcss", that's literally just me calling mapcss/update.py after implementing the area rules.

Famlam:
If it helps, I purposefully split out the 4 different individual parts in separate commits to aid reviewing. So everything related to `area` is in commit 1 (and 5, see below). Everything related to `:closed` is in commit 2 (and 5, see below). Everything related to tests in commit 3. Updating mapcss in commit 4.

(Note that the python execution will crash if commit 2 is omitted: function `relation` does not have access to variable `nds`)

Only commit 5 is a bit different, as that addresses your previous review comment.

frodrigo:
Ok. Please can you also update the doc https://github.com/osm-fr/osmose-backend/blob/dev/doc/2-PluginMapCSS.md

Famlam:
Done!

frodrigo:
Thank you for all this work.

PR #1868, opened 2023-05-18
Simplify and speed up analyser highway_vs_buildings

Famlam:
Part 2 of https://github.com/osm-fr/osmose-backend/issues/1852#issuecomment-1541835203

#### The commits
- (commit 1) Use common table `highways` for the highways:
  - Reduces code duplication
  - Speeds up the analyser, most of the time (table below). No longer the factor 3 improvement it used to be in the first tests, probably because sql00 became a lot more complex since. (see old numbers [here](https://github.com/osm-fr/osmose-backend/pull/1820#issuecomment-1524158565))
- (commit 1) Remove twice-whitelisted `highway=elevator`
- (commit 2) Remove 100 nodes limit (see [comments here](https://github.com/osm-fr/osmose-backend/pull/1851/#issuecomment-1553024798) )

#### Regarding commit 1:
- verified that the output was the same
  - (Except `highway=pedestrian` as an area without area tag, as that's now filtered out by is_area)
  - Except some cases where multiple "valid" answers are possible
    - example: sql31 first returned way 317057792, now 317057793, which both connect to the same tree.
    - example: sql40 first had a marker [here](https://www.openstreetmap.org/note/new?lat=36.535489466707&lon=136.891955434029), now [here](https://www.openstreetmap.org/note/new?lat=36.536616623674&lon=136.890345809238), both locations are an overlap between the same ways (found similar occasions for sql51, which is similar)
    - Maybe these cases should have some optimization to ensure the output is constant, but that can be in a later PR (in my opinion). (E.g. some locations also changed between commit 1 and 2)
- speed of *sql00* (not the full analyser, just this query) on osm110:

| extract | before (s) | after (s) | faster? |
| --- | --- | --- | --- |
| japan_chubu | 72.4 | 64.9 | yes |
| japan_chubu (repeat 1) | 74.6 | 65.0 | yes |
| japan_chubu (repeat 2) | 77.6 | 64.7 | yes |
| netherlands_gelderland | 16.0 | 19.7 | no |
| netherlands_gelderland (repeat 1) | 16.8 | 13.3 | yes |
| netherlands_gelderland (repeat 2) | 17.8 | 10.4 | yes |
| netherlands_gelderland (repeat 3) | 16.9 | 13.6 | yes |
| jersey | 0.4 | 0.4 | - |
| sweden_gotland | 1.0 | 0.7 | yes |
| slovenia | 27.6 | 26.6 | yes |
| slovenia (repeat 1) | 18.9 | 17.1 | yes |
[...]

PR #1885, opened 2023-06-03
Reduce and clarify `waterway=stream` warnings

Famlam:
See https://github.com/osm-fr/osmose-backend/pull/1868#issuecomment-1565357011

For Canana-BC:
- Disable "Highway intersecting small water piece" when the waterway is a stream (due to excessive reports)

For the remainder of the world
- Change the text to include that `ford=*` could be a valid tag on the crossing. (Most tracks/paths etc. don't have bridges over small streams)

Feel free to reject if you wish to wait for user reports instead.

### DIFF
--- a/analysers/analyser_osmosis_highway_vs_building.py
+++ b/analysers/analyser_osmosis_highway_vs_building.py
@@ -325,6 +325,7 @@ WHERE
     highway.layer = water.layer AND
     NOT highway.onwater AND
     (nodes.id IS NULL OR NOT nodes.tags?'ford')
+    {2} -- used to exclude specific types for specific countries
 ORDER BY
     highway.id,
     water.id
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_highway_vs_building.py
```diff
@@ -503,7 +505,10 @@ def analyser_osmosis_full(self):
         self.run(sql30.format("", ""), self.callback30)
         self.run(sql31.format("", ""), self.callback31)
         self.run(sql32.format("", ""), self.callback32)
-        self.run(sql40.format("", ""), self.callback40)
+        if self.country and self.country == "CA-BC": # Too many results
```
frodrigo, line 508: Use `startswith` to be region split ready.

Famlam: Good point, done!

frodrigo:
Thank you

PR #1891, opened 2023-06-05
Avoid dealing with very large multi-polygons

frodrigo:
Very large multi-poly like Lake Nubia, more than 1 300 000 nodes, raise issue while passing on postgis geometrical function.

Just skip such large geometry.


Famlam:
This is not possible by using the ST_Subdivide approach?

frodrigo:
> This is not possible by using the ST_Subdivide approach?

The lake is so big, ST_Buffer take 10min on it. ST_Subdivide no idea, server fail after 100 GB produced in 5 hours... We are fight servers issues for 10 days because of this.

Bbox index of such big geom will intersect with so much features.

Famlam:
I understand :)

Issue #1887, opened 2023-06-03
analyser_osmosis_highway_area_access item=2130, class=2 produce duplicate issues


frodrigo:
At least item=2130, class=2  produce non distinguishable osmose issues in the frontend. Each issue overwrite the previous for the same node.
The result should contain a subclass, on fo each vehicle type.

https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_osmosis_highway_area_access.py#L221

Maybe a better way to fix it is to make only one report for the node ?

frodrigo:
Required for https://github.com/osm-fr/osmose-frontend/pull/462

Famlam:
I wouldn't know a way to keep this analyser flexible 
(i e. avoiding  `(3 lines for emergency) OR (3 lines for motorcycle) OR ( 3 lines for hgv) OR ...`) so I think subclasses are the simplest solution?

Probably there should be a generic test that no duplicates exist, since this would get unnoticed on the backend?

frodrigo:
> I wouldn't know a way to keep this analyser flexible (i e. avoiding `(3 lines for emergency) OR (3 lines for motorcycle) OR ( 3 lines for hgv) OR ...`) so I think subclasses are the simplest solution?

Yes, I'm ok for that.

Famlam:
Which parts exactly have to be unique for the issue ID hash?
(There is of course always a small possibility of two different errors getting the same ID, given the huge size of OSM and the large number of analysers)

frodrigo:
UUID is computed here https://github.com/osm-fr/osmose-frontend/blob/master/control/update_utils.py#L265

PR #1893, opened 2023-06-08
Add subclass to class 2 elems

Famlam:
See #1887 
Add a unique subclass. Not sure if node_id and way_id are necessary, but it doesn't hurt either I guess to have some randomness.


frodrigo:
Thank you.

Issue #1904, opened 2023-06-13
 Exception: class 1 already present as item 9020


Famlam:
For any Dutch extract, the following message shows up.
Item 9020 is indeed associated with the Dutch rules, but it isn't used anywhere else as far as I could see (and the rules are constructed the same way as JOSM). Hence I don't really understand why it shows the error?

```
2023-06-13 18:27:33   error: error on analyse sax...
2023-06-13 18:27:33     Traceback (most recent call last):
2023-06-13 18:27:33       File "/data/project/osmose/backend/./osmose_run.py", line 244, in execc
2023-06-13 18:27:33         with obj(analyser_conf, logger.sub()) as analyser_obj:
2023-06-13 18:27:33       File "/data/project/osmose/backend/analysers/analyser_sax.py", line 42, in __init__
2023-06-13 18:27:33         self._init_plugins(plugins)
2023-06-13 18:27:33       File "/data/project/osmose/backend/analysers/analyser_sax.py", line 463, in _init_plugins
2023-06-13 18:27:33         raise Exception("class {0} already present as item {1}".format(cl, self._Err[cl]['item']))
2023-06-13 18:27:33     Exception: class 1 already present as item 9020
```

https://buildbot.osmose.openstreetmap.fr/#/builders/565/builds/979/steps/3/logs/stdio

Famlam:
It seems to have worked fine until two days ago.
https://buildbot.osmose.openstreetmap.fr/#/builders/565

frodrigo:
The class should be unique for the analyse. Here the analyser is sax. So the class should be unique for all plugins.

For manual class number we usually use number like 90201.

Famlam:
I understand, but it's a mapcss python file. The class numbers are generated automatically (and haven't changed in months)

frodrigo:
Maybe something else add the class `1`.

This class number come from item_map.py. It was a bad move to set this number here.

Class `1` was introduced in indoor plguin.

frodrigo:
https://github.com/osm-fr/osmose-backend/commit/74c73012d077a5b5366d35fc2ca2d257298797b0#diff-8f4a60d6a610930798736bfedf269112cfb944d56e7158333229650ce8268289R114

Famlam:
Updating mapcss rules after implementing a new feature is more dangerous than I thought 🙈

Famlam:
> Once deployed we can renumber the issues in the frontend to keep history and false positives.

[...]

PR #1905, opened 2023-06-13
Fix class numbers mapcss

Famlam:
Updates the Dutch mapcss rule classes by adding 90200 to each (unfortunately resets the false positives I think?)

Updates indoor validator rules by adding a class number. (The other classes started at 50, so I continued that numbering). Interesting that it managed to compile actually, since there was no `item` set either.

See https://github.com/osm-fr/osmose-backend/issues/1904


frodrigo:
Ok, thank you. Once deployed we can renumber the issues in the frontend to keep history and false positives.

Issue #1915, opened 2023-06-22
Split Turkey


frodrigo:
Split Turkey by admin_level=3

```
@id	@type	name:en	cg_ref
1726294	relation	Black Sea Region	TR-2
2092903	relation	Mediterranean Region	TR-6
2094154	relation	Eastern Anatolia Region	TR-4
2094155	relation	Southeastern Anatolia Region	TR-7
2094170	relation	Aegean Region	TR-1
2094193	relation	Marmara Region	TR-5
2094194	relation	Central Anatolia Region	TR-3
```

```
[out:csv(::id,::type,"name:en","cg_ref")];
{{geocodeArea:turkey}}->.searchArea;
relation[type=boundary][boundary=administrative][admin_level=3](area.searchArea);
out body geom;
```

frodrigo:
```python
tr_part('black_sea', 1726294, 'TR-2')
tr_part('mediterranean', 2092903, 'TR-6')
tr_part('eastern_anatolia', 2094154, 'TR-4')
tr_part('southeastern_anatolia', 2094155, 'TR-7')
tr_part('aegean', 2094170, 'TR-1')
tr_part('marmara', 2094193, 'TR-5')
tr_part('central_anatolia', 2094194, 'TR-3')
```

PR #1918, opened 2023-06-23
Split Turkey #1915

frodrigo:


jocelynj:
Diffs are enabled on http://download.openstreetmap.fr/replication/europe/turkey/

Issue #1912, opened 2023-06-20
Item 4130 - Marker jumps around, changing issue id


Famlam:
The screenshots are from: https://osmose.openstreetmap.fr/nl/map/#zoom=16&lat=51.83874&lon=5.86743&item=4130&level=1%2C2%2C3&tags=&fixable

Item 4130 / class 1

The marker jumps around and (due to this) the issue ID changes all the time. Hence, if I want to mark this one as a false positive, I would have to mark it a dozen of times. 

For example,
- on 12-6 the issue ID was https://osmose.openstreetmap.fr/nl/issue/15da88fd-869f-8342-1c1e-690314061fd1
- on 20-6 the issue ID was https://osmose.openstreetmap.fr/nl/issue/2132165d-c7a3-8a5d-8908-f3d6a195805a

![image](https://github.com/osm-fr/osmose-backend/assets/1315520/28089086-1a77-4d4f-9b51-23fd54aa399b)
![image](https://github.com/osm-fr/osmose-backend/assets/1315520/5c752d38-487b-4861-ade6-7b7d10290017)
![image](https://github.com/osm-fr/osmose-backend/assets/1315520/05de0f2f-644a-4b41-bab7-5b5ecdd7ed22)

I guess there should be an `ORDER BY id` or `ORDER BY tid` at the end of the query?

frodrigo:
Not sure. Because sub query already ordered. It is not required by SQL, but I think postgres keep the order. on the the group by.

We can try to refactor all order by to the main query.

frodrigo:
@Famlam do you think it is ok now ?

Famlam:
I'll check tonight. For some reason I can't find the marker for the item in the first post

Famlam:
Seems pretty stable :)

PR #1922, opened 2023-06-25
(Attempt to) stop jumping markers for duplicate wiki pages

Famlam:
Always use the same element to show the marker: #1912

Also removes the need to propagate `id, type` together with `tid`

However, seeing if this stops the 'jumping markers' probably has to be observed "live".

### DIFF
--- a/analysers/analyser_osmosis_wikipedia.py
+++ b/analysers/analyser_osmosis_wikipedia.py
@@ -26,14 +26,12 @@ from .Analyser_Osmosis import Analyser_Osmosis
 sql10 = """
 SELECT
   (array_agg(tid))[1:10],
-  ST_AsText(any_locate((array_agg(type))[1], (array_agg(id))[1])),
+  ST_AsText(any_locate(substring(MIN(tid), 1, 1), substring(MIN(tid), 2)::bigint)),
   w
 FROM ((
   SELECT
     tags->'wikipedia' AS w,
-    'N' || id AS tid,
-    'N' AS type,
-    id
+    'N' || id AS tid
   FROM
     nodes
   WHERE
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_wikipedia.py
```diff
@@ -26,44 +26,34 @@
 sql10 = """
 SELECT
   (array_agg(tid))[1:10],
-  ST_AsText(any_locate((array_agg(type))[1], (array_agg(id))[1])),
+  ST_AsText(any_locate(substring(MIN(tid), 1, 1), substring(MIN(tid), 2)::bigint)),
```
frodrigo, line 29: It is not the min as expected, because it compare on strings.

Famlam: Ah, true, n56789 would be after n100000. 
However, my goal was just to have a stable output (as long as no object is added/removed) to get rid of the jumping markers, so whether it's alphabetically the first or numerically the first should not make a difference, right?

frodrigo: I think older is more stable.
It just about concatenate type later.

Famlam: Not sure how to get the oldest element.
If I add a new way today, the ID will be [around 1185000000](https://www.openstreetmap.org/way/1185000000). A [node with the same ID](https://www.openstreetmap.org/node/1185000000) would be 12 years old. So comparison by id will not retrieve the oldest element

frodrigo: Just order by type and id will do the job. The older by type is sufficient, isn't it ?

Famlam: I'll have another look at a later moment.

Just calling ORDER BY won't work, since I can't add id or type to GROUP BY or it'll be filtered out. Probably have to modify the COUNT

Famlam: Nope, also with `COUNT(w)` it doesn't work. Probably I misunderstood what you meant @frodrigo, where would I have to call the ORDER BY type, id?

Famlam: Actually, the issue ID changes anyway if elements are added or deleted, right? So is it really needed to be the same element every time, also when elements are added? (Of course, the jumping when there's no changes must be fixed)

frodrigo:
Closed in favour of #1958

PR #1924, opened 2023-06-27
Fix mapcss `any(...)` function

Famlam:
In mapcss_lib the `any` function of mapcss is defined as `any_`, probably to avoid overwriting the python `any` function (which has a completely different meaning and is also used in the mapcss python code)

However, mapcss2osmose just converts the `any` function to `mapcss.any`, therefore throwing `AttributeError: module 'modules.mapcss_lib' has no attribute 'any'` on execution.

In addition, calls like `tag("non-existing key")` don't return `None`, but `'None'` (type `str_value_`), therefore failing the `item is not None` check in case of `any(tag("non-existing key"), ...)` and returning bogus results. `str_value_`-objects store the property `none` if the value was originally a true `None`, hence check that too. 

Tests are added too of course :)

Note that this patch is needed before the external mapcss rules can be updated. I started using the `any` function in the Dutch rules, that's why I noticed this bug.

### DIFF
--- a/mapcss/mapcss2osmose.py
+++ b/mapcss/mapcss2osmose.py
@@ -425,6 +425,7 @@ def functionExpression_runtime(t, c):
         t['name'] = (
             "keys.__contains__" if t['name'] == 'has_tag_key' else
             "mapcss.list_" if t['name'] == 'list' else
+            "mapcss.any_" if t['name'] == 'any' else
             "mapcss.round_" if t['name'] == 'round' else
             "mapcss." + t['name']
         )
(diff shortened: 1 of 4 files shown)


Famlam:
p.s. the current mapcss `any` function implementation will also give an error if no instances are not-None, e.g. `any(tag("nonexistent1"), tag("nonexistent2"))`. The easiest solution is probably to catch the `StopIteration` exception and output `None_value` instead in that case. This matches the JOSM result of simply outputting `null`. (Of course, such rules are broken and should be fixed by the mapcss maintainer). However, this is a separate bug, so I'll fix that in a new PR after this one is merged, unless you prefer me to fix it here.

frodrigo:
If you no fixing the other issue now, please open an issue.

Thank you.

Famlam:
> If you no fixing the other issue now, please open an issue.

No worries, I had the PR ready on my computer. Just waited pushing it because it would conflict (and resolving conflicts is pretty time-consuming)

PR #1927, opened 2023-06-29
Upgrade openjdk from version 11 to 17

carlbalmer:
Docker build fails with `Package 'openjdk-11-jre-headless' has no installation candidate` due to the python:3.9 tag being updated to debian bookworm and bookworm no longer supporting openjdk-11.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9
+FROM python:3.9-bullseye
 
 RUN apt-get update && \
     apt-get -y dist-upgrade && \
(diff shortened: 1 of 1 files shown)


carlbalmer:
An alternative solution would be to downgrade back to debian bullseye. (Se previous commit [51de07a](https://github.com/osm-fr/osmose-backend/pull/1927/commits/51de07ab5e21b16144bcb4d26b0ff624382da307))

frodrigo:
It is ok to switch to bookworm.

frodrigo:
Thank you.

jocelynj:
Shouldn't we also switch to python3.11, which is the version delivered with debian bookworm?

frodrigo:
> Shouldn't we also switch to python3.11, which is the version delivered with debian bookworm?

Yes we also have to update python. But not required has docker image provided by Python project are available in any version.

PR #1933, opened 2023-06-30
Check if mapcss functions exist

### LINKED ISSUE #1925: Mapcss should auto-remove or warn on compile about unknown functions
## Example
Have the following mapcss code, containing two functions that we don't support yet. These could be functions that JOSM recently added and we don't support yet (such as `parent_way_angle`) or just typos.

```
node[abcdefghijkl("3") == 2] {
  throwWarning: tr("{0}", mnopqrstuvwxyz("1"));
}
```

Compiling gives:
```
    def init(self, logger):
        super().init(logger)
        tags = capture_tags = {} # noqa
>>>     self.errors[1] = self.def_class(item = 0, level = 3, tags = [], title = mapcss.tr('{0}', mapcss.mnopqrstuvwxyz('1')))



    def node(self, data, tags):
        capture_tags = {}
        keys = tags.keys()
        err = []


        # node[abcdefghijkl("3")==2]
        if True:
            match = False
            if not match:
                capture_tags = {}
>>>             try: match = ((mapcss.abcdefghijkl('3') == 2))
                except mapcss.RuleAbort: pass
            if match:
                # throwWarning:tr("{0}",mnopqrstuvwxyz("1"))
>>>             err.append({'class': 1, 'subclass': 114025324, 'text': mapcss.tr('{0}', mapcss.mnopqrstuvwxyz('1'))})

        return err
```
Note that two out of the three function calls are outside `try/catch` blocks. One of them is even in the init (since there is no `group` property, the `throwWarning` is copied to the title)

## Will it be detected
You will *not* notice this during compilation
If the rule (like the above) *does not* have a `group` property and the error is in the positive match code (e.g. throwWarning, fixAdd, ...) it will be detected by the tests. Also if the rule *does* have a `group` property and the error has an `assertMatch`, it will be detected by the tests
However, if the rule *does* have a `group` property and the error does *not* have an `assertMatch` (or `-osmoseAssertMatchWithContext`), it'll **only be detected during runtime**

## Most simple solution:
Import `mapcss_lib` as module in `mapcss2osmose`, then have function `functionExpression_rule_flags` check if there is a function with said name. If not, flag it. Rules flagged will be replaced by `# Part of rule not implemented`
(As a sidenote: it would be nice if that text would also show what part is not implemented, for debugging purposes)

It wouldn't be perfect (you can still call a function with the wrong number of arguments), but maybe an version with the signature could do this trick. https://docs.python.org/3/library/inspect.html#inspect.signature

EDIT: probably this is an impossible idea, due to the function rewrites in functionExpression_runtime (which is executed long after functionExpression_rule_flags). The code I have now is:
```py

from . import mapcss_lib
from inspect import signature, Parameter

[...]

def functionExpression_rule_flags(t, c):
    """
    type = functionExpression
    flag unsupported functions
    """
    if not t['name'] in dir(mapcss_lib):
        c['flags'].append(t['name'])
        print(t['name'])
    else:
        sig = signature(getattr(mapcss_lib, t['name']))
        argcount = len(sig.parameters)
        has_vararg = any(map(lambda p: p.kind == Parameter.VAR_POSITIONAL, sig.parameters.values()))
        if has_vararg:
            argcount = argcount - 1
        if len(t['params']) < argcount or (not has_vararg and len(t['params']) > argcount):
            c['flags'].append(t['name'])
            print("{name} - min_arg: {argcnt} - variable: {var}".format(name=t['name'], argcnt=str(argcount), var=str(has_vararg)))
    return t
```
It may still be possible to check the functions in `functionExpression_runtime`. If an unknown function is encountered there, it's probably too late for a flag, but we can still throw an error (alerting the person who updates mapcss that there's an unsupported function being used). That person can then add it to the hardcoded list in `functionExpression_rule_flags` and run the update script again. Please let me know if this would be preferred; if not, please close this issue; I'll just update the hardcoded list then.

Famlam:
Check if mapcss functions exist and have the right number of arguments
Raises a NotImplementedError during compilation, rather than an ArgumentError/TypeError during runtime on the server

Fixes #1925

### DIFF
--- a/mapcss/mapcss2osmose.py
+++ b/mapcss/mapcss2osmose.py
@@ -443,7 +443,7 @@ def checkValidFunction(fn_name, num_params):
     if fn_name[0:7] != "mapcss.":
         return
     if not fn_name[7:] in dir(mapcss_lib):
-        raise NotImplementedError("Undefined function '{0}'. Blacklist or implement to avoid runtime errors".format(fn_name))
+        raise NotImplementedError("Undefined function '{0}'. Blacklist or implement to avoid errors".format(fn_name))
     else:
         sig = signature(getattr(mapcss_lib, fn_name[7:]))
         argcount = len(sig.parameters) # includes optional arguments
@@ -452,7 +452,7 @@ def checkValidFunction(fn_name, num_params):
         if has_vararg:
             argcount = argcount - 1 # *args can be zero-length
         if num_params < argcount - num_optional_arg or (not has_vararg and num_params > argcount):
-            raise NotImplementedError("Undefined function '{0}' with {1} arguments. Blacklist or implement to avoid runtime errors".format(fn_name, num_params))
+            raise NotImplementedError("Undefined function '{0}' with {1} arguments. Blacklist or implement to avoid errors".format(fn_name, num_params))
 
 
 rewrite_rules_clean = [
--- a/mapcss/mapcss2osmose.py
+++ b/mapcss/mapcss2osmose.py
@@ -13,6 +13,8 @@ from .generated.MapCSSParser import MapCSSParser
 from .MapCSSListenerL import MapCSSListenerL
 from typing import Dict, List, Set, Optional
 from copy import deepcopy
+from . import mapcss_lib
+from inspect import signature, Parameter
 
 
 # Clean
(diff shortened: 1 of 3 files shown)


frodrigo:
Please, can you rebase this PR.

mapcss/mapcss2osmose.py
```diff
@@ -429,8 +431,30 @@ def functionExpression_runtime(t, c):
...
     return t
 
+def checkValidFunction(fn_name, num_params):
+    """
+    Tests if mapcss.* functions actually exist, throws a compile error otherwise
+    """
+    if fn_name[0:7] != "mapcss.":
+        return
+    if not fn_name[7:] in dir(mapcss_lib):
+        raise NotImplementedError("Undefined function '{0}'. Blacklist or implement to avoid runtime errors".format(fn_name))
```
frodrigo, line 446: Message can be confusing as raised at compile time.

```suggestion
        raise NotImplementedError("Undefined function '{0}'. Blacklist or implement to avoid errors".format(fn_name))
```

mapcss/mapcss2osmose.py
```diff
@@ -429,8 +431,30 @@ def functionExpression_runtime(t, c):
...
+    else:
+        sig = signature(getattr(mapcss_lib, fn_name[7:]))
+        argcount = len(sig.parameters) # includes optional arguments
+        has_vararg = any(map(lambda p: p.kind == Parameter.VAR_POSITIONAL, sig.parameters.values()))
+        num_optional_arg = len(set(filter(lambda p: p.default != Parameter.empty, sig.parameters.values())))
+        if has_vararg:
+            argcount = argcount - 1 # *args can be zero-length
+        if num_params < argcount - num_optional_arg or (not has_vararg and num_params > argcount):
+            raise NotImplementedError("Undefined function '{0}' with {1} arguments. Blacklist or implement to avoid runtime errors".format(fn_name, num_params))
```
frodrigo, line 455:
```suggestion
            raise NotImplementedError("Undefined function '{0}' with {1} arguments. Blacklist or implement to avoid errors".format(fn_name, num_params))
```

frodrigo:
Thank you

PR #1937, opened 2023-07-03
Remove unneeded blacklist

Famlam:
Remove unneeded blacklist that became obsolete as the buggy rule was removed in JOSM.

Didn't update item_map.py fully to avoid hardcoding issue 1934's bug


frodrigo:
This PR can be merger as is now?

Famlam:
Correct

Issue #1873, opened 2023-05-29
Extend polygon_small to leisure=park


Famlam:
@Marc-marc-marc 
the class about "too small landduse" may also be extented to include landuse-like object for exl leisure=garden https://www.openstreetmap.org/way/1155425895 (it's of course not a garden but a small cemetery with plaques and cremation urns) 

@famlam
I'm not in favor of leisure=garden in the small polygon analyser as there's no common limit on i.e. residential gardens, which area also part of leisure=garden. If I look in my area (if they were mapped) they can be as small as a 200x30 cm long strip along houses.
However, if read your text as if leisure=garden is just an example of a "landuse-like object": for instance having a size-filter for leisure=park sounds very reasonable to me, plenty of abuse it seems (examples: https://overpass-turbo.eu/s/1vuB ). We just have to find a good cutoff size to 1. exclude true parks, and 2. prevent flooding the map (it's 10x more popular than i.e. village_green, so if the bad-use rate is the same...)
Perhaps it's the easiest to create an issue on the issue tracker for this, if you agree?

@Marc-marc-marc 
2m x 0,3m a garden along house ? 0.3m ok, 2m look like very strange for me (but maybe a house of 5m, half used for someting else... ok that's only 0,6m2... hard so find a usefull limit
for a park (like the main issue), i'm in favor of starting with a very very conservative value for ex 1m2 or 10m2 to detect indisputable cases... and then gradually raise up
lenght<10 https://www.openstreetmap.org/way/31583825

@famlam
That's extremely conservative, and may even give people the impression that larger than 10m2 is probably alright. I'd start with 100m2 (10x10m "parks"). Or at least 25m2 (5x5m2, barely enough to sit on with a single group of friends). Also, when would you raise it up? There's probably areas where there's no-one using Osmose.
Anyway, I ran Overpass (using the "all parks are square assumption")
100m2 (40m perimeter) gives just 14873 detections worldwide
25m2 (20m perimeter) gives only 2503 detections worldwide
10m2 (simplified to 13m perimeter) gives 906 detections worldwide

Famlam:
@Marc-marc-marc This has been deployed, per your request with a very conservative starting value.
https://osmose.openstreetmap.fr/nl/issues/open?item=3100&source=&class=5&username=&bbox=

[...]

PR #1944, opened 2023-07-06
Detect (too) small parks

Famlam:
Uses the extremely conservative starting value of 10m2 (e.g. just over 3m wide if round or square) as preferred by Marc. Should give less than 1000 detections worldwide. To be increased as soon as feasible (in discussion of course).

See #1873

@Marc-marc-marc also looking forward to your review :)

p.s.: only ran tests on 1 country to ensure it didn't give bad errors.


frodrigo:
Ok. Let's try like this.

PR #1948, opened 2023-07-08
Fix looped drivethroughs

### LINKED ISSUE #1946: Dupplicate report on drivethroughs loop
osmosis_highway_deadend drivethroughs `service=drive-through` (item=1210 class=5 france_rhone_alpes_isere osmosis_highway_deadend) reports duplicate issue on loop. Duplicate report is one issue to the frontend end. But it should report nothing as the end are not dead and (it is a loop).

Secondary issue. As it is a loop, there is entrance / exits at other nodes.

This ways is in this case
https://www.openstreetmap.org/way/1186289900

Maybe the only case world wide.

https://buildbot.osmose.openstreetmap.fr/#/builders/580/builds/1042/steps/3/logs/stdio

Famlam:
If a drivethrough is a loop (and not a floating way without connections - which would be detected by a different analyser), then there must be a way out already. This way out is not necessarily connected to the same way as the first=final node of the loop. Hence, exclude such cases.
Also adds a test case.

Resolves #1946
Fixes https://www.openstreetmap.org/way/965343808, https://www.openstreetmap.org/way/196851026, https://www.openstreetmap.org/way/1186289900

p.s.: I can also use is_polygon if preferred? (Downside: it's more sensitive to invalid geometries).

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -235,6 +235,7 @@ WHERE
   drivethroughs.tags?'service' AND
   drivethroughs.tags->'service' = 'drive-through' AND
   NOT drivethroughs.is_oneway AND
+  drivethroughs.nodes[1] != drivethroughs.nodes[array_length(drivethroughs.nodes,1)] AND
   other_highways.id IS NULL
 """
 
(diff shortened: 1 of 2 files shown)


frodrigo:
> p.s.: I can also use is_polygon if preferred? (Downside: it's more sensitive to invalid geometries).

No as the intent is not a polygon.

Thank you.

Issue #1947, opened 2023-07-07
Polygon on border marked as "Should be polygon"


Famlam:
See https://osmose.openstreetmap.fr/nl/issue/cbdb8e5a-6eab-0cc7-c7cb-1d58b018c098
Item 1170 / class 4

This polygon (https://www.openstreetmap.org/way/1059958114) is located on the border of the Dutch province of Zeeland, mostly located outside the province border. It seems this is a false positive related to the border, not to the way itself.

Shouldn't this one have been filtered out already? I thought ways which crossed the extract border were excluded?

frodrigo:
> Shouldn't this one have been filtered out already? I thought ways which crossed the extract border were excluded?

No issue marker outside the border are excluded. But here is depend where the marker is set.

Famlam:
Ah ok. 
I'll check (soon) whether this is a rare case (in which case I'll mark it as false positive and just close the issue) or whether it's more common. 
In the latter case, probably we should just check that `node[1] != node[last]`. That should also work for invalid (due to being incomplete) polygons.

Famlam:
It appears to be a relatively common issue. For example, the province of Zeeland (from the initial post) has two:
https://osmose.openstreetmap.fr/nl/issue/cbdb8e5a-6eab-0cc7-c7cb-1d58b018c098
https://osmose.openstreetmap.fr/nl/issue/8a989242-5c8f-ea0d-a679-9ac9e7e8d210

An example from Chubu:
https://osmose.openstreetmap.fr/nl/issue/24b70783-3872-83b0-3071-9672416b9c51

Venezuela & surrounding have a lot:
https://osmose.openstreetmap.fr/nl/map/#zoom=8&lat=2.695&lon=-64.132&item=1170&level=1&tags=&fixable=&class=4
(Zoom out a bit to the right to find many more for Brazil)

The trick to find them is to add the following line to the SQL query:
`ways.nodes[1] = ways.nodes[array_length(ways.nodes,1)] AND NOT ST_IsClosed(ways.linestring)`
The nodes id check will work if not all nodes are downloaded. St_IsClosed does not work if not all nodes are downloaded. Hence, the difference between the two are the bad items.

Since there is already an "invalid polygon" check in another analyser, I suspect it's fine to add
`ways.nodes[1] != ways.nodes[array_length(ways.nodes,1)]` 
to this analyser to exclude such cases

Famlam:
Are both PR's already live?
[...]

PR #1951, opened 2023-07-09
Exclude partially downloaded, closed ways from "Should be polygon"-check

Famlam:
See #1947

This PR excludes partially downloaded, closed ways from the "should be polygon" check. This occurs quite frequently for larger landuses near the extract borders, see the issue for examples.

In addition it removes `leisure=user` and `leisure=defined`, which don't exist in OSM and are very likely just a copy-paste mistake (because the table on the wiki ends with `user defined`)

On top of this, as a positive side effect, the PR also removes the very strange warnings for:
- `building=*` with an `attraction=roller_coaster` tag. 
For example: https://osmose.openstreetmap.fr/nl/map/#zoom=18&lat=51.64769&lon=5.051999&item=1170&level=1&tags=&fixable= . 
It's a strange way of tagging it (which is used for 322 out of 2503 rollercoaster ways), but the warning didn't make sense as it's a polygon. (Note: `attraction=roller_coaster` is whitelisted [in ImportDatabase_Prepare](https://github.com/osm-fr/osmose-backend/blob/dev/osmosis/ImportDatabase_Prepare.sql#L18))
- invalid polygons. These are already warned for by analyser_osmosis_polygon, and that warning is better for such cases than "Should be polygon"

### DIFF
--- a/analysers/analyser_osmosis_relation_multipolygon.py
+++ b/analysers/analyser_osmosis_relation_multipolygon.py
@@ -183,12 +183,13 @@ WHERE
         ways.tags?'landuse' OR
         (ways.tags?'natural' AND ways.tags->'natural' in ('wood', 'scrub', 'heath', 'moor', 'grassland', 'fell', 'bare_rock', 'scree', 'shingle', 'sand', 'mud', 'water', 'wetland', 'glacier', 'bay', 'beach', 'hot_spring', 'rock', 'stone', 'sinkhole')) OR
         (ways.tags?'waterway' AND ways.tags->'waterway' in ('boatyard', 'dock', 'riverbank', 'fuel')) OR
-        (ways.tags?'leisure' AND ways.tags->'leisure' in ('adult_gaming_centre', 'amusement_arcade', 'beach_resort', 'bandstand', 'bird_hide', 'common', 'dance', 'dog_park', 'firepit', 'fishing', 'fitness_centre', 'garden', 'golf_course', 'hackerspace', 'horse_riding', 'ice_rink', 'marina', 'miniature_golf', 'nature_reserve', 'park', 'picnic_table', 'pitch', 'playground', 'sports_centre', 'stadium', 'summer_camp', 'swimming_area', 'swimming_pool', 'water_park', 'wildlife_hide', 'user', 'defined')) OR
+        (ways.tags?'leisure' AND ways.tags->'leisure' in ('adult_gaming_centre', 'amusement_arcade', 'beach_resort', 'bandstand', 'bird_hide', 'common', 'dance', 'dog_park', 'firepit', 'fishing', 'fitness_centre', 'garden', 'golf_course', 'hackerspace', 'horse_riding', 'ice_rink', 'marina', 'miniature_golf', 'nature_reserve', 'park', 'picnic_table', 'pitch', 'playground', 'sports_centre', 'stadium', 'summer_camp', 'swimming_area', 'swimming_pool', 'water_park', 'wildlife_hide')) OR
         (ways.tags?'amenity' AND ways.tags->'amenity' in ('bar', 'biergarten', 'cafe', 'fast_food', 'food_court', 'ice_cream', 'pub', 'restaurant', 'college', 'kindergarten', 'library', 'public_bookcase', 'school', 'music_school', 'driving_school', 'language_school', 'university', 'bicycle_repair_station', 'bicycle_rental', 'boat_sharing', 'bus_station', 'car_rental', 'car_sharing', 'car_wash', 'ferry_terminal', 'fuel', 'motorcycle_parking', 'parking', 'parking_space', 'taxi', 'bank', 'baby_hatch', 'clinic', 'dentist', 'doctors', 'hospital', 'nursing_home', 'pharmacy', 'social_facility', 'veterinary', 'blood_donation', 'arts_centre', 'brothel', 'casino', 'cinema', 'community_centre', 'fountain', 'gambling', 'nightclub', 'planetarium', 'social_centre', 'studio', 'swingerclub', 'theatre', 'animal_boarding', 'animal_shelter', 'courthouse', 'coworking_space', 'crematorium', 'crypt', 'dive_centre', 'dojo', 'embassy', 'fire_station', 'firepit', 'game_feeding', 'grave_yard', 'gym', 'hunting_stand', 'internet_cafe', 'kneipp_water_cure', 'marketplace', 'place_of_worship', 'police', 'post_office', 'prison', 'public_building', 'ranger_station', 'recycling', 'rescue_station', 'sauna', 'shelter', 'shower', 'toilets', 'townhall', 'waste_transfer_station')) OR
         ways.tags?'building'
     ) AND
     ways.linestring IS NOT NULL AND
     NOT ways.is_polygon AND
+    ST_IsValid(ways.linestring) AND -- avoid confusing warnings for invalid ways (checked elsewhere)
     relation_members.member_id IS NULL
 """
 


analysers/analyser_osmosis_relation_multipolygon.py
```diff
@@ -183,12 +183,13 @@
...
         ways.tags?'building'
     ) AND
     ways.linestring IS NOT NULL AND
     NOT ways.is_polygon AND
+    ways.nodes[1] != ways.nodes[array_length(ways.nodes,1)] AND -- See #1947, excludes partially downloaded closed polygons
```
frodrigo, line 192: Sorry. But I do not understand the trick here.

Famlam: For a (valid) polygon where some nodes aren't downloaded,  `is_polygon` (and even ST_IsClosed) all return false, even though they would return true if the way would've been fully downloaded.
Hence, the border region in i.e. Venezuela, Brazil etc (see issue) is cluttered with false positive markers.

However, for the nodes inside the borders of the extract, checking that the first node (id) isn't the last node (id) works very well.

Although I could technically LEFT JOIN on all nodes to verify they exist, and ignore any way where a node isn't downloaded this would also remove correct detections for partially downloaded (invalid) ways. (In addition to being slower).

In addition, adding the first node != last node check removes a lot of other false positives or redundant detections. And for this analyzer it doesn't affect what's detected.

frodrigo: Not sure about that. We use the `enableKeepPartialLinestring` osmosis option (I added it myself https://github.com/openstreetmap/osmosis/pull/121/files) to avoid break/lost linestring on border.

Not sure. But  is_polygon / ST_IsClosed should be valid, I guess. Maybe the issue is here.

Famlam: Not sure where I can modify the behavior of ST_IsClosed?

frodrigo: No, `ST_IsClosed` is a Postgis function.

Famlam: I can file an issue at openstreetmap/osmosis if you like 👍.

Would you like me to change this PR to use `ST_IsClosed` to fix the other issues already (and automatically fix the original issue when osmosis is fixed)? With "the other issues" I'm referring to the ones listed under _'On top of this, as a positive side effect, the PR also removes the very strange warnings for: [...]'_ in the first post

(The [definition of is_polygon](https://github.com/osm-fr/osmose-backend/blob/dev/osmosis/ImportDatabase_Prepare.sql#L14C7-L14C18) also calls ST_IsClosed, so it may be that this function is the only issue)

Famlam: I was about to file the following issue report at openstreetmap/osmosis:

> With the parameter `enableKeepPartialLinestring` enabled, the partial linestring of an incomplete way is kept.
> Calling the PostGIS function `ST_IsClosed` on this object however yields unexpected results:
>
> Based on the known ID of the nodes of a way, the ST_IsClosed function should be able to determine whether the first node equals the last node. In reality, ST_IsClosed will return False for all incomplete closed ways. This occurs both in case the first (=last) node is or is not present in the dataset containing the incomplete way.
>
> This may or may not be related to ST_NPoints from PostGIS returning the number of actually present (not incomplete) nodes, while the way object still contains the references to all nodes. So if the way in reality has 21 nodes (of which node 1 == node 21), but only 19 of them are actually present, it may be that the id of node 1 is compared to the id of node 19 instead of node 21. This is an unverified hypothesis.

But I realized that if I'm right*, fixing this (by removing non-existent nodes from the nodes list of the way) would be problematic for us as it would no longer allow to find incomplete ways (like we need to do for the multipolygon implementation). Hence I'm hesitating whether I should really request a change of the osmosis code by filing an issue.

'* the implementation of ST_NPoints and ST_IsClosed both use the internal variable `npoints` of the way, but I don't know where it grabs the (memory pointer of the) id of the node for comparing the first & last node that is used here: https://github.com/postgis/postgis/blob/3bbaa720c547c96a4fb2f8e4b08f6f18eb00a9cd/liblwgeom/ptarray.c#L723C47-L723C64


p.s.: offtopic, but due to this I realized that we can probably find incomplete ways by simply comparing ST_NPoints with array_length(ways.nodes,1)

frodrigo: > It may also be that it simply doesn't close the linestring if the first=last node is missing (zeeland case). 

That is the issue I think. As we know on Osmosis first_id=last_id, the partial linestring should be closed.

> offtopic, but due to this I realized that we can probably find incomplete ways by simply comparing ST_NPoints with array_length(ways.nodes,1)

No. As superposed nodes result in only one point in geometry.

Famlam: Correct, this is the geometry we get for the Zeeland case (way 1059958114). Here the end node *is outside* the extract.
```
LINESTRING(3.60106 51.643859500000005,3.6032414 51.6446471,3.6053391 51.645518300000006,3.6073448 51.646469700000004,3.6092506 51.6474976,3.611049 51.648597800000005,3.6127328 51.6497661,3.6142955 51.6509979,3.6157309 51.6522882,3.6170333 51.653632,3.6181975 51.655024100000006,3.619219 51.6564588,3.6200937 51.6579306,3.6208181 51.6594337,3.6213894 51.660962000000005,3.6218053 51.6625097,3.6220641 51.6640705,3.6221648 51.665638400000006,3.6221071 51.667207100000006)
```
The linestring is not closed here: first node isn't equal to the last node.

---

For the cases where the first=last node *is inside* the extract (example: https://www.openstreetmap.org/way/1118585665), the issue appears to be different. The linestring it generates is:
```
LINESTRING(-63.8188941 2.1040961,-63.6424262 2.1034099,-63.8188941 2.1040961)
```
ST_IsClosed does work on those, but obviously the linestring is invalid as it overlaps itself.

Famlam: Filed https://github.com/openstreetmap/osmosis/issues/133

Now the question is, should we:
1. Temporary use this patch as it is to bypass the issue, or
2. Switch to `ST_IsClosed` to fix the other issues and wait for the Osmosis fix to resolve the original issue?

frodrigo: I prefer fixing Osmosis. The fix should be simple and avoid doing more works.

Famlam: Ok, I'll change the PR to use ST_IsClosed then to solve i.e. the Venezuelan example (and the examples listed in the initial post, with the rollercoaster and the invalid ways) and leave the remaining issues to wait for Osmosis

frodrigo: I make an Osmosis PR https://github.com/openstreetmap/osmosis/pull/134

@Famlam if you wan to play locally with it I uploaded this osmosis version here https://cijoint.org/r/G8xz6FMa#l1RH21pm9hh+EWD2Et1ggRQ88Q7TTPS1ACiyzI+p8nI=

Famlam: Thanks! I think I'll have time to 'play' with it on Wednesday and see what still needs to be fixed on Osmoses side.
(P.s. not sure if intentional, but shouldn't it be `linePoints.size() > 1` instead of `> 0`? Otherwise you're generating a 1-duplicated-node-linestring. I'm not sure and don't want to create confusion in that PR, hence mentioning it here)

frodrigo: > (P.s. not sure if intentional, but shouldn't it be linePoints.size() > 1 instead of > 0? Otherwise you're generating a 1-duplicated-node-linestring when there's only one node. I'm not sure and don't want to create confusion in that PR, hence mentioning it here)

You are right. It is better.

Famlam: > @Famlam if you wan to play locally with it I uploaded this osmosis version here https://cijoint.org/r/G8xz6FMa#l1RH21pm9hh+EWD2Et1ggRQ88Q7TTPS1ACiyzI+p8nI=

I assumed I had to place this file in this folder: \osmosis\osmosis-0.48.3-34-gb5383475-SNAPSHOT\bin, the only place with a file called 'osmosis' but that doesn't seem to work (also, this file is almose 20MB, while the original file was a few kb). Any other place where I should put it? Below the output of osm110
```
  File "/./osmose_run.py", line 404, in run
    return execc(conf, logger, analysers, options, osmosis_manager)
  File "/./osmose_run.py", line 191, in execc
    osmosis_manager.init_database(conf, options)
  File "/modules/OsmOsisManager.py", line 203, in init_database
    bg_proc.append((self.logger.execute_err(cmd, background=parallel), "osmosis"))
  File "/modules/OsmoseLog.py", line 67, in execute_err
    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1823, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: '/osmosis/osmosis-0.48.3-34-gb5383475-SNAPSHOT/bin/osmosis'
```


---

However, it turns out that what I thought was one issue, were actually many issues:
- The rollercoaster one -> unrelated, now in #1960 
- The partially downloaded ways with the first=last node outside the extract -> pending osmosis approving the PR
- The partially downloaded ways with the first=last node *inside* the extract -> still to be fixed in Osmose

Hence, I've changed the PR to check against invalid ways and exclude those. This:
1. excludes truly invalid ways. This is a good change as the current warning is confusing, while there is another analyser that gives a much more useful warning
2. excludes ways where only 2 nodes are *inside* the extract (and more nodes outside the extract). Hence, the linestring created with these nodes is A -> B -> A, i.e. self-overlapping and thus invalid. Here no warning should be thrown as this is due to Osmose/Osmosis; the map data itself is valid.

frodrigo: > I assumed I had to place this file in this folder: \osmosis\osmosis-0.48.3-34-gb5383475-SNAPSHOT\bin, the only place with a file called 'osmosis' but that doesn't seem to work (also, this file is almose 20MB, while the original file was a few kb). Any other place where I should put it?

You have to uncompres it in place of `osmosis-0.48.3-34-gb5383475-SNAPSHOT`

Famlam: Hmm, hadn't thought of it being a compressed file in a compressed file :)

Works for netherlands_zeeland; testing it against others too now. I assume you don't want me to include it in this PR yet?
(The PR as it is now already fixes some issues and will be needed anyway, so IMO it can be merged while we wait for Osmosis to merge it)

frodrigo: If it is ok for you. We need to make a PR on osmose to change osmosis, event before the osmosis PR was merged.

Famlam: I tested it on 13 countries. So far the results look good: removing a lot of false positives. Note this was with the version which did add a second node for a 1-node-in-the-extract-way.

> We need to make a PR on osmose to change osmosis, event before the osmosis PR was merged.

Would you be able to do this? I can't build the package myself (I think) and since the change with >0 nodes -> >1 nodes was made after you build the package, it would need to be rebuild.

frodrigo: I made the Osmose Osmosis PR #1963

Famlam: Thanks! I'm testing that one now. Is there anything that I have to do here or can this one be merged? This one should be merged before the other PR (otherwise it'll give errors on every closed way with 2 nodes in the extract).

PR #1955, opened 2023-07-12
Rewrite analyse merge for public services in France

nlehuby:
The source dataset changed format and structure in July last year.

The classes 2, 3 and 4 seems unused for this item, so I've used them for some new public service types.


frodrigo:
If it is a replacement, please keep the same class id.

The converter is also written in python https://github.com/nlehuby/sepuqu/blob/main/transform_dataset.py

It could be better to integrate it directly to Osmose, couldn't ?

nlehuby:
Yes, of course, but it depends on the possibilities of supervision provided by Osmose: how often would the script be run, and how could failures be monitored?
I wrote the previous script here, and have no idea when it last ran, nor have I been alerted to the fact that the format has changed for almost a year already.
The dedicated script seems easier to monitor to my opinion.

nlehuby:
it looks like the Osmose analysis logs, but I was talking about the script next to it, the one that downloads the data and preprocesses it into a more readable format for Osmose, this one https://github.com/osm-fr/osmose-backend/blob/master/merge_data/service_public_FR.py that I deleted here to replace it with my external repo where the runs can be more easily monitored. Is it also handled by Osmose buildbot ?

frodrigo:
> the one that downloads the data and preprocesses it into a more readable format for Osmose, this one https://github.com/osm-fr/osmose-backend/blob/master/merge_data/service_public_FR.py that I deleted here to replace it with my external repo where the runs can be more easily monitored. Is it also handled by Osmose buildbot ?

No. It only a manual run.

> but I was talking about the script next to it,

The right way to do it now, it is using a custom Source eg. https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_merge_fuel_IT.py#L92

nlehuby:
ok, i'll try to do a custom source instead.

analysers/analyser_merge_public_services_FR.py
```diff
@@ -154,3 +160,197 @@ def __init__(self, config, logger = None):
...
+
+class Public_Services_Source(Source):
+    def __init__(self, source):
+        self.source = source
+
+    def __getattr__(self, name):
+        return getattr(self.source, name)
+
+    def open(self):
+        return open(downloader.update_cache('join://' + self.source.fileUrl, 15, fetch=self.fetch))
```
frodrigo, line 172:
```suggestion
        return open(downloader.update_cache('parse://' + self.source.fileUrl, 15, fetch=self.fetch))
```

analysers/analyser_merge_public_services_FR.py
```diff
@@ -0,0 +1,356 @@
...
+import tarfile
+
+class _Generic_Analyser_Merge_Public_Services_FR_(Analyser_Merge_Point):
+    def __init__(self, config, logger, osmose_class, public_service_category, osm_select_tags, osm_default_tags):
+        Analyser_Merge_Point.__init__(self, config, logger)
+        self.def_class_missing_official(item = 8110, id = osmose_class, level = 3, tags = ['merge', 'fix:imagery', 'fix:survey', 'fix:picture'],
+            title = T_('Public service not integrated'))
+
+        self.init(
+            u"https://www.data.gouv.fr/fr/datasets/service-public-fr-annuaire-de-l-administration-base-de-donnees-locales/",
```
frodrigo, line 39:
```suggestion
            "https://www.data.gouv.fr/fr/datasets/service-public-fr-annuaire-de-l-administration-base-de-donnees-locales/",
```

frodrigo:
Thank you

PR #1957, opened 2023-07-13
Add warning for repeated conditions & avoid duplicate reports

Famlam:
This adds a warning when a condition is repeated, overruling (in case of `;`-separated conditions) or at least making redundant (in case of ` AND `-separated conditions) the preceding condition.

In addition it should fix getting a duplicate marker for such cases, in case there's a warning for the condition. (Using some parts of todays earlier commit by Frodrigo)

Also whitelists `*:conditional` from TagFix_DuplicateValue as the new check is more thorough. (TagFix_DuplicateValue wouldn't find it if the values differed, but the conditions were the same)

### DIFF
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -42,6 +42,8 @@ class ConditionalRestrictions(Plugin):
     # Workaround https://bugs.kde.org/show_bug.cgi?id=452236
     self.kOpeningHours452236 = re.compile(r'(20\d\d [A-Z][a-z][a-z] \d\d?\s?-\s?)20\d\d ([A-Z][a-z][a-z] \d\d?)')
 
+    self.getDuplicatesFromList = lambda lst: [x for x in set(lst) if lst.count(x) > 1]
+
     OHplugin = TagFix_Opening_Hours(None)
     self.sanitize_openinghours = OHplugin.sanitize_openinghours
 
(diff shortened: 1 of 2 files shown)


plugins/ConditionalRestrictions.py
```diff
@@ -42,6 +42,8 @@ def init(self, logger):
     # Workaround https://bugs.kde.org/show_bug.cgi?id=452236
     self.kOpeningHours452236 = re.compile(r'(20\d\d [A-Z][a-z][a-z] \d\d?\s?-\s?)20\d\d ([A-Z][a-z][a-z] \d\d?)')
 
+    self.getDuplicatesFromList = lambda list: set([x for x in list if list.count(x) > 1])
```
frodrigo, line 45: Better make the set before. Something like
```suggestion
    self.getDuplicatesFromList = lambda list: [x for x in set(list) if list.count(x) > 1]
```

frodrigo:
At first, the idea was to avoid duplicate on thus
https://www.openstreetmap.org/way/24096310
https://www.openstreetmap.org/way/24100844
https://www.openstreetmap.org/way/24108427
https://www.openstreetmap.org/way/24106677
https://www.openstreetmap.org/way/929053287

Famlam:
> At first, the idea was to avoid duplicate on thus

Ah, I see, I'm missing the parameter `condition` in the subclass of the errors on lines 197, 199, 205. Will update tonight/tomorrow.

plugins/ConditionalRestrictions.py
Famlam: Done; also renamed list to lst to not unnecessarily overwrite any python function

Famlam:
Implemented! I now also understand why you wanted to get the indent of the AND-splitted list down: no need for the same warning (even if the subclass differs) for the same textual warning

frodrigo:
Thank you for dig into this.

Issue #1896, opened 2023-06-10
error: NULL location provided in osmosis_relation_public_transport for several norway*


Marc-marc-marc:
https://buildbot.osmose.openstreetmap.fr/#/builders/639/builds/1002/steps/3/logs/stdio
2023-06-10 06:48:33 norway_aust_agder : osmosis_relation_public_transport
2023-06-10 06:48:33   osmosis resume post scripts
2023-06-10 06:48:35   run osmosis base analyser Analyser_Osmosis_Relation_Public_Transport
2023-06-10 06:48:35   requires table highways
2023-06-10 06:48:41   analyser_osmosis_relation_public_transport.py:616 sql
2023-06-10 06:48:41   analyser_osmosis_relation_public_transport.py:617 sql
2023-06-10 06:48:41   analyser_osmosis_relation_public_transport.py:618 xml generation
2023-06-10 06:48:41   analyser_osmosis_relation_public_transport.py:619 sql
2023-06-10 06:48:42   analyser_osmosis_relation_public_transport.py:620 sql
2023-06-10 06:48:42   analyser_osmosis_relation_public_transport.py:621 sql
2023-06-10 06:48:42   analyser_osmosis_relation_public_transport.py:622 xml generation
2023-06-10 06:48:42   analyser_osmosis_relation_public_transport.py:623 xml generation
2023-06-10 06:48:42   analyser_osmosis_relation_public_transport.py:624 xml generation
2023-06-10 06:48:42   analyser_osmosis_relation_public_transport.py:625 xml generation
2023-06-10 06:48:42   error: NULL location provided

frodrigo:
Not a real error. The analyse emit an issue without location. Is logged,
but not blocking error.

Marc-marc-marc:
Could the log level be changed to warning in this case?
This complicates targeting real errors if non-errors have the error level log

frodrigo:
We log it initially at error to improve the analysers. But now we can
switch it warning.

Famlam:
https://github.com/osm-fr/osmose-backend/blob/3559f8519cad84d520b6d7399f0bf7de5b2fb9fa/analysers/Analyser_Osmosis.py#L637

Famlam:
Hmm, it's a bit odd...
Running it on netherlands_gelderland gives (among others) the following output of sql50:

| variable | output |
| --- | --- |
|`parent.id` | [13990411](https://www.openstreetmap.org/relation/13990411) |
[...]

PR #1961, opened 2023-07-19
warn on NULL location instead of error

Famlam:
See #1896 (the first 4 messages)

Change err to warn


frodrigo:
Ok. Thank you.

cc @Marc-marc-marc

PR #1966, opened 2023-07-22
Make subclass unique

### LINKED ISSUE #1962: Highway_Parking_Lane, duplicate issue item=31625 class=3162
Way https://www.openstreetmap.org/way/173095901 raise duplicate issue item=31625 class=3162

With
- parking:both 	lane
- parking:both:orientation 	parallel
- parking:condition:right:maxstay 	3
- parking:left 	no
- parking:left:reason 	narrow
- parking:right 	no
- parking:right:reason 	narrow

Famlam:
Fix #1962 

In case `parking:both:*=* + parking:right=no + parking:left=no` are all present simultaneously, the subclass was not unique. (Idem if the latter two tags are `=separate`)


frodrigo:
Ok. Thank you.

PR #1973, opened 2023-07-26
Multilingual ast / es order

### LINKED ISSUE #1972: Multilingual ast / es order
I'm noticing an issue [here](https://osmose.openstreetmap.fr/es/issue/a95bfd0e-a463-3db4-7c20-f1028d442fc1) that should't. The element has these tags:

```
name = Xichón / Gillón
name:ast = Xichón
name:es = Gillón 
```

I presume the issue comes up due to the order. Osmose expects `name = Gillón / Xichón`, but that's wrong. The official name is in Asturian, then Spanish. But the solution is not about reversing the order, since another places are named firstly in Spanish, then Asturian.

Either you duplicate the following condition to deal ["es", "ast"] and ["ast", "es"], or you can check just the presence of the tags, no matter the order.

https://github.com/frodrigo/osmose-backend/blob/16582a0245fb8a874044c55dcac47aab831515c4/plugins/Name_Multilingual.py#L72

Same case for Navarre, basque / spanish, or spanish / basque, both are right. Perhaps the same for Basque country

_Originally posted by @Crashillo in https://github.com/osm-fr/osmose-backend/issues/1285#issuecomment-1652026460_

Crashillo:
Fixes #1972

### DIFF
--- a/plugins/Name_Multilingual.py
+++ b/plugins/Name_Multilingual.py
@@ -69,6 +69,7 @@ class Name_Multilingual(Plugin):
                     {"name": tags["name:"+lang[0]].strip()},
                     {"name": tags["name:"+lang[1]].strip()},
                     {"name": tags["name:"+lang[0]].strip() + separator + tags["name:"+lang[1].strip()]},
+                    {"name": tags["name:"+lang[1]].strip() + separator + tags["name:"+lang[0].strip()]},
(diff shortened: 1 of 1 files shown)


frodrigo:
Look good, thank you.
Merged but not deployed yet.

PR #1974, opened 2023-07-29
Remove false positives for incorrect tag check

pixunil:
I found three documented tags that I believe are false positives as they are documented in the wiki. Especially `bicycle=optional_sidepath` and `crossing=informal` have a rather large occurrence.

### DIFF
--- a/plugins/TagFix_Access.py
+++ b/plugins/TagFix_Access.py
@@ -38,9 +38,9 @@ class TagFix_Access(Plugin):
                        "hazmat:water", "hgv", "horse", "hov", "inline_skates", "light_rail", "minibus", "mofa", "moped", "motor_vehicle", "motorboat", "motorcar", "motorcycle", "motorhome", "psv",
                        "share_taxi", "ship", "ski:nordic", "ski", "small_electric_vehicle", "snowmobile", "speed_pedelec", "subway", "swimming", "tank", "taxi", "tourist_bus", "trailer", "train", "tram", "vehicle"]
     self.accessValuesGeneral = ["yes", "no", "private", "permissive", "permit", "destination", "delivery", "customers", "designated", "use_sidepath", "agricultural", "forestry", "discouraged"]
-    self.accessValuesSpecial = {"bicycle": ["dismount"],
+    self.accessValuesSpecial = {"bicycle": ["optional_sidepath", "dismount"],
                                 "canoe": ["put_in", "portage"],
-                                "dog": ["leashed", "unleashed"],
+                                "dog": ["leashed", "unleashed", "outside"],
                                 "horse": ["dismount"],
                                 "mofa": ["dismount"],
                                 "moped": ["dismount"],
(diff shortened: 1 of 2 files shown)


Famlam:
`bicycle=optional_sidepath` is controversial according to the wiki of key:bicycle. Not sure what it tries to imply either, if you're allowed to be there with equal priority as on the optional cycleway, it's simply a duplicate of "bicycle=yes".

`crossing=informal` is another Streetcomplete tag that causes a lot of bad tagging in my area since some streetcomplete users don't understand the difference between a purpose crossing and an informal one because their quest seems to imply that any crossing without a lowered curb or lines or so is an informal one :/. I usually have to revert them. So personally I like keeping the warning, makes it easy to spot the bad tagging :)

`dog=outside` seems to be used only 98 times worldwide, but makes sense to me.

Just my opinion :)

pixunil:
Thanks for your feedback!
Regarding `bicycle=optional_sidepath`, I only found on [`bicycle`](https://wiki.openstreetmap.org/wiki/Key:bicycle) this sentence:
> note that standard tagging for that info is [cycleway](https://wiki.openstreetmap.org/wiki/Key:cycleway)=[separate](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dseparate) and [cycleway:left](https://wiki.openstreetmap.org/wiki/Key:cycleway:left)=[separate](https://wiki.openstreetmap.org/wiki/Tag:cycleway:left%3Dseparate) / [cycleway:right](https://wiki.openstreetmap.org/wiki/Key:cycleway:right)=[separate](https://wiki.openstreetmap.org/wiki/Tag:cycleway:right%3Dseparate) / [cycleway:both](https://wiki.openstreetmap.org/wiki/Key:cycleway:both)=[separate](https://wiki.openstreetmap.org/wiki/Tag:cycleway:both%3Dseparate) variants

Is this the controversion you are referring to? I would argue that `optional_sidepath` matches `use_sidepath` to refer to the sidepath while keeping the access to `yes`. However, it might be not applicable to all countries.

Regarding `crossing=informal`, I share your criticism of the wording in the quest. If in the current situation it is better for Osm QA to manually resolve real informal crossings as the tagging is so error-prone, it would be fine to me to keep them flagged.

Famlam:
Yes, that's what I'm referring to.
All other bicycle values give access values (use_sidepath for instance in NL means "not allowed, except when you're too wide", no means no, designated means it's specially for you, etc).
This one just says "yes, with another possibility nearby". Which is just "yes", because the optional cycleway doesn't affect the access of the nearby way.

For `crossing=informal` I guess I (or you?) will have to write a separate SQL query to catch the many rotten apples. But I guess that doesn't have to stop this PR until then.. Just not so happy to lose my "how to find them" method I guess 😅

Famlam:
Filed an issue to add an analyzer for bad informal crossings later on :)

frodrigo:
So, @Famlam your are in favour of only keep `dog=outside` ?

Famlam:
I think it can be merged as it is, or without optional_sidepath.


I don't like the tag crossing=informal due to its common misuse, but the tag itself is valid, so we need to get a more appropriate check to find the bad ones. (For which I filed a corresponding issue). So this can be merged.


Optional_sidepath so far mostly seems to be a German tag, and with the necessary controversy on the wiki. So I personally would prefer to keep it out until resolved, but on the other hand, no hard feelings if we stop warning about it. (I'll bring it up in the Dutch community when my schedule allows)

Up to you :)

frodrigo:
Thank you. Merged. Deploy in progress.

PR #1977, opened 2023-08-01
Update TagFix_Maxspeed.py

pl71:
Max speed on Bulgarian motorways is 140 km/h
There are proposals in law for lowering it to 130 km/h, but they are still not in force.

https://www.mfa.bg/upload/649/09-ZDvP-bg-en.pdf
Art. 21. (1) (page 8)


frodrigo:
Thank you. also to be reported here (same data) : https://github.com/Project-OSRM/osrm-backend/blob/master/profiles/car.lua#L270

pl71:
@frodrigo Done.

frodrigo:
Merged, but not deployed, yet.

Issue #1844, opened 2023-05-03
Wrong syntax for start_date value on power plant


Stefal:
Item : 8270
class: 1
link example: https://osmose.openstreetmap.fr/fr/issue/6921eeb7-e2f8-1251-87db-3ffc5ae8da02
![image](https://user-images.githubusercontent.com/8255984/235836888-d8931f49-beb7-46d9-8115-5ad9c2a7df69.png)

The start_date tag should contains value in [ISO8601](https://wiki.openstreetmap.org/wiki/Key:start_date) YYYY-MM-DD format, instead of DD/MM/YYYY

kshitiz305:
Hi I wish to contribute to it may I know which file needs to be changed for this?

frodrigo:
The link to the source code is in the osmose issue detail.

https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_merge_power_plant_FR.py#L33

kshitiz305:
May I know which is the file through which this script can be run as

frodrigo:
https://github.com/osm-fr/osmose-backend/blob/dev/doc/0-Index.md#execution

And other docs : https://github.com/osm-fr/osmose-backend/tree/dev/doc

Prefer using Docker to seutp, also explain how to run : 
https://github.com/osm-fr/osmose-backend/tree/dev/docker

kshitiz305:
But still Analyser_Merge_Power_Plant_FR Class is not getting executed anywhere?

frodrigo:
> But still Analyser_Merge_Power_Plant_FR Class is not getting executed anywhere not it is getting used anywhere?

It is not referenced. It is in the config file https://github.com/osm-fr/osmose-backend/blob/dev/osmose_config.py#L594

But can be tested manually using `osmose_run.py` option.
```
./osmose_run.py --no-clean --country=france_aquitaine_landes --analyser=merge_power_plant_FR
```

kshitiz305:
`/data/work/{user}/extracts/france_aquitaine_landes.osm.pbf`   the following file is not found please let me know how to download it or where to fetch it from.

kshitiz305:
additionally `psycopg2.errors.InvalidSchemaName: schema "france_aquitaine_landes" does not exist` is additional error

frodrigo:
Was only an example. Please reread the docs.

PR #1980, opened 2023-08-06
Fix date format in analyser_merge_power_plant_FR #1844

frodrigo:


Famlam:
If the dates in the source are now in `/`-separated format, are the "-01-01" and "-12-31" checks still needed then?

frodrigo:
> If the dates in the source are now in `/`-separated format, are the "-01-01" and "-12-31" checks still needed then?

Right. Need also format update. Thank you.

Famlam:
Not startswith and the other way around? Since the format needs to be inverted after the split?
(If these data entries still exist)

frodrigo:
> Not startswith and the other way around? Since the format needs to be inverted after the split? (If these data entries still exist)

Still right ! Sorry for too quick fixing.

Issue #1493, opened 2022-07-10
Amenities ignored by analysis north of France


ThmVlm:
Some amenities located in the very north of France (above lat 50.82) are not spotted by the analysis whereas they are well located in osm : 
- schools : http://osmose.openstreetmap.fr/fr/map/#zoom=10&lat=50.8514&lon=2.477&item=8030&level=1%2C2%2C3
- stations : http://osmose.openstreetmap.fr/fr/map/#zoom=10&lat=50.8514&lon=2.477&item=8050&level=1%2C2%2C3
- police : http://osmose.openstreetmap.fr/fr/map/#zoom=10&lat=50.8514&lon=2.477&item=8190&level=1%2C2%2C3
- gas stations : http://osmose.openstreetmap.fr/fr/map/#zoom=10&lat=50.8679&lon=2.419&item=8200&level=1%2C2%2C3

ThmVlm

frodrigo:
Please, can you explain what is wrong for you, and what are you expected ?

ThmVlm:
All this amenities in the very north of France are listed in osmose as missing, whereas they all already are in osm. 
It seems that the analysis ignores the amenities above latitude 50.82

frodrigo:
Ok. I see your point.

It looks to be the case for various and unrelated data source.

But it is not the case for all data sources.

frodrigo:
I do not see the same issue on local run.

@jocelynj do you think we can try to delete the exact from the backend ?
- france_nord_pas_de_calais_nord
- france_nord_pas_de_calais_pas_de_calais

frodrigo:
Problem is linked to special backend `france_local_db`.

frodrigo:
Fixed.

PR #1982, opened 2023-08-08
Compute bbox on global projection to avoid miscomputation #1493

frodrigo:


flacombe:
Sounds like a good solution.
What remains to be done on testing about that?

Can I help?

frodrigo:
Not remember the details, now. But the implementation is wrong.

PR #1983, opened 2023-08-08
Add analyser that verifies suspicious angles in railway

Tristramg:
This analyser is quite similar to `Analyser_Osmosis_Way_Approximate` however it works a bit differently as it verifies the angle consecutive nodes of a way. It can make sens on roads, but very rarely on trains.

A few things I’m still not sure:
- I’m still confused about the class and item I should use
- Translations: I kept pretty much the same wording as the Way Approximate analyser as the analyses still make sens, but I didn’t how I could generate the right entries for the `fr.po`

### DIFF
--- a/analysers/analyser_osmosis_way_angle.py
+++ b/analysers/analyser_osmosis_way_angle.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Tristram Gräbener 2023                                     ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+# This analyser is similar to Analyser_Osmosis_Way_Approximate
+# However, this one measures angles and not radiuses
+# It most likely indicates a node that was accidentaly dragged
+# While Analyser_Osmosis_Way_Approximate indicates that the way was drawn
+# with too few nodes
+
+from modules.OsmoseTranslation import T_
+from .Analyser_Osmosis import Analyser_Osmosis
+
+sql10 = """
+WITH
+    points AS (
+        SELECT
+            id,
+            tags->'{tag}' AS type,
+            (ST_DumpPoints(linestring)).geom AS geom4326,
+            (ST_DumpPoints(ST_Transform(linestring, {srid}))).geom AS geom,
+            (ST_DumpPoints(ST_Transform(linestring, {srid}))).path[1] AS index
+        FROM
+            {table_prefix}ways as ways
+        WHERE
+            tags != ''::hstore AND
+            tags?'{tag}' AND tags->'{tag}' IN ('{tag_values}') AND
+            ST_NPoints(linestring) >= 3
+    ),
+    angles AS (
+        SELECT
+            id, index, type,
+            lead(geom4326) OVER (PARTITION BY id ORDER BY index) AS geom4326,
+            lead(geom) OVER (PARTITION BY id ORDER BY index) AS geom,
+            degrees(
+                ST_Angle(geom,
+                     lead(geom, 1) OVER (PARTITION BY id ORDER BY index),
+                     lead(geom, 2) OVER (PARTITION BY id ORDER BY index)
+                )
+            ) angle,
+            ST_Distance(
+                geom,
+                lead(geom, 1) OVER (PARTITION BY id ORDER BY index)
+            ) AS length_1,
+            ST_Distance(
+                lead(geom, 1) OVER (PARTITION BY id ORDER BY index),
+                lead(geom, 2) OVER (PARTITION BY id ORDER BY index)
+            ) AS length_2
+        FROM
+            points
+    )
+
+    SELECT
+        id,
+        ST_AsText(geom4326),
+        angle,
+        type,
+        index
+    FROM
+        angles
+    WHERE
+        (angle < 150 OR angle > 210)
+        AND length_1 < 100
+        AND length_2 < 100
+"""
+
+class Analyser_Osmosis_Way_Suspicious_Angle(Analyser_Osmosis):
+
+    def __init__(self, config, logger = None):
+        Analyser_Osmosis.__init__(self, config, logger)
+
+        self.classs_change[11] = self.def_class(item = 1190, level = 2, tags = ['geom', 'railway', 'fix:imagery'],
+            title = T_('Suspicious angle in way'),
+            detail = T_(
+'''Sharp angles on a railway are suspicious. Maybe a node was accidentally dragged?'''),
+            fix = T_(
+'''After checking orthophotos, add nodes or move existing nodes.'''),
+            trap = T_(
+''' On service ways, train stations and train workshops this could be a false positive.'''),
+            example = T_(
+'''![](https://wiki.openstreetmap.org/w/images/9/9d/Osmose-eg-error-1190.png)
+
+`railway=rail` crudely drawn.'''))
+
+        self.tags = ("rail", "light_rail")
+        self.callback10 = lambda res: {"class":11, "subclass":res[4], "data":[self.way_full, self.positionAsText], "text": T_("railway={0} with suspicious angle {1}°", res[3], int(res[2]))}
+
+    def analyser_osmosis_full(self):
+        self.run(sql10.format(table_prefix="", tag="railway", tag_values="', '".join(self.tags), srid=self.config.options.get("proj")), self.callback10)
+
+    def analyser_osmosis_diff(self):
+        self.run(sql10.format(table_prefix="touched_", tag="railway", tag_values="', '".join(self.tags), srid=self.config.options.get("proj")), self.callback10)
--- a/analysers/analyser_osmosis_way_angle.py
+++ b/analysers/analyser_osmosis_way_angle.py
@@ -49,7 +49,6 @@ WITH
         SELECT
             id, index, type,
             lead(geom4326) OVER (PARTITION BY id ORDER BY index) AS geom4326,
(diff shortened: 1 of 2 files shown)


Famlam:
Just for my understanding, how do these results differ with the results from the existing analyser, e.g. http://osmose.openstreetmap.fr/nl/issues/open?source=33602&item=1190&class=10 ? (Except for that it includes light rail)?

Because if there's a lot of duplicate reports, with basically the same message, it may be relevant to UNION the results? Or exclude those ways which have been detected already? (Just my opinion)

Don't worry about the translations, that's synchronized :)

P.s. sharp angles are wrong in any rail, not just rail and light_rail, right? Or does this give too much output?

Famlam:
> From my understating from the other analyser, it tries to compute radiuses instead of angles. Both are complementary and won’t detect the same things. (But my trigonometry is bad, maybe I missed an important point).

I can run a comparison to see how much overlap there is on some countries, but won't be able to do so until mid/end next week. I suspect that the bad radiuses will commonly be correlated with bad angles, but not sure.

> Other (significant) railway are narrow_gauge, subway and tram. Because they are used with smaller vehicles, sharper angles are more common. Maybe an approach would be to have different angles depending on the kind of rail, but that would require some trial and error to find good values.

Yes, and if the angle becomes a variable, it may be implemented for i.e. (car) highways too. A 170° turn in one node sounds unlikely to be precise. But this would be a next PR I think

frodrigo:
`Analyser_Osmosis_Way_Approximate` look for circle radius. While angle depend on map data scale, radius not. With a low resolution of highway or railway, a shape angle could be see but at hight resolution it will be not so sharp for the same radius. It is all about level of detail in OSM data.

Maybe computing direct angle make sens with small way segment only, so we sure we are at hight resolution?

Tristramg:
Indeed, the distance make a difference. What we were trying to catch are
more likely to be mapping errors, like a node dragged by accident.
Maybe restricting this analyser when both segments are less than 100m might
make sense? What do you think?

If yes, the projection passed as parameters computes distances as meters,
am I right?

frodrigo:
> Indeed, the distance make a difference. What we were trying to catch are
more likely to be mapping errors, like a node dragged by accident.
Maybe restricting this analyser when both segments are less than 100m might
make sense? What do you think?

I think both are interesting

> If yes, the projection passed as parameters computes distances as meters,
am I right?

yes

Tristramg:
I added a length limit, a comment to explain the difference with the existing analyser and changed a bit the explanation strings.

An example could be https://www.openstreetmap.org/node/4983237612#map=21/48.21104/12.58973&layers=T
image

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,105 @@
...
+            id, index, type,
+            lead(geom4326) OVER (PARTITION BY id ORDER BY index) AS geom4326,
+            lead(geom) OVER (PARTITION BY id ORDER BY index) AS geom,
+            degrees(
+                ST_Angle(geom,
+                     lead(geom, 1) OVER (PARTITION BY id ORDER BY index),
+                     lead(geom, 2) OVER (PARTITION BY id ORDER BY index)
+                )
+            ) angle,
+            ST_Length(geom, lead(geom, 1) OVER (PARTITION BY id ORDER BY index)) AS length_1,
```
Famlam, line 59: 2023-08-16 10:31:29     ST_Length(geom, lead(geom, 1) OVER (PARTITION BY...
2023-08-16 10:31:29     ^
2023-08-16 10:31:29     HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

Tristramg: Oh sorry :facepalm: 
I pushed the fixed version

Famlam:
Comparing the results of the analysers I guess there's no need to filter to prevent duplicates: most results are unique.

| country | # way approx | (of which) # railway approx | # suspicious angle | # of duplicate reports |
| --- | --- | --- | --- | --- |
| venezuela | 11796 | 1 | 0 | 0 |
| usa_iowa | 134 | 3 | 13 | 0 |
| slovenia | 22 | 0 | 9 | 0 |
| japan_chubu | 127 | 0 | 27 | 0 |
| kazakhstan | 19222 (2727 ways) | 689 | 164 (158 ways) | 2 ways |
| china_xinjiang | 15248 | 246 | 19 | 0 |

Results attached: [results.zip](https://github.com/osm-fr/osmose-backend/files/12360905/angle.zip)

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,111 @@
...
+'''After checking orthophotos, add nodes or move existing nodes.'''),
+            trap = T_(
+'''On service ways, train stations, train workshops that may be either a
+false positive'''),
+            example = T_(
+'''![](https://wiki.openstreetmap.org/w/images/9/9d/Osmose-eg-error-1190.png)
+
+`railway=rail` crudely drawn.'''))
+
+        self.tags = ("rail", "lightrail")
```
Famlam, line 104: https://wiki.openstreetmap.org/wiki/Tag:railway%3Dlight_rail
light_rail with an underscore

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,111 @@
...
+        self.callback10 = lambda res: {"class":10, "subclass":res[2], "data":[self.way_full, self.positionAsText], "text": T_("railway with suspicious angle {0}°", res[1])}
```
Famlam, line 105: Analyser way_approximate uses the same class and item id. To allow users to filter, I would use class `11` here (and thus also in line 90)

---

Additionally, I think the angle is in `res[2]`?
The current message reads: `railway with suspicious angle POINT(81.36814960000001 50.936915400000004)°` with `subclass="352.6731599505589"` (the angle).
(I think subclass must be an integer? Maybe this is where you wanted to use `index` (=`res[4]`)?

`res[3]` (the value of tag `railway=*`) is currently not used.

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,111 @@
...
+        Analyser_Osmosis.__init__(self, config, logger)
+
+        self.classs_change[10] = self.def_class(item = 1190, level = 2, tags = ['geom', 'railway', 'fix:imagery'],
+            title = T_('Suspicious angle in railway'),
+            detail = T_(
+'''Sharp angles on a railway are suspicious. Maybe a node was accidentaly dragged'''),
+            fix = T_(
+'''After checking orthophotos, add nodes or move existing nodes.'''),
+            trap = T_(
+'''On service ways, train stations, train workshops that may be either a
```
Famlam, line 97: This is also in the other analyser and should probably be fixed in both, but I think the grammar of this line could be improved. Suggestion:
`On service ways, train stations and train workshops this could be a false positive.`

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,111 @@
...
+
+class Analyser_Osmosis_Rail_Suspicious_Angle(Analyser_Osmosis):
+
+    def __init__(self, config, logger = None):
+        Analyser_Osmosis.__init__(self, config, logger)
+
+        self.classs_change[10] = self.def_class(item = 1190, level = 2, tags = ['geom', 'railway', 'fix:imagery'],
+            title = T_('Suspicious angle in railway'),
+            detail = T_(
+'''Sharp angles on a railway are suspicious. Maybe a node was accidentaly dragged'''),
```
Famlam, line 93: Typo: `accidentally`
Also, please end the line with a `.` or `?` ;)

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,111 @@
...
+        index
+    FROM
+        angles
+    WHERE
+        (angle < 150 OR angle > 210)
+        AND length_1 < 100
+        AND length_2 < 100
+"""
+
+class Analyser_Osmosis_Rail_Suspicious_Angle(Analyser_Osmosis):
```
Famlam, line 85: Just with an eye on the future: 
I guess this analyser (with flexible angle restrictions) could in the future be expanded to include car highways with sharp angles. Maybe better to name it more neutral with `way` instead of `railway`. (Also, I guess it's convenient to have the same file name as analyser class name)

analysers/analyser_osmosis_railway_angle.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
```
Famlam, line 1: General note: to enable the analyser for all countries, it must be added to this list too:
https://github.com/osm-fr/osmose-backend/blob/dev/osmose_config.py#L197

Famlam (changes requested):

analysers/analyser_osmosis_railway_angle.py
Tristramg: I changed it to way, but I not yet convinced that it would make sens for roads and might result in a lot of false positive (e.g. https://www.openstreetmap.org/way/689270247)

Tristramg:
Thank you for this precious review!

analysers/analyser_osmosis_railway_angle.py
Famlam: Yes, for roads (for cars) it should probably be limited to e.g. <30° angles or so, definitely less than 90 or even 150°.

Problem for the future, not for now 😉

Famlam (approved):
No further comments from me :)
p.s. ignore the "Famlam self-requested a review". Wrong button, I wanted to remove my earlier "requested changes" review ;)

analysers/analyser_osmosis_way_angle.py
```diff
@@ -0,0 +1,110 @@
...
+        WHERE
+            tags != ''::hstore AND
+            tags?'{tag}' AND tags->'{tag}' IN ('{tag_values}') AND
+            ST_NPoints(linestring) >= 3
+    ),
+    angles AS (
+        SELECT
+            id, index, type,
+            lead(geom4326) OVER (PARTITION BY id ORDER BY index) AS geom4326,
+            lead(geom) OVER (PARTITION BY id ORDER BY index) AS geom,
```
frodrigo, line 52: Could be removed.

```suggestion
            lead(geom) OVER (PARTITION BY id ORDER BY index) AS geom,
```

Tristramg: I’m don’t understand the suggested change, it looks the same. What could be removed?

Famlam: The fix suggestion is strange indeed, but I think @frodrigo means that `geom` isn't used, i.e. line 52 can be removed

Tristramg: Thank you! It’s removed

frodrigo:
Merged. Thank you all.

frodrigo:
The result is live
https://osmose.openstreetmap.fr/en/issues/open?item=1190&class=11

Tristramg:
Thanks both of you for your patience to get me to a working plugin!

PR #1989, opened 2023-08-18
Fix single-quoted strings

### LINKED ISSUE #1934: Parse '' as ' in `mapcss.tr`
In https://github.com/osm-fr/osmose-backend/issues/1827 we had the following conversation:

Famlam:
> `'` is a special character in JOSM. You can't use it for string quotations, and if you want to use it in a `tr(...)` string, you'll have to use `''` to get a single apostrophe (we don't parse this: we just output two aposthrophes). Or replace it by i.e. `.

Frodrigo:
> We can fix this, but probably not in the grammar.

Famlam:
> Yes, it's not an urgent issue though as '' looks almost the same as " and usually that works out well. Since it only affects `tr()`-translated strings, probably the fix should be in its definition in mapcss_lib.py? [...]
---

Although it wasn't blocking that specific issue, it just became a lot more urgent: recently JOSM has added the following rule in [numeric.mapcss](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/numeric.mapcss) 
`throwWarning: tr("unusual value of {0}: use '' for foot and \" for inches, no spaces", "{0.key}");`

In JOSM this translates to:
> unusual value of {0}: use ' for foot and " for inches, no spaces

In Osmose this is parsed as:
> unusual value of {0}: use '' for foot and " for inches, no spaces

So Osmose now seems to suggest to use a double apostrophe for foot units

Famlam:
In JOSM `'` is a special character, so to print an actual `'` in JOSM, you have to use `''`. Hence, the JOSM translations contain `''` where `'` is the desired output. In Osmose, the character is not special, so it was previously (wrongly) printed as a double `''`.

In order to not get strange translations, such as:
> unusual value of width: use '' for foot and " for inches, no spaces

instead of:

> unusual value of width: use ' for foot and " for inches, no spaces

, we have to convert `''` to `'`. 

Fixes #1934
Attached the output of usa_iowa before and after this change to see the effect: [tr.zip](https://github.com/osm-fr/osmose-backend/files/12383866/tr.zip)

p.s.: please note that this does not affect the strings in mapcss/item_map.py. Since they're just an internal mapping, they can be double-quoted; keeps the code much simpler.
p.p.s.: the `''` to `'` conversion is only needed for `tr(...)` strings. If `tr(...)` is omitted (meaning a string literal, without translations or placeholders) `''` is not converted to `'`

### DIFF
--- a/mapcss/mapcss_lib.py
+++ b/mapcss/mapcss_lib.py
@@ -445,7 +445,10 @@ def JOSM_search(string):
 #    translate from English to the current language (only for strings in the JOSM user interface) [since 6506]
 def tr(string, *args):
     if string is not None:
-        return T_(string, *args)
+        # Treat '' as ' so JOSM translations work in Osmose too.
+        # A ' is a special character in JOSM, see https://josm.openstreetmap.de/wiki/Translations
+        t = T_(string, *args)
+        return {k: t[k].replace("''", "'") for k in t.keys()}
 
 #regexp_test(regexp, string)
 #    test if string matches pattern regexp [since 5699]


frodrigo:
Thank you.

Issue #1950, opened 2023-07-09
Flag urls with click trackers


ivanbranco:
I would like to propose a check to report click trackers in urls, such as: https://overpass-turbo.eu/s/1xaq

Famlam:
Sounds like a nice idea (such tracker query parameters don't make sense to have in OSM.
I wonder if we'd be allowed to use the list from here:
https://github.com/duckduckgo/privacy-configuration/blob/main/features/tracking-parameters.json

Famlam:
It's implemented and deployed :)
https://osmose.openstreetmap.fr/nl/map/#item=3093&source=&class=30934&username=&bbox=&zoom=7&lat=50.1&lon=6.507&level=1%2C2%2C3

ivanbranco:
Thanks for your work

PR #1998, opened 2023-08-25
Detect tracking parameters in URLs

Famlam:
Implements #1950

This PR detects tracking parameters in URLs, such as fbclid or utm_campaign, and strips them from the URL.
~It uses the list of https://github.com/duckduckgo/privacy-configuration/blob/main/features/tracking-parameters.json, which is CC BY-NC-SA 4.0 licensed. Based upon my understanding, we can use this for the detection as long as we refer to its origin (via `resource=*`) as we're not commercial (but mind, I'm not a legally educated person).~

As the easiest way to parse the URL was to use urlsplit, I also added a warning (class 30933) for broken URLs that give a ValueError in that function. These are however pretty rare, so I suspect it'll barely detect something. So if you prefer, I can also skip over those.

### DIFF
--- a/plugins/Website.py
+++ b/plugins/Website.py
@@ -22,7 +22,8 @@
 from modules.OsmoseTranslation import T_
 from plugins.Plugin import Plugin
 from modules.Stablehash import stablehash64
-
+from urllib.parse import urlsplit, urlunsplit
+from modules.downloader import urlread
 
 class Website(Plugin):
 
@@ -39,14 +40,28 @@ class Website(Plugin):
         import re
         # From RFC 1738 paragraph 2.1
         self.HasScheme = re.compile(r"^[a-zA-Z0-9.+-]+://")
+        self.strippable_queryparameters = self._load_trackingparameter_list()
 
         self.errors[30931] = self.def_class(item = 3093, level = 2, tags = ['value', 'fix:chair'],
             title = T_('The URL contains a space'))
         self.errors[30932] = self.def_class(item = 3093, level = 3, tags = ['value', 'fix:chair'],
             title = T_('The URL does not have a valid scheme'))
+        self.errors[30933] = self.def_class(item = 3093, level = 3, tags = ['value', 'fix:chair'],
+            title = T_('Invalid URL'))
+        self.errors[30934] = self.def_class(item = 3093, level = 3, tags = ['value', 'fix:chair'],
+            title = T_('Tracking parameter in URL'),
+            fix = T_('Strip the tracking parameter from the URL. Verify that the URL still works afterwards.'),
+            resource = 'https://github.com/mpchadwick/tracking-query-params-registry')
 
     def _bad_url(self, tag, tags):
         return T_("Bad URL {k}={v}", k = tag, v = tags[tag])
+
+    def _load_trackingparameter_list(self):
+        # Permission to use via https://github.com/mpchadwick/tracking-query-params-registry/issues/15
+        url = "https://raw.githubusercontent.com/mpchadwick/tracking-query-params-registry/master/_data/params.csv"
+        csvlines = urlread(url, 30).split("\n")[1:] # Discard first line, it contains headers
+        # Filter entries of len<=2 for safety. First column contains the parameter
+        return set(filter(lambda p: len(p) > 2, map(lambda line: line.split(",", 1)[0].strip(), csvlines)))
 
     def check(self, tags):
         err = []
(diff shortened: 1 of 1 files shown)


frodrigo:
> which is CC BY-NC-SA 4.0 licensed. Based upon my understanding, we can use this for the detection as long as we refer to its origin (via resource=*) as we're not commercial (but mind, I'm not a legally educated person).

I'm unsure about that. As OSM is not `NC`. Maybe is better to ask ?

cc @jocelynj

jocelynj:
Are you sure about the license?

I see on https://github.com/duckduckgo/privacy-configuration/blob/main/LICENSE that it is using Apache license v2, which is an open-source license.  CC BY-NC-SA 4.0 seems to be only about DuckDuckGo logos and marks.

frodrigo:
> Are you sure about the license?
> 
> I see on https://github.com/duckduckgo/privacy-configuration/blob/main/LICENSE that it is using Apache license v2, which is an open-source license. CC BY-NC-SA 4.0 seems to be only about DuckDuckGo logos and marks.

It is not the same as here https://github.com/duckduckgo/privacy-configuration/tree/main#licensing

jocelynj:
So, we have two different information about license :(

It would be better to ask which license is applied, and if CC BY-NC-SA is confirmed, to ask for explicit authorization.

In the meantime, we could limit to utm_* variables, which can be found in a free license here: https://en.wikipedia.org/wiki/UTM_parameters

Famlam:
> It would be better to ask which license is applied, and if CC BY-NC-SA is confirmed, to ask for explicit authorization.

Drafted an email on Matrix. If that's fine, we could leave it for a few weeks on draft until we get a confirmation?

Famlam:
New implementation using the tracking parameter list from https://github.com/mpchadwick/tracking-query-params-registry is ready :)

plugins/Website.py
```diff
@@ -39,15 +40,29 @@ def init(self, logger):
...
 
     def _bad_url(self, tag, tags):
         return T_("Bad URL {k}={v}", k = tag, v = tags[tag])
 
+    def _load_trackingparameter_list(self):
+        # Permission to use via https://github.com/mpchadwick/tracking-query-params-registry/issues/15
+        url = "https://raw.githubusercontent.com/mpchadwick/tracking-query-params-registry/master/_data/params.csv"
+        csvlines = urlread(url, 30).split("\n")[1:] # Discard first line, it contains headers
+        # Filter entries of len<=2 for safety. First column contains the parameter
+        return set(filter(lambda p: len(p) > 2, map(lambda line: line.split(",", 1)[0].strip(), csvlines)))
```
frodrigo, line 64: Use the python CSV parser rather than manual parser

plugins/Website.py
```diff
@@ -66,9 +81,25 @@ def check(self, tags):
...
+                    err.append({"class": 30933, "subclass": stablehash64(tag), "text": T_('Bad URL in `{0}`: {1}', tag, str(e))})
+                    continue
+                queryparams = parsed_url.query.split("&") # not parse_qs/parse_qsl because we don't want to change whether i.e. + is encoded in the fix
+                if any(map(lambda qs: qs.split("=")[0] in self.strippable_queryparameters, queryparams)):
+                    stripped_query = '&'.join(list(filter(lambda qs: qs.split("=")[0] not in self.strippable_queryparameters, queryparams)))
+                    parsed_url = parsed_url._replace(query = stripped_query)
+                    err.append({
+                        "class": 30934, "subclass": stablehash64(tag),
+                        "text": T_('Tracking parameter in `{0}`', tag),
+                        "fix": [{"~": {tag: urlunsplit(parsed_url)}}]
```
frodrigo, line 96:
```suggestion
                        "fix": {"~": {tag: urlunsplit(parsed_url)}}
```

plugins/Website.py
Famlam: Done

plugins/Website.py
Famlam: Done.

plugins/Website.py
```diff
@@ -39,15 +41,30 @@ def init(self, logger):
...
+            resource = 'https://github.com/mpchadwick/tracking-query-params-registry')
 
     def _bad_url(self, tag, tags):
         return T_("Bad URL {k}={v}", k = tag, v = tags[tag])
 
+    def _load_trackingparameter_list(self):
+        # Permission to use via https://github.com/mpchadwick/tracking-query-params-registry/issues/15
+        url = "https://raw.githubusercontent.com/mpchadwick/tracking-query-params-registry/master/_data/params.csv"
+        csvlines = urlread(url, 30).splitlines()[1:] # Discard first line, it contains headers. Use urlread to allow caching the external file
+        csvparsed = csv.reader(csvlines, delimiter=',')
```
frodrigo, line 64: There is an header option to strip it.

Famlam: I'm not able to find that in the python documentation. Also Analyser_Merge simply seems to call `next` (it uses a stream) instead of having a dedicated header option to the csv reader. Could you please point me at the relevant documentation of the build in python csv reader that describes the header option?

frodrigo: Right. I was probably in the Ruby API and not Python.

Famlam: So I assume I can close this conversation?

frodrigo:
Look ok for me. Thank you.

PR #2003, opened 2023-08-28
Remove 'name' tag from TagFix_MultipleValue.py

emersonveenstra:
Remove 'name' from the tags to check in TagFix_MultipleValue.py, because it's already checked more robustly in Name_Multiple


Famlam:
Not sure if it should be removed from that SQL?

frodrigo:
No, should not be removed from the SQL.

emersonveenstra:
okay, changed the sql statement back to how it was

frodrigo:
It is deployed.

PR #2007, opened 2023-08-30
Add French power tower line management merge analysis

### LINKED ISSUE #457: French power TSO location:transition towers
Hi all,

A recently recommended trick from the French TSO GIS team regarding the "air to ground" connections between power lines and underground cables.
They publish a layer called "underground crossing points" which I thought mentions the air to ground towers but actually not.

It may be possible to add a check on power=tower + location:transition=yes by finding connections between towers, underground cables and overhead lines layers.
All matching connections should be tagged with location:transition=yes on OSM.
See : https://wiki.openstreetmap.org/wiki/Tag:location:transition%3Dyes

Osmose already warns about missing towers from the same tower layers on item #8290 class 1.
Missing location:transition=yes should be added on the same items with such a query :

```
WITH air_to_ground_connections AS (select ST_Intersection(cable.geom, line.geom) AS point FROM power_lines line JOIN power_cables cable ON ST_Touches(line.geom, cable.geom))
SELECT towers.* FROM towers JOIN air_to_ground_connections ON ST_Contains(ST_Buffer(towers.geom::geography,5)::geometry, air_to_ground_connections.point)
```

Let me know if you have any issue regarding this point.
TSO doesn't plan to publish additional data-set or additional attribute on towers layer to cover this use case.

All the best

flacombe:
I propose to add a new merge analysis to conflate the file I've prepared based upon RTE opendata mentioned in #457 and existing (only) power supports in OSM.

It aims to add location:transition and line_management tags on particular power supports.

It is only supposed to produce update warnings, since this analysis only covers existing osm features.
If missing, they will be handled by 8290/1 analysis.

Fixes #457


frodrigo:
You also need to add it to configuration

flacombe:
That's right, I always forget about it. It's done

PR #2013, opened 2023-09-06
Fix substations select in 7190 and 8280 analysis

### LINKED ISSUE #2009: "Power substation is not known from operator" give too many false positives
https://osmose.openstreetmap.fr/en/issues/open?item=7190&class=2

This analyser (7190/2) is giving issues for all substations not known by RTE and not tagged with `substation=minor_distribution`.
Perhaps we should limit this to substations with `operator=RTE`?

Ping @flacombe

### LINKED ISSUE #2011: Missing generation substations in 8280/1
Previous fixes on 8280/1 analysis already shown good results:
https://osmose.openstreetmap.fr/fr/issues/graph.png?item=8280&class=1&start_date=2023-08

False positives remain on generation substations, which still miss from analysis select.
This area has good tags and operator=EDF is accurate
https://www.openstreetmap.org/way/148564021
This warning is a false positive: https://osmose.openstreetmap.fr/fr/issue/f0c9044f-1aef-35ef-3b5a-3888fefe6353

`generation` and `traction` should be added to the select here https://github.com/flacombe/osmose-backend/blob/master/analysers/analyser_merge_power_substation_FR.py#L51

flacombe:
Second fix for substations select with a little addition to Analyser_merge as well.

Fix #2009 
Fix #2011

### DIFF
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -1062,7 +1062,7 @@ class Select:
                     clauses.append("NOT " + k_not_exists)
                     clauses.append(v(k_value))
                 elif isinstance(v, list):
-                    actualList = filter(lambda i: i is not False, v)
+                    actualList = list(filter(lambda i: i is not False, v))
                     if len(actualList) > 1:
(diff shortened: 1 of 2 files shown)


frodrigo:
Merged manually and deployed.

PR #2018, opened 2023-09-10
Partial rewrite of deadend oneway detection

Famlam:
- Ensures that all simple deadends are found and marked on that exact spot (as the error position is well defined). Previously not all deadends were found (e.g. node 44643223) or they resulted in a trail of misplaced entries (see issue 1670). Now the "trail" will only be generated in case there's a loop of inescapable oneways but no clear failure point
- Avoid propagating border issues such as in issue 1949 by treating nodes outside the extract as valid

It basically splits the analyser for class 3 in two steps:
1. First find all obvious deadends due to oneways (points where you cannot arrive or depart)
2. Then traverse through the oneways to see if you can also reach a non-oneway road (previously the analyser only did this step) 

Tested it on 10 countries, didn't find any issues

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -105,7 +105,7 @@ WHERE
 """
 
 sql31 = """
-CREATE INDEX idx_oneway_highway_linestring ON oneway_highway USING GIST(linestring);
+CREATE INDEX idx_oneway_highway_linestring ON oneway_highway USING GIST(linestring) WHERE nid_index=1;
 """
 
 sql32 = """
@@ -116,7 +116,7 @@ SELECT DISTINCT
 FROM
   ways
 WHERE
-  ways.tags != '' AND
+  ways.tags != ''::hstore AND
   (
     (ways.tags?'amenity' AND ways.tags->'amenity' = 'parking') OR
     (ways.tags?'railway' AND ways.tags->'railway' = 'platform') OR
@@ -128,7 +128,7 @@ SELECT
 FROM
   nodes
 WHERE
-  tags != '' AND
+  tags != ''::hstore AND
   (
     (tags?'amenity'  AND tags->'amenity' = 'parking_entrance') OR
     (tags?'entrance' AND tags->'entrance' IN ('garage', 'emergency'))
@@ -177,7 +177,8 @@ FROM
   allowed_end_nodes
   JOIN oneway_highway ON
     -- Note that we cannot literally only get the first and last node, as i.e. in a looped oneway the 'oneway-entrance/exit node' might be in the middle
-    allowed_end_nodes.id = ANY(oneway_highway.nodes)
+    allowed_end_nodes.id = ANY(oneway_highway.nodes) AND
+    oneway_highway.nid_index = 1 -- no need to compare length(nodes) times
 """
 
 sql35 = """
@@ -190,23 +191,19 @@ sql36 = """
 CREATE TEMP TABLE oneway_terminal AS
 SELECT DISTINCT ON (nid)
   oneway.id AS wid,
-  terminal.id AS nid,
-  terminal.geom AS geom,
-  terminal.id = oneway.nodes[oneway.length] AS is_oneway_last_node
+  oneway.nodes[oneway.nid_index] AS nid,
+  (SELECT geom FROM nodes WHERE id = oneway.nodes[oneway.nid_index]) AS geom,
+  oneway.nid_index = oneway.length AS is_oneway_last_node
 FROM
   oneway_highway AS oneway
-  JOIN nodes AS terminal ON
-    (
-      terminal.id = oneway.nodes[1] OR
-      terminal.id = oneway.nodes[oneway.length]
-    )
   LEFT JOIN allowed_oneway_end_nodes ON
-    terminal.id = allowed_oneway_end_nodes.id
+    oneway.nodes[oneway.nid_index] = allowed_oneway_end_nodes.id
 WHERE
+  oneway.nid_index IN (1, oneway.length) AND
   allowed_oneway_end_nodes IS NULL AND
-  array_length(array_positions(oneway.nodes, terminal.id), 1) = 1 -- exclude end nodes folding back into the same way, e.g. in P or O-shaped ways
+  array_length(array_positions(oneway.nodes, oneway.nodes[oneway.nid_index]), 1) = 1 -- exclude end nodes folding back into the same way, e.g. in P or O-shaped ways
 ORDER BY
-  terminal.id,
+  nid,
   oneway.id
 """
 
@@ -222,6 +219,7 @@ FROM
   LEFT JOIN oneway_highway AS other_highway ON
     oneway_terminal.nid = ANY(other_highway.nodes) AND
     oneway_terminal.geom && other_highway.linestring AND -- much faster than comparison with nid
+    other_highway.nid_index = 1 AND -- no need to compare length(nodes) times
     oneway_terminal.wid != other_highway.id AND
     (
       array_length(array_positions(other_highway.nodes, oneway_terminal.nid), 1) != 1 OR -- end node of loop/self-intersecting way that also occurs elsewhere
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -23,6 +23,18 @@
 from modules.OsmoseTranslation import T_
 from .Analyser_Osmosis import Analyser_Osmosis
 
+sql10 = """
+-- From https://wiki.postgresql.org/wiki/Array_reverse
+-- Inverts the sequence of the elements in an array
+CREATE OR REPLACE FUNCTION array_reverse(anyarray) RETURNS anyarray AS $$
+SELECT ARRAY(
+    SELECT $1[i]
+    FROM generate_subscripts($1,1) AS s(i)
+    ORDER BY i DESC
+);
+$$ LANGUAGE 'sql' STRICT IMMUTABLE;
+"""
+
 sql20 = """
 SELECT
     wid,
@@ -77,8 +89,10 @@ sql30 = """
 CREATE TEMP TABLE oneway_highway AS
 SELECT
   id,
-  nodes,
-  tags?'oneway' AND tags->'oneway' = '-1' AS oneway_reverse,
+  CASE -- If oneway=-1, pretent the way is the other way around
+    WHEN tags?'oneway' AND tags->'oneway' = '-1' THEN array_reverse(nodes)
+    ELSE nodes
+  END AS nodes,
   linestring,
   array_length(nodes, 1) AS length,
   generate_subscripts(nodes, 1) AS nid_index
@@ -178,10 +192,7 @@ SELECT DISTINCT ON (nid)
   oneway.id AS wid,
   terminal.id AS nid,
   terminal.geom AS geom,
-  CASE
-    WHEN oneway_reverse THEN terminal.id = oneway.nodes[1]
-    ELSE terminal.id = oneway.nodes[oneway.length]
-  END AS is_oneway_last_node
+  terminal.id = oneway.nodes[oneway.length] AS is_oneway_last_node
 FROM
   oneway_highway AS oneway
   JOIN nodes AS terminal ON
@@ -215,17 +226,8 @@ FROM
     (
       array_length(array_positions(other_highway.nodes, oneway_terminal.nid), 1) != 1 OR -- end node of loop/self-intersecting way that also occurs elsewhere
       (
-        other_highway.oneway_reverse AND
-        (
-          (NOT oneway_terminal.is_oneway_last_node AND oneway_terminal.nid != other_highway.nodes[other_highway.length]) OR
-          (oneway_terminal.is_oneway_last_node AND oneway_terminal.nid != other_highway.nodes[1])
-        )
-      ) OR (
-        NOT other_highway.oneway_reverse AND
-        (
-          (oneway_terminal.is_oneway_last_node AND oneway_terminal.nid != other_highway.nodes[other_highway.length]) OR
-          (NOT oneway_terminal.is_oneway_last_node AND oneway_terminal.nid != other_highway.nodes[1])
-        )
+        (oneway_terminal.is_oneway_last_node AND oneway_terminal.nid != other_highway.nodes[other_highway.length]) OR
+        (NOT oneway_terminal.is_oneway_last_node AND oneway_terminal.nid != other_highway.nodes[1])
       )
     )
 WHERE
@@ -244,10 +246,7 @@ FROM (
     id AS wid,
     nid_index,
     length,
-    CASE
-      WHEN oneway_reverse THEN nodes[length - nid_index + 1]
-      ELSE nodes[nid_index]
-    END AS nid
+    nodes[nid_index] AS nid
   FROM
     oneway_highway
   ) AS t
@@ -423,6 +422,7 @@ Ensure that `service=drive-through` is the correct tag.''')),
         elif not boundary_relation:
           boundary_relation = "(0)"
 
+        self.run(sql10)
         self.run(sql30)
         self.run(sql31)
         self.run(sql32.format(boundary_ids=boundary_relation))
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -117,7 +117,7 @@ WHERE
   tags != '' AND
   (
     (tags?'amenity'  AND tags->'amenity' = 'parking_entrance') OR
-    (tags?'entrance' AND tags->'entrance'!= 'garage')
+    (tags?'entrance' AND tags->'entrance' IN ('garage', 'emergency'))
   )
 UNION ALL
 SELECT DISTINCT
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +71,227 @@
...
+UNION ALL
+SELECT
+  id
+FROM
+  nodes
+WHERE
+  tags != '' AND
+  (
+    (tags?'amenity'  AND tags->'amenity' = 'parking_entrance') OR
+    (tags?'entrance' AND tags->'entrance'!= 'garage')
```
frodrigo, line 120:
```suggestion
    (tags?'entrance' AND tags->'entrance' != 'garage')
```

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +71,227 @@
...
+UNION ALL
+-- Include nodes outside the extract; the recursive check will otherwise fail on oneways originating outside of the extract. See #1949
+SELECT DISTINCT
+  way_nodes.node_id
+FROM
+  way_nodes
+  LEFT JOIN nodes AS n ON
+    n.id = way_nodes.node_id
+WHERE
+  n IS NULL
```
frodrigo, line 146: Look expensive to me, doesn't ? Maybe there is other ways to do this.

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +71,227 @@
...
+  JOIN nodes AS terminal ON
+    (
+      terminal.id = oneway.nodes[1] OR
+      terminal.id = oneway.nodes[oneway.length]
+    )
+  LEFT JOIN allowed_oneway_end_nodes ON
+    terminal.id = allowed_oneway_end_nodes.id
+WHERE
+  allowed_oneway_end_nodes IS NULL AND
+  array_length(array_positions(oneway.nodes, terminal.id), 1) = 1 -- exclude end nodes folding back into the same way
```
frodrigo, line 192: Not sure to understand this condition ? Why not just test oneway is a loop ?

analysers/analyser_osmosis_highway_deadend.py
Famlam: It is the most expensive part indeed (if you exclude sql20, which I left untouched). Do you think it would be better to:
1. Get all ways that cross an extract boundary (like we do here: https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_osmosis_highway_floating_islands.py#L48-L60 )
2. Then 'unnest' all nodes of that way into this table? 

Note: we can't check if the nodes are NULL then: it may cross the boundary and directly after that become a different way (still in the extract, but outside the boundary), which can have the NULL nodes. Hence we'd have to grab all nodes of that boundary-crossing highway

analysers/analyser_osmosis_highway_deadend.py
Famlam: Because it can also be a "P-shaped" way. In that case, the point in the middle (where the circle of the P comes back to the vertical line) should be excluded (but not the tip of the "P")

frodrigo: Ok, Thank I understand.

Using the start or end node as P junction point depend on the way of the onway. It a detail, it may be ignored.

analysers/analyser_osmosis_highway_deadend.py
frodrigo: The 1 Look good to me. Way fully outside of the boundary will be clipped anyway by later result sanitation.

Famlam: I'll have a look later this week!

analysers/analyser_osmosis_highway_deadend.py
Famlam: Thanks! It's actually a typo, I wanted to retain the old behavior of the analyser on `entrance` nodes, but accidentally typed `!=` instead of `=`. Fixed.  (I also decided to whitelist entrance=emergency when fixing this: I don't want to encourage people to draw explicit paths to emergency exits while commonly there is no defined path to such an exit in the inside of a building, e.g. in a hall where they're "just along the sides". Drawing explicit paths that you don't see in the real location isn't good)

analysers/analyser_osmosis_highway_deadend.py
Famlam: Done! It's about a factor 2.5 faster on larger extracts (like japan_chubu)

analysers/analyser_osmosis_highway_deadend.py
Famlam: I clarified the comment a bit.

I would personally prefer to keep this in rather than having to use `NOT ST_IsClosed(linestring) AND NOT ST_IsSimple(linestring)`. It seems to be fast, and it's also more accurate than ST_IsSimple about which node to whitelist.

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +71,231 @@
...
+  oneway.id AS wid,
+  terminal.id AS nid,
+  terminal.geom AS geom,
+  CASE
+    WHEN oneway_reverse THEN terminal.id = oneway.nodes[1]
+    ELSE terminal.id = oneway.nodes[oneway.length]
+  END AS is_oneway_last_node
+FROM
+  oneway_highway AS oneway
+  JOIN nodes AS terminal ON
```
frodrigo, line 187: I think you can remove the join on `nodes` table here. As you only need it for `id`.

The geom can be fetch latter only on selected node ids.

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +71,231 @@
...
+FROM
+  oneway_terminal
+  LEFT JOIN oneway_highway AS other_highway ON
+    oneway_terminal.nid = ANY(other_highway.nodes) AND
+    oneway_terminal.geom && other_highway.linestring AND -- much faster than comparison with nid
+    oneway_terminal.wid != other_highway.id AND
+    (
+      array_length(array_positions(other_highway.nodes, oneway_terminal.nid), 1) != 1 OR -- end node of loop/self-intersecting way that also occurs elsewhere
+      (
+        other_highway.oneway_reverse AND
```
frodrigo, line 218: Maybe you just can reverse the nodes in `oneway_highway` rather than use `oneway_reverse`?

Famlam: Good suggestion, implemented (and much simpler indeed)

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +83,218 @@
...
+
+sql32 = """
+-- Collect all nodes that could be possible (valid) end nodes
+CREATE TEMP TABLE allowed_end_nodes AS
+SELECT DISTINCT
+  unnest(nodes) AS id
+FROM
+  ways
+WHERE
+  ways.tags != '' AND
```
frodrigo, line 119: Keep `::hstore`, doesn'it ?

analysers/analyser_osmosis_highway_deadend.py
Famlam: I think you're correct. A local implementation works, I'll upload it tomorrow if the results on real data also look good (and fast) :)

analysers/analyser_osmosis_highway_deadend.py
Famlam: Yes, right, should've been `''::hstore`. Will include it in the next commit

analysers/analyser_osmosis_highway_deadend.py
Famlam: Done!

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -71,121 +83,216 @@
...
+DROP TABLE allowed_end_nodes;
+"""
+
+sql36 = """
+-- All end nodes of oneway streets
+CREATE TEMP TABLE oneway_terminal AS
+SELECT DISTINCT ON (nid)
+  oneway.id AS wid,
+  oneway.nodes[oneway.nid_index] AS nid,
+  (SELECT geom FROM nodes WHERE id = oneway.nodes[oneway.nid_index]) AS geom,
```
frodrigo, line 195: Not sure it is efficient.
Not sure if the planner will do the same or not. But I was more thinking about a join on the result of the first select.

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -194,21 +301,47 @@
...
+    wid,
+    nid,
+    geom
+  FROM
+    results_straightforward
+  UNION ALL
+  SELECT
+    wid,
+    nid,
+    (SELECT geom FROM nodes WHERE id = nid) AS geom
```
frodrigo, line 335: Better use a join.

Famlam: Sure, this was just because it was already like that in the original implementation ( https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_osmosis_highway_deadend.py#L202 ), so I didn't change it, but in terms of execution it's almost equivalent:

`(SELECT geom FROM nodes WHERE id = nid) AS geom` ->
```
['Unique  (cost=10538.70..10553.30 rows=200 width=48) (actual time=0.315..0.362 rows=205 loops=1)']
['  ->  Sort  (cost=10538.70..10546.00 rows=2920 width=48) (actual time=0.314..0.326 rows=205 loops=1)']
['        Sort Key: results_straightforward.nid, results_straightforward.wid']
['        Sort Method: quicksort  Memory: 47kB']
['        ->  Result  (cost=0.00..10370.62 rows=2920 width=48) (actual time=0.042..0.259 rows=205 loops=1)']
['              ->  Append  (cost=0.00..6691.42 rows=2920 width=48) (actual time=0.040..0.137 rows=205 loops=1)']
['                    ->  Seq Scan on results_straightforward  (cost=0.00..20.70 rows=1070 width=48) (actual time=0.039..0.066 rows=202 loops=1)']
['                    ->  Seq Scan on results_recursive  (cost=0.00..6656.12 rows=1850 width=48) (actual time=0.028..0.053 rows=3 loops=1)']
['                          SubPlan 1']
['                            ->  Index Scan using pk_nodes on nodes  (cost=0.56..3.58 rows=1 width=32) (actual time=0.014..0.015 rows=1 loops=3)']
['                                  Index Cond: (id = results_recursive.nid)']
['Planning Time: 0.180 ms']
['Execution Time: 0.442 ms']
```

`JOIN nodes ON nodes.id = results_recursive.nid` ->
```
['Unique  (cost=10534.70..10549.30 rows=200 width=48) (actual time=0.306..0.354 rows=205 loops=1)']
['  ->  Sort  (cost=10534.70..10542.00 rows=2920 width=48) (actual time=0.305..0.318 rows=205 loops=1)']
['        Sort Key: results_straightforward.nid, results_straightforward.wid']
['        Sort Method: quicksort  Memory: 47kB']
['        ->  Result  (cost=0.00..10366.63 rows=2920 width=48) (actual time=0.053..0.246 rows=205 loops=1)']
['              ->  Append  (cost=0.00..6687.43 rows=2920 width=48) (actual time=0.051..0.134 rows=205 loops=1)']
['                    ->  Seq Scan on results_straightforward  (cost=0.00..20.70 rows=1070 width=48) (actual time=0.050..0.078 rows=202 loops=1)']
['                    ->  Nested Loop  (cost=0.56..6633.63 rows=1850 width=48) (actual time=0.019..0.039 rows=3 loops=1)']
['                          ->  Seq Scan on results_recursive  (cost=0.00..28.50 rows=1850 width=16) (actual time=0.002..0.002 rows=3 loops=1)']
['                          ->  Index Scan using pk_nodes on nodes  (cost=0.56..3.57 rows=1 width=40) (actual time=0.010..0.010 rows=1 loops=3)']
['                                Index Cond: (id = results_recursive.nid)']
['Planning Time: 0.206 ms']
['Execution Time: 0.449 ms']
```

analysers/analyser_osmosis_highway_deadend.py
Famlam: > Not sure it is efficient.

The old implementation gave (for japan_chubu, the slowest of the extracts I tested on):
```
['Unique  (cost=2565300.11..2565300.21 rows=21 width=49) (actual time=10072.100..10127.713 rows=28966 loops=1)']
['  ->  Sort  (cost=2565300.11..2565300.16 rows=21 width=49) (actual time=10072.097..10092.663 rows=371818 loops=1)']
['        Sort Key: terminal.id, oneway.id']
['        Sort Method: quicksort  Memory: 41337kB']
['        ->  Merge Left Join  (cost=2564506.66..2565299.65 rows=21 width=49) (actual time=9372.642..9927.462 rows=371818 loops=1)']
['              Merge Cond: (terminal.id = allowed_oneway_end_nodes.id)']
['              Filter: (allowed_oneway_end_nodes.* IS NULL)']
['              Rows Removed by Filter: 458495']
['              ->  Sort  (cost=2548983.05..2548993.68 rows=4254 width=84) (actual time=9284.612..9491.013 rows=830313 loops=1)']
['                    Sort Key: terminal.id']
['                    Sort Method: external merge  Disk: 165400kB']
['                    ->  Nested Loop  (cost=2.86..2548726.64 rows=4254 width=84) (actual time=0.082..8367.067 rows=830313 loops=1)']
['                          ->  Seq Scan on oneway_highway oneway  (cost=0.00..34089.47 rows=425447 width=44) (actual time=0.036..407.462 rows=425447 loops=1)']
['                          ->  Bitmap Heap Scan on nodes terminal  (cost=2.86..5.90 rows=1 width=40) (actual time=0.013..0.017 rows=2 loops=425447)']
['                                Recheck Cond: ((id = (oneway.nodes)[1]) OR (id = (oneway.nodes)[oneway.length]))']
['                                Filter: (array_length(array_positions(oneway.nodes, id), 1) = 1)']
['                                Rows Removed by Filter: 0']
['                                Heap Blocks: exact=742047']
['                                ->  BitmapOr  (cost=2.86..2.86 rows=2 width=0) (actual time=0.009..0.009 rows=0 loops=425447)']
['                                      ->  Bitmap Index Scan on pk_nodes  (cost=0.00..1.43 rows=1 width=0) (actual time=0.004..0.004 rows=1 loops=425447)']
['                                            Index Cond: (id = (oneway.nodes)[1])']
['                                      ->  Bitmap Index Scan on pk_nodes  (cost=0.00..1.43 rows=1 width=0) (actual time=0.003..0.003 rows=1 loops=425447)']
['                                            Index Cond: (id = (oneway.nodes)[oneway.length])']
['              ->  Sort  (cost=15523.61..15909.37 rows=154305 width=40) (actual time=87.495..131.298 rows=575799 loops=1)']
['                    Sort Key: allowed_oneway_end_nodes.id']
['                    Sort Method: quicksort  Memory: 18200kB']
['                    ->  Seq Scan on allowed_oneway_end_nodes  (cost=0.00..2226.05 rows=154305 width=40) (actual time=0.058..37.032 rows=154305 loops=1)']
['Planning Time: 0.222 ms']
['Execution Time: 10136.535 ms']
```

The new implementation gives:
```
['Unique  (cost=41251.58..41251.98 rows=21 width=49) (actual time=5192.359..5202.764 rows=28970 loops=1)']
['  ->  Sort  (cost=41251.58..41251.78 rows=81 width=49) (actual time=5192.356..5195.468 rows=50519 loops=1)']
['        Sort Key: (oneway.nodes[oneway.nid_index]), oneway.id']
['        Sort Method: quicksort  Memory: 5483kB']
['        ->  Nested Loop Left Join  (cost=10.62..41249.01 rows=81 width=49) (actual time=4.009..5158.143 rows=50519 loops=1)']
['              Filter: (allowed_oneway_end_nodes.* IS NULL)']
['              Rows Removed by Filter: 57880']
['              ->  Seq Scan on oneway_highway oneway  (cost=0.00..39407.56 rows=21 width=48) (actual time=0.037..442.005 rows=108399 loops=1)']
['                    Filter: (((nid_index = 1) OR (nid_index = length)) AND (array_length(array_positions(nodes, nodes[nid_index]), 1) = 1))']
['                    Rows Removed by Filter: 317048']
['              ->  Bitmap Heap Scan on allowed_oneway_end_nodes  (cost=10.62..66.14 rows=772 width=40) (actual time=0.006..0.006 rows=1 loops=108399)']
['                    Recheck Cond: ((oneway.nodes)[oneway.nid_index] = id)']
['                    Heap Blocks: exact=57880']
['                    ->  Bitmap Index Scan on idx_allowed_oneway_nodes_id  (cost=0.00..10.42 rows=772 width=0) (actual time=0.003..0.003 rows=1 loops=108399)']
['                          Index Cond: (id = (oneway.nodes)[oneway.nid_index])']
['              SubPlan 1']
['                ->  Index Scan using pk_nodes on nodes  (cost=0.56..3.58 rows=1 width=32) (actual time=0.077..0.077 rows=1 loops=50519)']
['                      Index Cond: (id = (oneway.nodes)[oneway.nid_index])']
['Planning Time: 3.469 ms']
['Execution Time: 5204.168 ms']
```

So it's definitely an significant improvement.

> But I was more thinking about a join on the result of the first select.

I'm unfortunately not sure what you mean?
You want me to change:
```sql
SELECT
  ...
  (SELECT ... ) AS geom
FROM
  ...
```
to:
```sql
SELECT
  t.*,
  nodes.geom
FROM (
   ... current code minus the geom ...
) AS t
JOIN nodes ON
  t.nid == nodes.id
```
?

frodrigo: > I'm unfortunately not sure what you mean?
You want me to change:

Yes that my point.

Famlam: Ok, done :)

frodrigo:
@Famlam it is now ok for me, can we merge it ?

Famlam:
Yes, I think it's ready to be merged.

frodrigo:
Thank you ! Merged and deployed.

frodrigo:
The updated data look nice to me.

Famlam:
Agree, but I think I'll need to add `aerialway=*` connections to allowed_end_nodes, based on the false positive reports of today.

PR #2028, opened 2023-09-17
Fix missing space

Famlam:
Add a missing space. (My bad, introduced in https://github.com/osm-fr/osmose-backend/commit/0a51489621af1433162522d5b91fe9ddbbfe20b2 )

This string was not translated in any language yet (with the translations in the `dev` branch). Hence, I guess we can just simply replace it by one with a space.

@jocelynj Since there's a lot of translation activity in the HU strings: could you please double-check that it's not recently translated, when you're synchronizing the translations with Transifex?


frodrigo:
I think small changes are not an issue. The string are just flagged are fuzzy.

Issue #1985, opened 2023-08-17
Add more imcompatible: picnic


frodrigo:
Add imcompatible tags to TagRemove_Incompatibles.py
* leisure=picnic_table + tourism=picnic_site

Famlam:
Maybe better to add this ([and the other hardcoded combinations](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/TagRemove_Incompatibles.py#L124-L128)) to the [mapcss rules](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/TagFix_MultipleTag2.validator.mapcss) so that they can be loaded into JOSM too?

PR #2040, opened 2023-10-03
Add incompatibles as mapcss rules

Famlam:
See #1985

1. Convert existing hardcoded rules in TagRemove_Incompatibles.py to mapcss (allows loading it also in i.e. JOSM)
  a. For `crossing=no`: also add support for `crossing=informal`; only check it for nodes; add fix suggestion
2. Adds a warning for `leisure=picnic_table + tourism=picnic_site` (#1985)
3. Adds a warning for `roller_coaster=track + attraction=roller_coaster` (spin-off from #1960)

Since the 4 rules would conflict with each other if I did them as separate PRs, I did it in one PR, but every commit (except the python file regeneration) can be considered a separate PR.


frodrigo:
Ok for me. Thank you.

PR #2065, opened 2023-11-07
Remove diff from osmosis_multipolygon sql30

Famlam:
sql30 relies on `COUNT(*)`, but in diff mode not all joined elements are present in the `touched_*` datasets, hence the count may be lower than the true count


frodrigo:
Thank you.

PR #2070, opened 2023-11-14
Fix inconsistencies in date check

Famlam:
1. Although the regex checks deny any date past the year 2999, the `convert2date` function check (which is called if the regex checks fail) still allowed any year except 9999. As there's no logical reason to expect any years >=3000 in OSM, mark them as bad too. (Reason: in the Dutch community we spotted some very strange years in the `start_date` tag of buildings, containing start_dates far in the future)

2. Add some extra safety for the "~", "s" and " BC" year checks. If the string consists only of those characters it should be treated as an invalid date. (The check function returns True on empty strings due to the handling of e.g. `1920..`, so a string `~` would return True after stripping the `~`)

### DIFF
--- a/plugins/Date.py
+++ b/plugins/Date.py
@@ -49,7 +49,7 @@ class Date(Plugin):
     def convert2date(self, string):
         try:
             date = dateutil.parser.parse(string, default=self.default_date)
-            if date.year != 9999:
+            if date.year < 3000:
                 return date
         except (ValueError, TypeError, OverflowError):
             pass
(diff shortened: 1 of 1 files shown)


Famlam:
Hmm, I see that 2019-99-77 is also considered a valid date by the regexes, despite that there is no 99th month. @frodrigo I can fix the regexes, but is there a reason why we don't just call `convert2date` all the time?

Famlam:
I'll fix the month/day check in a separate PR. This one is ready for review (next PR with the fix for invalid months is ready on my computer ;) )

frodrigo:
Thank you

PR #2071, opened 2023-11-16
Fixed Improper Method Call: Replaced `NotImplementedError`

fazledyn-or:
## Details
While triaging your project, our bug fixing tool generated the following message(s)-

> In file: [mapcss_lib.py](https://github.com/osm-fr/osmose-backend/blob/dev/mapcss/mapcss_lib.py#L72), class: str_value_, there is a special method `__rmul__` that raises a NotImplementedError. If a special method supporting a binary operation is not implemented it should return NotImplemented. On the other hand, NotImplementedError should be raised from abstract methods inside user defined base classes to indicate that derived classes should override those methods. iCR suggested that the special method `__rmul__` should return NotImplemented instead of raising an exception. An example of how NotImplemented helps the interpreter support a binary operation is here.

### Related Documentation
- [docs.python.org - NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented)
- [docs.python.org - NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError)
- [docs.python.org - Implementing Arithmetic Operators](https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations)

Moreover, we noticed that a method called `__rdiv__` was defined in the same file. It is to be noted that this method belongs to Python 2.0 according to the [documentation](https://docs.python.org/2/reference/datamodel.html#object.__rdiv__). The Python 3 equivalent of that method is [`__rtruediv__`](https://docs.python.org/3/reference/datamodel.html#object.__rtruediv__). In that's the case and you're willing, I can rename the `__rdiv__` method to `__rtruediv__`.

## Changes
- Replaced `NotImplementedError` with `NotImplemented`

## Previously Found & Fixed
- https://www.github.com/SciTools/iris/pull/5544
- https://www.github.com/cupy/cupy/pull/7900
- https://www.github.com/ethereum/web3.py/pull/3080

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/mapcss/mapcss_lib.py
+++ b/mapcss/mapcss_lib.py
@@ -59,7 +59,7 @@ class str_value_(str):
         elif o.__class__ == int:
             return str_value(o - self.to_n())
         else:
-            raise NotImplementedError
+            return NotImplemented
 
     def __sub__(self, o):
         if self.none:
@@ -67,7 +67,7 @@ class str_value_(str):
         elif o.__class__ == int:
             return str_value(self.to_n() - o)
         else:
-            raise NotImplementedError
+            return NotImplemented
 
     def __rmul__(self, o):
         if self.none:
@@ -75,7 +75,7 @@ class str_value_(str):
         elif o.__class__ == int:
             return str_value(o * self.to_n())
         else:
-            raise NotImplementedError
+            return NotImplemented
 
     def __mul__(self, o):
         if self.none:
@@ -83,7 +83,7 @@ class str_value_(str):
         elif o.__class__ == int:
             return str_value(self.to_n() * o)
         else:
-            raise NotImplementedError
+            return NotImplemented
 
     def __rdiv__(self, o):
         if self.none:
@@ -91,7 +91,7 @@ class str_value_(str):
         elif o.__class__ == int:
             return str_value(float(o) / self.to_n())
         else:
-            raise NotImplementedError
+            return NotImplemented
 
     def __truediv__(self, o):
         if self.none:
@@ -99,7 +99,7 @@ class str_value_(str):
         elif o.__class__ == int:
             return str_value(float(self.to_n()) / o)
         else:
-            raise NotImplementedError
+            return NotImplemented
 
     def __lt__(self, o):
         if self.none:
--- a/mapcss/mapcss_lib.py
+++ b/mapcss/mapcss_lib.py
@@ -85,7 +85,7 @@ class str_value_(str):
         else:
             return NotImplemented
 
-    def __rdiv__(self, o):
+    def __rtruediv__(self, o):
         if self.none:
             return None_value
         elif o.__class__ == int:


Famlam:
I tested it with the following code, which obviously should fail as you cannot subtract "8" from a string.
```css
node[x][tag("a")-tag("b")="hello"] {
  throwWarning: "x";
  assertNoMatch: "node a=test b=8 x=abc";
}
```

Compilation of the mapcss code works in either case (the minus is not evaluated during compiling)

Running the code with `raise` fails at the position of `raise` and isn't caught
```
plugins/test.py:31: in node
    try: match = ((mapcss._tag_capture(capture_tags, 0, tags, 'x')) and (mapcss.tag(tags, 'a')-mapcss.tag(tags, 'b') == 'hello'))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = 'test', o = '8'

    def __sub__(self, o):
        if self.none:
            return None_value
        elif o.__class__ == int:
            return str_value(self.to_n() - o)
        else:
>           raise NotImplementedError
E           NotImplementedError

modules/mapcss_lib.py:72: NotImplementedError
```

Running the code with `return NotImplemented` fails at the code point where the minus is evaluated:
```
self = <plugins.test.test object at 0x7f78ebf72c50>, data = {'id': 0, 'lat': 0, 'lon': 0}
tags = {'a': 'test', 'b': '8', 'x': 'abc'}

    def node(self, data, tags):
        capture_tags = {}
        keys = tags.keys()
        err = []


        # node[x][tag("a")-tag("b")="hello"]
        if ('x' in keys):
            match = False
            if not match:
                capture_tags = {}
>               try: match = ((mapcss._tag_capture(capture_tags, 0, tags, 'x')) and (mapcss.tag(tags, 'a')-mapcss.tag(tags, 'b') == 'hello'))
E               TypeError: unsupported operand type(s) for -: 'str_value_' and 'str_value_'

plugins/test.py:31: TypeError
```

I don't have a clear favorite out of the two, but it indeed seems that Python advocates the `return` option. It would differ in case we're evaluating an operator with a `str_value_`-type with a non-`str_value_`-type where the arithmetic operators are defined differently, but I can't think of a test case for this. For example, in the following case (string minus number) the behavior is the same with `raise` or `return` (a silent "no match"):
```css
node[x][tag("a")-number_of_tags()="hello"] {
  throwWarning: "x";
  assertNoMatch: "node a=test b=8 x=abc";
}
```

---

> Moreover, we noticed that a method called __rdiv__ was defined in the same file. It is to be noted that this method belongs to Python 2.0 according to the [documentation](https://docs.python.org/2/reference/datamodel.html#object.__rdiv__). The Python 3 equivalent of that method is [__rtruediv__](https://docs.python.org/3/reference/datamodel.html#object.__rtruediv__). In that's the case and you're willing, I can rename the __rdiv__ method to __rtruediv__.

Seems to be correct based on the documentation (same for `__div__` vs `__truediv__`, but that was already done [prior](https://github.com/osm-fr/osmose-backend/commit/a3b8765b27265a874a4e6bf40dbc9c1f1f0f1e72)) although I don't know any cases where mapcss uses this function actually.

frodrigo:
Ok thank you a lot for looking into this.

> Seems to be correct based on the documentation (same for __div__ vs __truediv__, but that was already done [prior](https://github.com/osm-fr/osmose-backend/commit/a3b8765b27265a874a4e6bf40dbc9c1f1f0f1e72)) although I don't know any cases where mapcss uses this function actually.

Probably forgotten here from the py2 -> py3 migration. I think the method could be removed.

fazledyn-or:
@frodrigo @Famlam thanks for the detailed explanation.
So, should I go and remove the `__rdiv__` method?

frodrigo:
It looks like `__rdiv__` should be `__rtruediv__` in py3.

Famlam:
Correct, the question is mostly whether we want to remove it (since apparently it's never called) or rename it. I'm fine with it either way. I think you mean your preference is to rename it?

fazledyn-or:
In `mapcss_lib.py` file, I can see that the implementation of `__truediv__` is similar to `__rdiv__`. I think it'd be best to just renamed `__rdiv__` to `__rtruediv__`.

frodrigo:
Merged. Thank you.

PR #2073, opened 2023-11-17
Enable diff mode for sql31 of highway_area_access

Famlam:
sql31 is a simple comparison so it should be able to run in diff mode I think
(but please correct me if I'm wrong)

### DIFF
--- a/analysers/analyser_osmosis_highway_area_access.py
+++ b/analysers/analyser_osmosis_highway_area_access.py
@@ -98,8 +98,8 @@ SELECT
   barrier.id,
   ST_AsText(barrier.geom)
 FROM
-  highways
-  JOIN nodes AS barrier ON
+  {0}highways as highways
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_highway_area_access.py
```diff
@@ -94,25 +94,25 @@
...
 sql31 = """
 SELECT
-  highways.id,
+  ways.id,
   barrier.id,
   ST_AsText(barrier.geom)
 FROM
-  highways
-  JOIN nodes AS barrier ON
+  {0}highways as ways
```
frodrigo, line 101: You can also just keep the name of the table ?
```suggestion
  {0}highways as highways
```

Famlam: True, I've updated the PR

frodrigo:
Thank you

PR #2080, opened 2023-11-18
Warn about invalid months/days

Famlam:
Follow-up of #2070
Instead of fixing the regexes, just let `dateutil` take care of it. `dateutil` already took care of the `yyyy-mm` scenario.

See https://github.com/osm-fr/osmose-backend/pull/2070#issuecomment-1811439988

Some examples of newly detected items:
`2017-12-00`	(should be: `2017-12`)
`2023-09-31`	(september has only 30 days)
`2021-22-05`	(there is no 22th month, probably `2021-05-22` is meant)
`1999-00-00`	(should be `1999`)

### DIFF
--- a/plugins/Date.py
+++ b/plugins/Date.py
@@ -41,9 +41,6 @@ class Date(Plugin):
             "mhs:inscription_date", # Heritage
         ]
         self.default_date = datetime.datetime(9999, 12, 1)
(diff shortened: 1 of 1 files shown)


frodrigo:
Ok. Thank you.

PR #2081, opened 2023-11-19
Support `@supports {rule}` in mapcss

Famlam:
Adds basic support for `@supports {}` blocks in mapcss

It doesn't really parse the `@supports` conditions, that would be:
1. Unnecessary difficult :)
2. Give dilemma's, e.g. we try to mimic JOSM's parsing, so should we return true or false on `user-agent: josm`?

Instead, it checks if there is a `-osmoseItemClassLevel` property present, and only if so, it assumes it's safe for Osmose. All other (possibly unsafe) rules are discarded.

Note that the regeneration of mapcss didn't add new rules to transport.validator.mapcss because last time the rule that is now in `@supports` was already present; it was just not in a `@supports` context. (See [this old-new comparison](https://github.com/Jungle-Bus/transport_mapcss/compare/1871b6d55a7ac38c7e7e5b47a822c950a1911404..fc5279aaaf3e4edbb2770c6ba316216fb07a5251))

Note on the changes in the *.g4 file: the whitespace after the `SUPPORTS_NOT` must be present. If not, antlrs generated file will detect the `not` in i.e. `*[not:name][name]["not:name"=*"name"],` from `plugins/notprefix.validator.mapcss` as such a SUPPORTS_NOT, failing to compile after that. (This also happens with `set`, but as we don't have rules with `*[set*` it doesn't cause issues)


frodrigo:
Ok. Thank you.

PR #2092, opened 2023-12-03
New analyser to detect duplicate relation members

Famlam:
This adds an analyser to detect duplicate relation members, such as in relation 9104912 (see issue 2090).

Not sure if it's the best choice for the item number

### DIFF
--- a/analysers/analyser_osmosis_relation_duplicate_member.py
+++ b/analysers/analyser_osmosis_relation_duplicate_member.py
@@ -27,11 +27,13 @@ sql10 = """
 SELECT
   id,
   ST_AsText(relation_locate(id)),
-  array_agg(duplicate)
+  array_agg(duplicate_typeid ORDER BY duplicate_typeid), -- sequence doesn't matter as long as it's constant until relation changes
+  array_agg(duplicate_string ORDER BY duplicate_string) -- keep same sequence as line above
 FROM (
   SELECT
     id,
-    member_type || member_id || ' (' || COUNT(*) || ')' AS duplicate
+    member_type || member_id AS duplicate_typeid,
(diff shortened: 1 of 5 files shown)


analysers/analyser_osmosis_relation_duplicate_member.py
```diff
@@ -0,0 +1,94 @@
...
+
+sql10 = """
+SELECT
+  id,
+  ST_AsText(relation_locate(id)),
+  array_agg(duplicate)
+FROM (
+  SELECT
+    id,
+    member_type || member_id || ' (' || COUNT(*) || ')' AS duplicate
```
frodrigo, line 34: I think it is possible to report the duplicated member object using `any_full`.

Famlam: I was considering this at first (well, with `array_*` instead of `any_*` to avoid many reports for the same relation if there's more than one member duplicated, which is true for about half the cases I tested), but I think it would become confusing as soon as there's more than one duplicate member.

If so the UI would show something like:
```
relation 123
way 123
way 456
way 789
```
which would be confusing because way 123, 456 and 789 are not duplicates. Hence I output it as text, for example: `w123 (2), w456 (3), w789 (2)`.

I would do this if it was possible to repeat the entries that were duplicated, e.g. 
```
relation 123
way 123
way 123
way 456
way 456
way 456
way 789
way 789
```
In that case it would be clear to the user that those are three different repeated members.

Famlam: I'll check in the coming days if I can combine the two approaches.

Famlam: Done :)

frodrigo:
Good to me thank you.

PR #2094, opened 2023-12-05
Allow `disused:type` in Osmose

Famlam:
After writing my reply in issue 2093 I remembered we have a copy of the JOSM rule (with extra `-osmoseDetail`), so we can edit it ourselves. This does not fix the issue (different plugin), it's just related to it

p.s. this is JOSMs reasoning against it, they advocate `state=*` https://josm.openstreetmap.de/ticket/23194


frodrigo:
Not sure about this. There is few `disused:type` and not documented in the wiki.

Famlam:
Ok, it seemed to me like a better approach to avoid more "validator bypasses" like `type=disused:X`. On the other hand, as long as JOSM keeps its rule, the issue will continue to exist.

PR #2098, opened 2023-12-09
Fix #2096

### LINKED ISSUE #2096: Housenumber rule to restrictive for CW / Curacao
source 	9640
item 	2060
class 	10 

After touching a couple of houses in Curaçao which have a house number starting with a letter (verified locally), osmose marks the edit with [2060 street numbers ] [Invalid addr:housenumber value].

I've marked most as invalid, but left one open as an example:
https://osmose.openstreetmap.fr/en/issue/2137130a-e344-63a8-5afe-302c2bd7ef4e

This refers to:
https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagFix_Housenumber.py#L33

Possible solutions

Implement similar rule as for NL for CW, or just add CW to the same set of rules for NL, as these already provide for this scenario and Curaçao actually [is part of](https://en.wikipedia.org/wiki/Cura%C3%A7ao) the Kingdom of the Netherlands so it kinda makes sense that the same applies. 

```
elif country == 'NL':
            # Baseline:
            #   https://imbag.github.io/catalogus/hoofdstukken/attributen--relaties#734-huisnummertoevoeging
            #   (7.3.2 huisnummer, 7.3.3 huisletter and 7.3.4 huisnummertoevoeging)
            # Exceptions to the rule:
            #   https://nl.wikipedia.org/wiki/Huisnummer
            # This pattern isn't exhaustive, but it should catch most weirdness.
            self.housenumber = re.compile(
                # Houseboats, 't/o X' stands for 'opposite X', where 'X' is an address on shore
                r"^(t/o )?"
                # 'Pekela'-style exception, leading letter (e.g., 'C54')
                "(([A-Z][1-9][0-9]{0,3})|"
                # Normal base numbers, no leading zero, not exceeding 5 digits.
                "([1-9][0-9]{0,4}))"
                # Up to four optional extensions (can have leading zeroes in the extension part, e.g., '2K-008')
                "([ -/]?(([0-9]{1,4})|([A-Za-z]{1,5}))){0,4}$")
```

I'm unsure whether this also applies to other islands similarly related to the Kingdom of the Netherlands.

Famlam:


Famlam:
It seems I accidentally did this PR on master instead of dev...

Issue #1471, opened 2022-06-08
Fence type vs. material


hungerburg:
I got a bunch of level 1 errors recently - https://github.com/osm-fr/osmose-backend/blob/34efa0287acfea481d36bf34452cf6662a3f5349/plugins/TagFix_MultipleTag.py#L42

The table here https://wiki.openstreetmap.org/wiki/Key:fence_type recommends it the other way around, or rather, if I understand correctly, fence_type and material are not exclusive, at least in case of "material=concrete | metal | wood"?

I might _add_ a fence_type, but there should not be a need to remove material tag, as in fact, it is true. The message might need change in phrasing, and error level too?

Issue #803, opened 2020-03-20
#303210: barrier=fence and material


maro-21:
I don't think using `material` with `barrier=fence` is incorrect. The preset in iD for fence has both tags: `fence_type` and `material`.

For me, `fence_type` is rather a shape of a fence, so not "wood" or "metal". There could be for example two fences with the same shape but one made of metal and one made of wood.
[Wiki says](https://wiki.openstreetmap.org/wiki/Key:fence_type) "Currently, this tag mixes function, construction type and material of fences."

Also it is very often used
`barrier=fence`: 3,908,705 features
`barrier=fence` + `material`: 22,387 features
https://taginfo.openstreetmap.org/tags/barrier=fence#combinations

Issue 303210: http://osmose.openstreetmap.fr/en/errors/?item=3032&class=303210

I understand the code: it lists fences with "material" tag but without "fence_type". But I don't think it's an issue and should be removed. Because currently there is a contradiction between Osmose and iD preset.
What do you think?
https://github.com/osm-fr/osmose-backend/blob/ffed9fa9b89c52a1854ce3ffdbca4ec4561a5b2f/plugins/TagFix_MultipleTag.py#L150-L151

frodrigo:
Check added after query in #447

Maybe we can raise issue when there is only `material` but no `fence_type`?

maro-21:
Currently there is a contradiction between Osmose and iD preset. People have two field to fill in: fence_type and material. And I think it's correct not to mix the values like "chain_link" and "metal" in one tag.
Typing `fence_type=wood` doesn't say if it's 
https://wiki.openstreetmap.org/wiki/File:Zaun_Freilichtmuseum_Salzburg_02.JPG
https://wiki.openstreetmap.org/wiki/File:Wooden_Fence_-_geograph.org.uk_-_1499882.jpg
or https://wiki.openstreetmap.org/wiki/File:J%C3%A4gerzaun.jpg

But let's wait with this issue, I'll start a discussion on Wiki and ask about it.

maro-21:
> I'll start a discussion on Wiki and ask about it.

I forgot what I was going to ask ;)
But such issue is already at https://wiki.openstreetmap.org/wiki/Talk:Key:fence_type#Material_vs_type

hungerburg:
I second that and the reasoning in #1116 - Steel/Metal/Wood are materials; Barbed_wire/Chain_link/&c are types.

[...]

PR #2099, opened 2023-12-09
Fix #2095, #1471, #803 - fix fence_type vs material warning

### LINKED ISSUE #2095: Should "Fence with material tag" really be level 1?
Currently [fence type](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/TagFix_MultipleTag.py#L35) errors are level 1, which seems to be a high value for what is essentially a cosmetic tag that doesn't affect rendering or routing. I think it should be level 3 to match the similar cosmetic tag errors like `area=yes` on relations, brand name,  missing altitude, etc.

Famlam:
1. Lower the priority of the `Fence with material tag, better use fence_type tag` warning (#2095)
2. Make clear that the intention is to add a fence_type, not to remove the material (e.g. `fence_type=chain_link + material=metal`) (#1471, #803)
3. Convert to mapcss so it can also be loaded into JOSM. (Uses placeholders to avoid people from translating `material` or `fence_type`; also drops the tag `highway` (unrelated) and the tag `fix:chair` (as from my chair I probably won't be able to distinguish values like `barbed_wire` from `wire`; usually it'll require a survey)

### DIFF
--- a/plugins/TagFix_MultipleTag.py
+++ b/plugins/TagFix_MultipleTag.py
@@ -32,8 +32,6 @@ class TagFix_MultipleTag(Plugin):
 
         self.errors[30323] = self.def_class(item = 3032, level = 3, tags = ['tag', 'fix:chair'],
             title = T_('Watch multiple tags'))
-        self.errors[303210] = self.def_class(item = 3032, level = 3, tags = ['tag', 'highway', 'fix:chair'],
-            title = T_('Fence with `material` tag, also add `fence_type`'))
         self.errors[20803] = self.def_class(item = 2080, level = 2, tags = ['tag', 'highway', 'fix:chair'],
             title = T_('Tag highway missing for tracktype or lanes'))
         self.errors[21101] = self.def_class(item = 2110, level = 2, tags = ['tag'],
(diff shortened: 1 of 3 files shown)


frodrigo:
Nice. Thank you.

Famlam:
@frodrigo Did I do something wrong?
It seems that the subtitle of the issues now contains `Fence with material tag, also add fence_type` (as expected), but the title is still `Fence with material tag, better use fence_type tag` (the old title)

Example:
https://osmose.openstreetmap.fr/en/issue/5607e6f7-cd01-7fda-9b08-45610d0c4bca

frodrigo:
> @frodrigo Did I do something wrong? It seems that the subtitle of the issues now contains `Fence with material tag, also add fence_type` (as expected, this is the new text), but the title is still `Fence with material tag, better use fence_type tag` (the old title), and the priority is still "high" (instead of low) and the link still refers to [TagFix_MultipleTag.py](https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagFix_MultipleTag.py#L35) instead of TagFix_MultipleTag2.validator.mapcss.
> 
> Example: https://osmose.openstreetmap.fr/en/issue/5607e6f7-cd01-7fda-9b08-45610d0c4bca

It is now OK.

Famlam:
Thanks! Only the link is incorrect, but that's because the fix hasn't been merged into `master` yet

PR #2109, opened 2023-12-24
Add parking as allowed deadend in analyser_osmosis_highway_deadend

frodrigo:
The issue title is "One way inaccessible or missing parking or parking entrance", but only "parking entrance" was allowed. Just "parking" is missing.


Famlam (approved):
Looks good to me!

frodrigo:
I mainly find this issue with underground parking.

Famlam:
That surprises me, because usually for underground parkings that aren't mapped in detail, I only see the entrance tagged, e.g. then the way ends at amenity=parking_entrance.

Issue #2043, opened 2023-10-11
Way access mismatch relation route=bicycle with highway=pedestrian


blef00fr:
http://osmose.openstreetmap.fr/fr/issue/b36049b5-cc4d-865d-547f-79433092452d

https://www.openstreetmap.org/way/1110580785

https://www.openstreetmap.org/relation/1115348

At least here in France, bicycles are allowed on pedestrian areas: article R431-9 fourth paragraph of the highway code (article R431-9 4e paragraphe du Code de la Route).
https://www.legifrance.gouv.fr/codes/article_lc/LEGIARTI000030851697/2021-08-01

frodrigo:
Should be explicit in OSM
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dpedestrian#Vehicle_access

Famlam:
It's indeed listed as such here: https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions#By_country

blef00fr:
I'm preparing to integrate the rule quoted into the wiki to comply with it.

blef00fr:
I've modified the relevant wiki pages (EN + FR)
https://wiki.openstreetmap.org/w/index.php?title=Tag%3Ahighway%3Dpedestrian&type=revision&diff=2604031&oldid=2582222
https://wiki.openstreetmap.org/w/index.php?title=FR%3ATag%3Ahighway%3Dpedestrian&type=revision&diff=2604028&oldid=1385812
Nevertheless, I've kept the advice to specify the conditions of access despite the implicit authorization. Do you have an opinion on this? Or should I consult the community on a forum?
In any case, is Osmose's warning now timely?

frodrigo:
Ok, but Osmose does not use the access default matrix table and routing probably neither.
Maybe interesting to see if there is machine readable source for such data (dataitem ?)

Famlam:
> Ok, but Osmose does not use the access default matrix table and routing probably neither.

Correct. It also states that on the wiki: _As per 2022, no mainstream routing engine is known to use this page directly to influence routing behavior._
There is also no fixed format (see for example Austria, which has exceptions for Tirol; sometimes foot/pedestrian are combined (Austria), sometimes separate (Belgium); sometimes bicycle is combined with mofa (Brazil); some have practically a full wiki page as footnotes (Netherlands)). So I don't think there's any way of parsing this automatically.

---

We could probably add a fix message:
```
[...]

PR #2111, opened 2023-12-27
Add fix message for route highway mismatch

Famlam:
This is related to #2043.
It adds a fix suggestion to the `Way access mismatch relation route=*` errors

The question whether `highway=pedestrian` for bicycles should be exempted (e.g. the actual issue that was reported) isn't resolved yet. See (from that issue)
> we could just exclude pedestrian. Out of the 32 countries listed on that wiki, 12 permit cyclists on highway=pedestrian. (Just like cycleway is excluded for foot relations because it differs so much per country)


frodrigo:
Thank you.

PR #2132, opened 2024-01-16
Update TagFix_MultipleTag.py

emersonveenstra:
Add [`ruins:`](https://wiki.openstreetmap.org/wiki/Key:ruins:*) and [`destroyed:`](https://wiki.openstreetmap.org/wiki/Key:destroyed:*) lifecycle prefixes to the list of valid prefixes


frodrigo:
Ok, thank you. Merged, but not deployed, yet.

Issue #2135, opened 2024-01-17
Enable TagFix_Opening_Hours for `opening_hours:*`


Famlam:
TagFix_Opening_Hours only checks for `opening_hours`
It can also be enabled for selected `opening_hours:*`, at least the popular ones:
```
opening_hours:kitchen
opening_hours:drive_through
```
(I'm ignoring the `opening_hours:covid19` ones; to my knowledge there's no restrictions anywhere anymore, so I guess this tag should just be purged...)

Famlam:
Deployed

PR #2139, opened 2024-01-22
Enable opening_hours check for subkeys

Famlam:
See #2135
Enable the plugin for selected subkeys of opening_hours


frodrigo:
Thank you

PR #2140, opened 2024-01-25
Update TagFix_Maxspeed.py

Timmy-Tesseract:
Maxspeed on motorways in the Philippines is 100 km/h, on rural roads 80 km/h and on urban roads 40 km/h

https://trb.gov.ph/index.php/faqs/speed-limit
https://lawphil.net/statutes/repacts/ra1964/ra_4136_1964.html


frodrigo:

PR #2148, opened 2024-02-10
Fix typo in item ID

XioNoX:
Was set to charging station instead of bicycle parking.


frodrigo:
Merged and deployed.
Fronend database should be fixed after next data update.

XioNoX:
Thanks, however it looks like the issues are not showing up at all now : 
https://osmose.openstreetmap.fr/en/issues/open?item=8152&source=448196
Neither using the old ID.
I sent https://github.com/osm-fr/osmose-frontend/pull/484 but I don't know if more need to happen.

frodrigo:
New item have to be enabled manually.

Issue #231, opened 2017-09-14
oneway not reachable on motorway


frodrigo:
Report from osm-fr/osmose-frontend/issues/58

See for example:
http://osmose.openstreetmap.fr/de/map/#zoom=18&lat=50.647817&lon=7.334044&layer=Mapnik&overlays=FFFFFFFFFFFFFFFFFFFTT&item=&level=1%2C2%2C3&tags=&fixable=

Cannot find any error at that. There are lot of other of this false positives(?) in osmose.

frodrigo:
Yes. It look like a bug on motorway only.

frodrigo:
http://osmose.openstreetmap.fr/en/errors/?item=1210&class=3
http://osmose.openstreetmap.fr/en/errors/graph.png?item=1210&class=3

PR #2158, opened 2024-03-09
added motorways to oneways

ComradeRamen:
oneway=yes is implied for highway=motorway. See [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway)

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -44,7 +44,7 @@ SELECT
     is_polygon,
     tags->'highway' LIKE '%_link' AS is_link,
     (tags?'junction' AND tags->'junction' = 'roundabout') AS is_roundabout,
-    ((tags->'highway' = 'motorway' AND tags->'oneway' != 'no') OR (tags?'oneway' AND tags->'oneway' IN ('yes', 'true', '1', '-1'))) AS is_oneway,
+    ((tags->'highway' = 'motorway' AND NOT tags?'oneway') OR (tags?'oneway' AND tags->'oneway' IN ('yes', 'true', '1', '-1'))) AS is_oneway,
     CASE
       WHEN tags?'area' THEN tags->'area' != 'no'
       ELSE (is_polygon AND tags->'highway' = 'pedestrian')
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -44,7 +44,7 @@ SELECT
     is_polygon,
     tags->'highway' LIKE '%_link' AS is_link,
     (tags?'junction' AND tags->'junction' = 'roundabout') AS is_roundabout,
-    (tags?'oneway' AND tags->'oneway' IN ('yes', 'true', '1', '-1')) AS is_oneway,
+    ((tags->'highway' = 'motorway' AND tags->'oneway' != 'no') OR tags->'oneway' IN ('yes', 'true', '1', '-1')) AS is_oneway,
     CASE
       WHEN tags?'area' THEN tags->'area' != 'no'
       ELSE (is_polygon AND tags->'highway' = 'pedestrian')
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -44,7 +44,7 @@ SELECT
     is_polygon,
     tags->'highway' LIKE '%_link' AS is_link,
     (tags?'junction' AND tags->'junction' = 'roundabout') AS is_roundabout,
-    ((tags->'highway' = 'motorway' AND tags->'oneway' != 'no') OR tags->'oneway' IN ('yes', 'true', '1', '-1')) AS is_oneway,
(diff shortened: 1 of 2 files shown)


ComradeRamen:
oneway=yes is implied for highway=motorway. See [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dmotorway)

frodrigo:
Good spot. Thank you.
Before merge we need a quick check that will not have side effects on analyzers.

ComradeRamen:
I noticed this omission when there were no motorways in the issues I downloaded.
 
I've been manually sorting the issue:[One way inaccessible or missing parking or parking entrance] so I can find higher classification highways since most of the detections are service roads. 

Would love if the website allowed such sorting natively but I'm not sure if I have the technical expertise for that.

ComradeRamen:
Especially since I recently came across a reversed motorway_link in the US that had been live on the database for ~10 months.

Famlam:
I would leave out the oneway!=no, and only check that there's no oneway tag. If there is a oneway tag, then the "or" behind it should catch it

I think that most analysers have roundabouts and motorways hardcoded, but some use motorways as "starting points"

Famlam:
> Especially since I recently came across a reversed motorway_link in the US that had been live on the database for ~10 months.

Likely this won't be spotted anyway, there's no check against very sharp angles when entering a motorway, and this analyser is among those that use motorways as starting points

ComradeRamen:
> only check that there's no oneway tag. If there is a oneway tag, then the "or" behind it should catch it

Wouldn't this make the code less understandable at a glance; or is this an issue of efficiency?

Famlam:
> > only check that there's no oneway tag. If there is a oneway tag, then the "or" behind it should catch it
> 
> Wouldn't this make the code less understandable at a glance; or is this an issue of efficiency?

I would personally argue it makes it more easy to read as you only have to consider values of oneway once.

Remember there's also other values of oneway, like "alternating" and "reversible", which aren't a true oneway, but still override the default of motorway

ComradeRamen:
> leave out the oneway!=no, and only check that there's no oneway tag.

Done. (I think)

ComradeRamen:
> > Especially since I recently came across a reversed motorway_link in the US that had been live on the database for ~10 months.
> 
> Likely this won't be spotted anyway, there's no check against very sharp angles when entering a motorway, and this analyser is among those that use motorways as starting points

It was a very short segment in the middle of a long ramp. The connecting _links were the correct orientation

ComradeRamen:
Can someone merge this please?

Famlam:
I have no principal objections here, but I think this is why it hasn't been merged yet:

> Before merge we need a quick check that will not have side effects on analyzers.


I haven't had the time for that yet.
(In a second step we probably also have to clean up the analysers that have their own code for motorway)

Famlam:
Please revert the latest change (to analyser_osmosis_highway_deadend), here motorways (and a couple of other ones) are purposefully used as input points

ComradeRamen:
Should it be a separate pull request? Isn't this endpoint inclusion unnecessary since they're now defined as oneways already and thus already included in endpoints?

Famlam:
> Should it be a separate pull request? Isn't this endpoint inclusion unnecessary since they're now defined as oneways already and thus already included in endpoints?

Regarding *_deadend: here they're purposefully included together with *non*-oneways, for the simple reason that motorways are very high on the hierarchy network. (This analyser is based on "travelling" through the road network to find deadends)

I had a quick look (just searching all uses of is_oneway), probably my comment about cleanup was unnecessary as there's no cases where it says `is_oneway OR highway='motorway'` or similar

ComradeRamen:
> Regarding *_deadend: here they're purposefully included together with _non_-oneways, for the simple reason that motorways are very high on the hierarchy network. (This analyser is based on "travelling" through the road network to find deadends)

Since this is supposed to add "valid input nodes" my understanding is that without removal it would count this as a valid connection. What am I missing?
image

Famlam:
> What am I missing?

That this is not `is_oneway` related, unlike the remainder of the PR. I understood your earlier comment "since they're now defined as oneways already" that you wanted to remove it because it's now a oneway.

The origin of this exception stems from here: https://github.com/osm-fr/osmose-backend/issues/231. That issue should have been solved by that we nowadays include all nodes from "border crossing ways" as safe points.

We could look into how it behaves without that exception (especially with island regions), but not in this PR. Probably it'll be ok to remove, assuming that every island with a motorway also has at least either a ferry terminal connected to the road network or a non-oneway road somewhere.

ComradeRamen:
Is there a way to see what caused the extra detection in the failed test?

Famlam (approved):
I checked the results of jersey, netherlands_gelderland and switserland_basel_stadt with and without this change, results seem ok to me :)

Famlam:
> Is there a way to see what caused the extra detection in the failed test?

It's dummy test data, see `\tests\osmosis_highway_deadend.osm`. Probably node 55 was detected, which makes sense with your modification (and isn't bad to detect, so don't worry too much about this). But lets discuss this in a separate PR (after this one is merged).
Likely you'll need to add `self.check_err(cl="3", elems=[("node", "65"), ("way", "1028")])` to the tests near line 540 and increase the count of 22 to 23 at the final line.

frodrigo:
Merged, but not deployed yet.

PR #2159, opened 2024-03-10
Merge_Public_Services_FR - small changes

nlehuby:
* add an address tag (suggested in https://forum.openstreetmap.fr/t/projet-prefectures-et-sous-prefectures/21856)
* select France Travail instead of Pole Emploi in datasource, and update tags for those (see https://forum.openstreetmap.fr/t/de-pole-emploi-a-france-travail/20397/20)


frodrigo:
Merged and deployed.

Issue #1949, opened 2023-07-08
"One way inaccessible or missing parking or parking entrance" for way originating outside extract


Famlam:
See https://osmose.openstreetmap.fr/nl/map/#zoom=18&lat=47.543533&lon=7.620853&item=1210&level=1&tags=&fixable=&issue_uuid=abb567ff-d458-da9c-9c87-311f3f34149a&class=3

Item 1210 / class 3
"One way inaccessible or missing parking or parking entrance"

The oneways originate from outside the extract. Hence, no "input node" is detected.

Solution probably similar to the solution for floating_islands (below) or maybe we can use that the node with `id=nid` is NULL in such cases. However, both only work in case the "travel direction" of the analyser follows the oneway.
https://github.com/osm-fr/osmose-backend/blob/869fd9d37ada04a4f1033071afe875ab6455a38a/analysers/analyser_osmosis_highway_floating_islands.py#L48-L60

Famlam:
Working on a possible fix together with #1670

Famlam:
Fixed in #2018

PR #2160, opened 2024-03-11
Fix missing motorway detections.

ComradeRamen:
https://github.com/osm-fr/osmose-backend/pull/2158 "added motorways to oneways"  should've detected the motorway input node 55 in `\tests\osmosis_highway_deadend.osm` but the inclusion of all motorway nodes as valid inputs means it wasn't. 

https://github.com/osm-fr/osmose-backend/issues/231 should also be resolved by https://github.com/osm-fr/osmose-backend/pull/2158 so the workaround is no longer needed.

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -191,11 +191,10 @@ FROM
 WHERE
   -- Non-oneway highways are valid input nodes for oneways
   (NOT is_oneway AND NOT is_roundabout) OR
-  -- Force motorways as valid input nodes
   -- Raceways are commonly isolated from the main network
   -- Escape (emergency stop) ways are supposed to be not used / dead-ended
   -- Footway can lead to anything (attraction entrances, ...) so exclude them even if oneway
-  highway IN ('motorway', 'raceway', 'escape', 'footway') OR
+  highway IN ('raceway', 'escape', 'footway') OR
   -- Construction roads are usually temporary, the connections are likely access=no or similar
   is_construction
 UNION ALL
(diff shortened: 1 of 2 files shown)


Famlam:
Thanks! I'll run it over the country of 231 tonight or tomorrow just to be sure, and will try to find an island with a motorway (and no internal extract borders) to do the same

Note that 2158 didn't resolve 231, this was probably the fix for #1949 which made a more generic fix

Famlam:
Also ones that weren't detected by class 2 yet?

ComradeRamen:
> Also ones that weren't detected by class 2 yet?

The deadends are detected by class 2.
There's also more wrong way motorways segments than most places.

ComradeRamen:
> Note that 2158 didn't resolve 231, this was probably the fix for #1949 which made a more generic fix

`        ways.tags->'highway' = 'motorway' OR -- Force motorway as input nodes` was first added in https://github.com/osm-fr/osmose-backend/commit/7bf8f1c926779824533aa37524b5e3c42a065116 as a fix to 231. 

Rewritten in https://github.com/osm-fr/osmose-backend/pull/2018

Famlam:
> > Note that 2158 didn't resolve 231, this was probably the fix for #1949 which made a more generic fix
> 
> ` ways.tags->'highway' = 'motorway' OR -- Force motorway as input nodes` was first added in [7bf8f1c](https://github.com/osm-fr/osmose-backend/commit/7bf8f1c926779824533aa37524b5e3c42a065116) as a fix to 231.
> 
> Rewritten in #2018

I should've been more clear. #2018 added - as a fix for 1949 - [another more generic approach](https://github.com/osm-fr/osmose-backend/pull/2018/files#diff-d17b00be7967eb39e666e3dcb7fec72f661aae4580e872545b000b81d3cbe1e5R152-R164) that made the original fix from 231 unnecessary, however I hadn't removed the original fix. Anyway, doesn't matter.

---

Can you please update line https://github.com/osm-fr/osmose-backend/blob/dev/tests/osmosis_highway_deadend.osm#L353 (the test case) to `<tag k='name' v='Class 2 and 3 bad' />`? Originally (test) way 1024 only served as a test for class 2 :)
Other than that it looks good to me.

ComradeRamen:
> Can you please update line https://github.com/osm-fr/osmose-backend/blob/dev/tests/osmosis_highway_deadend.osm#L353 (the test case) to `<tag k='name' v='Class 2 and 3 bad' />`? Originally (test) way 1024 only served as a test for class 2 :)
> Other than that it looks good to me.

Done

Famlam (approved):
Thanks! 
You can ignore the lint failure, it was a dependency that sometimes fails to install well

Fine with me :)

frodrigo:
Thank you all.

Issue #2163, opened 2024-03-12
Overlapping tunnel and bridge


ivanbranco:
I only recently found out a user that has been mapping nonexistent culverts and tunnels under bridges for months. I checked on Osmose and found out there are no checks for this. 

From bridge wiki:
> Bridge and tunnel options should not be used at the same time (a bridge tag for the upper way and a tunnel for the lower way for the same crossing).

I tried looking at old issues but I didn't find anything. Do you think this check could be practicable (I ask since the wiki says "should not" and not "must not")?

frodrigo:
It looks a reasonable check for me.

Famlam:
We might have to check that there's no extra way without tunnel nor bridge, e.g. I don't see an issue with hypothetical situations like this, where a culvert crosses a way while another way is going over that way as a bridge:
![image](https://github.com/osm-fr/osmose-backend/assets/1315520/66366fcc-a750-41b0-9dfd-ea5788bfc007)

PR #2164, opened 2024-03-12
Extend tunnel+bridge on same way check to all `bridge=*+tunnel=*` combinations

Famlam:
Instead of only warning about `tunnel=yes + bridge=yes` on the same way, warn about all `tunnel=* + bridge=*` on the same way (except when `*=no`).

Just a thought that crossed my mind in the other bridge/tunnel issue :)
I hope I'm not missing valid cases, e.g. I assume that for instance a bridge that crosses a building (above ground level) is a tunnel=building_passage without bridge tag. Example: https://mybestplace.com/en/article/chongqing-a-subway-passing-through-a-building

If in doubt, feel free to reject this PR


danieldegroot2:
Related #2163

Famlam:
Only a little bit, although it was my inspiration to revisit this check

2163 is about the two tags on different ways that are spatially intersecting.
This PR is about a single object having both tags simultaneously

PR #2165, opened 2024-03-12
Added ferries as valid endpoints

ComradeRamen:
I'll remove extra test if preferred.


Famlam:
Hi, can you give an example where this is an issue?
I find it hard to imagine a ferry route where there's loading, but no unloading (or the other way around).

ComradeRamen:
https://osmose.openstreetmap.fr/en/map/#zoom=18&lat=49.0056&lon=-123.129488&item=1210&level=1&issue_uuid=b9d8c63f-d560-a82f-58ef-89871d956cab

Famlam:
> https://osmose.openstreetmap.fr/en/map/#zoom=18&lat=49.0056&lon=-123.129488&item=1210&level=1&issue_uuid=b9d8c63f-d560-a82f-58ef-89871d956cab

Ah I see, thanks! Interesting case, hadn't thought of such constructions.

Isn't it better (performance wise) to whitelist amenity=ferry_terminal instead? That's only two nodes per (usual) route, rather than the route itself which can be hundreds of nodes. Saves running time for sql34 later on and nowadays ferry_terminal seems required for ferry routes. Just a thought to consider (yep, I realize class 1+2 lacks ferry_terminal too)

ComradeRamen:
Sorry about the commit spam but I think I got it .

ComradeRamen:
> yep, I realize class 1+2 lacks ferry_terminal too, perhaps your next PR ;)

Seems more complicated. What with the TEMP TABLE etc.

I'm not exactly an experienced coder.

Famlam (approved):

frodrigo:
Merged. Thank all.

Issue #2166, opened 2024-03-14
Item 1210 class 1+2: check if ferry check can be sped up


Famlam:
Just a thought that came up in #2165, filing it to look into in the future

Item 1210 class 1+2

We should probably investigate whether the expensive [`LEFT JOIN` on `route=ferry` ways](https://github.com/ComradeRamen/osmose-backend/blob/6764e24c7a6b863cac9592f62358912b28d1c05f/analysers/analyser_osmosis_highway_deadend.py#L101-L103) can be replaced by a simple check against `amenity=ferry_terminal` on the node tags [here](https://github.com/ComradeRamen/osmose-backend/blob/6764e24c7a6b863cac9592f62358912b28d1c05f/analysers/analyser_osmosis_highway_deadend.py#L90)

Based on the [wiki ](https://wiki.openstreetmap.org/wiki/Tag:route%3Dferry) all ferry routes should have a `amenity=ferry_terminal` at places where the ferry (un)loads. We should first check that this is also true on most ferry routes worldwide (otherwise the error `Unconnected highway` is very strange for such cases).

PR #2167, opened 2024-03-15
checking for ferry terminals instead of ferry routes.

ComradeRamen:
I started going through ferry route ends not tagged with these and will either fix them or relay back here.
Will give an update on progress tomorrow.

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -75,8 +75,7 @@ FROM
         nodes.id = way_ends.nid AND
         (NOT nodes.tags?'amenity' OR nodes.tags->'amenity' NOT IN ('bicycle_parking', 'ferry_terminal')) AND
         (NOT nodes.tags?'entrance' OR nodes.tags->'entrance' = 'no') AND
-        (NOT nodes.tags?'noexit' OR nodes.tags->'noexit' = 'no') AND
-        (NOT nodes.tags?'leisure' OR nodes.tags->'leisure' = 'slipway')
+        (NOT nodes.tags?'noexit' OR nodes.tags->'noexit' = 'no')
 WHERE
     way_ends.level < 3 OR way_ends.highway = 'cycleway'
 GROUP BY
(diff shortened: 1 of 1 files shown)


ComradeRamen:
https://github.com/osm-fr/osmose-backend/issues/2166

Famlam:
Hi,
Based on the description I suspect it's work in progress, so not reviewing the code yet, but would you expect slipways on bicycle paths, or secondary and above major roads? Slipways are for letting boats enter the water, which sounds wrong to me. (Imagine a high-priority motorway full of boat-carrying cars that all want to let their boat into the water there)

ComradeRamen:
Out of 4476 connecting nodes only 138 didn't have the `amenity=ferry_terminal` tag which I have already fixed. From what I can tell this PR should now not throw any new issues and should also resolve any "false positives" where the ferry_terminal is mapped but not the ferry route.

ComradeRamen:
Full [Script/s](https://pastebin.com/93EE3T40) used.

Famlam:
Thanks! I think you can remove lines 57 to 62 and line 86
```sql
SELECT
    wid,
    nid,
    geom,
    highway
FROM (
```
```sql
) AS t
```

They don't serve a purpose anymore now that the JOIN on ferry routes is gone

ComradeRamen:
Now I'm confused why node 158 isn't causing an extra error since its not tagged as a ferry terminal.

ComradeRamen:
> can remove lines 57 to 62 and line 86

Done.

Btw I interpreted "Highway from `motorway` to `tertiary` are important ways. They should
lead to somewhere and in particular to a network of minor roads." as meaning tertiary roads are included in the check even tho they're not.

Famlam:
> Btw I interpreted "Highway from `motorway` to `tertiary` are important ways. They should lead to somewhere and in particular to a network of minor roads." as meaning tertiary roads are included in the check even tho they're not.

The description is technically correct, because they *should* be connected, but it causes quite some issues in some less densely mapped areas for which I don't know better solutions. E.g. in sweden_gotland, for example https://www.openstreetmap.org/way/55408443 , on first glance I have the impression that the custom in this country is to map (nearly) all I-numbered roads as tertiary.
In addition, there's quite a couple of tertiaries that end in a `highway=construction` road which would all be flagged. It's very uncommon for an entire road to be mapped with a different classification only because there's a few months construction at a small part of the road.
Hence it's mentioned in the message that they should be connected, without actually giving a warning. 

> Done.

Thanks!

I'll give it a quick run on one or two countries later to be sure that there's no side effects (currently running something else...) but code-technically I don't have comments anymore.

ComradeRamen:
Why are construction roads excluded from endpoints unlike class 3.

Famlam (approved):

PR #2168, opened 2024-03-15
Reuse is_oneway instead of recalculating

Famlam:
This PR:
1. Re-uses the `is_oneway` property of table `highways`, rather than calculating it again
2. Adds a test case for `oneway=alternating`, in case the `is_oneway` check ever gets rewritten. (To ensure that an easy to miss bug like described in https://github.com/osm-fr/osmose-backend/pull/2158#issuecomment-1986826894 will never actually slip in)


frodrigo:
Thank you.

Issue #2138, opened 2024-01-21
Table multipolygons doesn't handle adjacent (closed) inner rings well: buildings on inner areas of farmland sharing sides (touching) w/other inner flagged as being on farmland.   


sekerob:
Could not find the previous topic open or closed regarding buildings on farming lands, but today saw a building of a group on an 'inner' farmland residential area being flagged ([Verrière, France](https://www.openstreetmap.org/changeset/146506037#map=18/45.74904/0.23103) ). Farmland is defined as a multipolygon relation. This area shared a side with a farmland 'inner' wood area (is considered touching) i.e. 2x inner where before one would map a ring around the 2 touching areas and make that one inner. Was told on the community forum this is no longer needed. For the exercise, drew that ring and made it inner to see if the building(s) get flagged again. Both the residential and wood are simple polygons.

Famlam:
I don't see any warnings related to buildings/farmlands in this region? (Note your changeset is only 5 hours old, but Osmose may take a day or so to update, so you may have seen an old marker)
https://osmose.openstreetmap.fr/nl/map/#zoom=17&lat=45.749629&lon=0.230487&item=xxxx&level=1%2C2%2C3&tags=&fixable=

Can you please give us:
1. The exact location
2. The item / class of the issue
3. Preferably the issue reported?
And as you mention a community discussion, please also link to the relevant (international) discussions. 

Note that the wiki of farmyard describes which buildings go on a farmyard rather than farmland. Farmland is really where the crops grow.
https://wiki.openstreetmap.org/wiki/Tag:landuse=farmyard

sekerob:
".... so you may have seen an old marker)", not really since the item appeared this morning in the 'personal' trouble list. Stared for longer at this and checked all possibles but the areas were truly 'inner' to the proper farmland, The only difference seen from all else was this side touching with that adjacent wood which was also mapped as a direct inner.

Your wiki link "farmhouse, dwellings, farmsteads, sheds, stables, barns, equipment sheds, feed bunkers, etc" is pretty much all and everything, all have building=yes but I'm curious which exceptions of buildings there are not in 'etc' which one can map on farmland. What are the exceptions to that rule of having to be on an inner area?

Anyway, the link was provided, so if nothing shows tomorrow I'll deem it a confirmation of a bug.

The building flagged was this elongated up against the road. As said all buildiings on that residential patch are =yes

[...]

PR #2169, opened 2024-03-16
Fix (if possible) MP with touching inners, drop otherwise

Famlam:
See #2138

Touching inners were dropped silently from the multipolygon geometry. Per the [wiki](https://wiki.openstreetmap.org/wiki/Relation:multipolygon#Touching_inner_rings), touching inners are valid in OSM (unlike OGC) and should just be considered as a single inner.

The reason for this 'silent dropping' of touching inners is that `ST_Union` doesn't join two touching closed ways unmodified, but rather converts it into three 'open' ways:
`ST_AsText(ST_Union(ARRAY[ST_GeomFromText('LINESTRING(1 0, 1 1, 0 0, 1 0)'), ST_GeomFromText('LINESTRING(1 0, 1 -1, 0 0, 1 0)')]))`
gives `MULTILINESTRING((1 0,1 1,0 0),(0 0,1 0),(1 0,1 -1,0 0))`
This cannot be resolved by `ST_BuildArea` anymore, so it drops it.

This can be resolved by first converting the rings to polygons already:
`ST_AsText(ST_Union(ARRAY[ST_MakePolygon(ST_GeomFromText('LINESTRING(1 0, 1 1, 0 0, 1 0)')), ST_MakePolygon(ST_GeomFromText('LINESTRING(1 0, 1 -1, 0 0, 1 0)'))]))`
gives `POLYGON((1 -1,0 0,1 1,1 0,1 -1))`

(`ST_BuildArea` accepts either closed ways and polygons for inners)

---

Results noticed during testing:
1. No new false positives in the changed entries - tested with osmosis_building_in_polygon, osmosis_polygon_intersects and osmosis_polygon_small on belgium_wallonia_german_community, france_centre_loiret, germany_saarland, japan_chubu, jersey, monaco, netherlands_gelderland, san_marino, slovenia, sweden_gotland, switzerland_basel_stadt, usa_iowa, venezuela)
2. Several false positives fixed, examples: [location 1](https://osmose.openstreetmap.fr/nl/map/#country=usa_iowa&item=1310&zoom=18&lat=41.526394&lon=-94.57104&level=3), [location 2](https://osmose.openstreetmap.fr/nl/map/#country=usa_iowa&item=1310&zoom=18&lat=41.753589&lon=-95.30946&level=3), [location 3](https://osmose.openstreetmap.fr/nl/map/#zoom=17&lat=36.717704&lon=138.48832&item=1070&level=1%2C2%2C3&class=12), [location 4](https://osmose.openstreetmap.fr/nl/map/#country=france_centre_loiret&item=1070&source=&class=15&username=&bbox=&zoom=18&lat=47.950698&lon=3.030552&level=2&tags=&fixable=) (and many more)
3. (Unexpected to me) better detection of invalid relations when `outer`s touch `inner`s, examples: [r14165680](https://www.openstreetmap.org/relation/14165680), [r14175064](https://www.openstreetmap.org/relation/14175064), so less false positives in general due to invalid relations being detected better. (My guess is that in those cases ST_BuildArea cannot generate an alternative valid structure anymore and thus must default to something that doesn't pass the ST_IsValid checks later on)

---

Unfortunately it's not perfect: in some rare cases, the inners are already dropped at `ST_BuildArea(mp.linestrings)`. Fixing it here is much more risky (we don't want to convert invalid MP to valid MP by accident) and this is pretty rare it seems, so instead the better solution seems to me to drop these cases (just like other MP that St_BuildArea on the linestrings cannot process). Cases found so far:
- https://www.openstreetmap.org/relation/10791091 
- https://www.openstreetmap.org/relation/9466424
- https://www.openstreetmap.org/relation/14164343
- https://www.openstreetmap.org/relation/14364124
- https://www.openstreetmap.org/relation/1288370
- https://www.openstreetmap.org/relation/5888956
- https://www.openstreetmap.org/relation/13559903
- https://www.openstreetmap.org/relation/13482010
- https://www.openstreetmap.org/relation/13384968
- https://www.openstreetmap.org/relation/12247611
- https://www.openstreetmap.org/relation/11073268
- https://www.openstreetmap.org/relation/11073166
- https://www.openstreetmap.org/relation/11069546
- https://www.openstreetmap.org/relation/3315160

Note, for the simpler cases of the above I noticed that it was always happening when the first=last nodes of the closed inner ways were in the overlapping area. Didn't check the complicated cases. Just listing them for future reference if needed.

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -110,7 +110,8 @@ WHERE
 ANALYZE {0}.highway_ends;
 """
 
-    # Multipolygons table is not complete. It does not contain multipolygons across extract border or invalid ones.
+    # Multipolygons table is not complete. It does not contain multipolygons across extract border,
+    # (some) invalid ones, and some where ST_BuildArea fails to build the full polygon with all inners
     sql_create_multipolygons = """
 DO $$
 DECLARE
@@ -150,7 +151,9 @@ BEGIN
             ST_NPoints(ST_Collect(ways.linestring)) < 100000
     ) LOOP
         BEGIN
-            IF ST_BuildArea(mp.linestrings) IS NOT NULL AND NOT ST_IsEmpty(ST_BuildArea(mp.linestrings)) THEN
+            IF ST_BuildArea(mp.linestrings) IS NOT NULL AND NOT ST_IsEmpty(ST_BuildArea(mp.linestrings))
+            -- Ensure no linestrings are dropped by ST_BuildArea, which would result in e.g. missing inners (see #2169)
+            AND ST_CoveredBy(ST_Points(mp.linestrings), ST_Points(ST_BuildArea(mp.linestrings))) THEN
                 WITH
                 unary AS (
                     SELECT
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -167,7 +167,7 @@ BEGIN
                         id, tags,
                         ST_BuildArea(ST_Collect(
                             ST_ExteriorRing(poly),
-                            (SELECT ST_Union(ST_InteriorRingN(poly, n)) FROM generate_series(1, ST_NumInteriorRings(poly)) AS t(n))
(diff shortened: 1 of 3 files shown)


frodrigo:
Thank you

PR #2176, opened 2024-03-21
adding tags to text field

ComradeRamen:
This is to allow sorting by subtitle on the website so I don't have to download and sort locally. Idk if its the most efficient way to do it.

### DIFF
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -450,12 +450,12 @@ Ensure that `service=drive-through` is the correct tag.''')),
             "class": 2,
             "data": [self.way_full, self.node_full, self.positionAsText],
             "text": {"en": "Unconnected highway: {}".format(res[3])}
-}
+        }
         self.callback43 = lambda res: {
             "class": 3,
             "data": [self.way_full, self.node_full, self.positionAsText],
             "text": {"en": "Oneway inaccessible: {}".format(res[3])}
-}
+        }
     def analyser_osmosis_common(self):
         boundary_relation = self.config.polygon_id # Either a number, None or (number, number, ...)
         if isinstance(boundary_relation, int):
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -449,7 +449,7 @@ Ensure that `service=drive-through` is the correct tag.''')),
         self.callback22 = lambda res: {
             "class": 2,
             "data": [self.way_full, self.node_full, self.positionAsText],
-            "text": {"en": "Unconnected highway: {}".format(res[3])}
+            "text": {"en": res[3]}
         }
         self.callback43 = lambda res: {
             "class": 3,
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -118,7 +118,7 @@ SELECT
   linestring,
   array_length(nodes, 1) AS length,
   generate_subscripts(nodes, 1) AS nid_index,
-  tags->'highway' AS highway_tag
+  highway
 FROM
   highways
 WHERE
@@ -351,7 +351,7 @@ SELECT DISTINCT ON(nid)
   wid,
   nid,
   ST_AsText(geom),
-  oneway_highway.highway_tag
+  oneway_highway.highway
 FROM (
   SELECT
     wid,
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -368,7 +368,8 @@ FROM (
     JOIN nodes ON
       nodes.id = results_recursive.nid
 ) AS t
-JOIN highways ON wid = id
+  JOIN highways ON
+    wid = id
 ORDER BY
   nid,
   wid
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -469,7 +469,7 @@ Ensure that `service=drive-through` is the correct tag.''')),
         self.run(sql40)
         self.run(sql41)
         self.run(sql42)
-        self.run(sql43, self.callback43)
+        self.run(sql43, lambda res: {"class":3, "data":[self.way_full, self.node_full, self.positionAsText], "text": T_("Oneway inaccessible: {0}", res[3])})
         self.run(sql50, lambda res: {"class":5, "data":[self.way_full, self.node, self.positionAsText]})
 
     def analyser_osmosis_full(self):
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -450,12 +450,12 @@ Ensure that `service=drive-through` is the correct tag.''')),
             "class": 2,
             "data": [self.way_full, self.node_full, self.positionAsText],
             "text": {"en": "Unconnected highway: {}".format(res[3])}
-            }
+}
         self.callback43 = lambda res: {
             "class": 3,
             "data": [self.way_full, self.node_full, self.positionAsText],
             "text": {"en": "Oneway inaccessible: {}".format(res[3])}
-            }
+}
     def analyser_osmosis_common(self):
         boundary_relation = self.config.polygon_id # Either a number, None or (number, number, ...)
         if isinstance(boundary_relation, int):
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -453,7 +453,7 @@ Ensure that `service=drive-through` is the correct tag.''')),
         self.callback43 = lambda res: {
             "class": 3,
             "data": [self.way_full, self.node_full, self.positionAsText],
-            "text": {"en": "Oneway inaccessible: {}".format(res[3])}
+            "text": T_("Oneway inaccessible: {0}", res[3])
         }
     def analyser_osmosis_common(self):
         boundary_relation = self.config.polygon_id # Either a number, None or (number, number, ...)
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -449,7 +449,7 @@ Ensure that `service=drive-through` is the correct tag.''')),
         self.callback22 = lambda res: {
             "class": 2,
             "data": [self.way_full, self.node_full, self.positionAsText],
-            "text": T_("Unconnected highway: {0}", res[3]}
+            "text": T_("Unconnected highway: {0}", res[3])
         }
         self.callback43 = lambda res: {
             "class": 3,
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -58,8 +58,7 @@ SELECT
     MIN(way_ends.id) AS wid,
     nodes.id AS nid,
     nodes.geom,
-    MIN(way_ends.highway) AS highway,
-    MIN(highways.tags->'highway') AS highway_tag
+    MIN(way_ends.highway) AS highway
 FROM
     {0}highway_ends AS way_ends
     JOIN highways ON
@@ -99,7 +98,7 @@ SELECT
     wid,
     nid,
     ST_AsText(geom),
-    highway_tag
+    highway
 FROM
     unconnected_highways
 WHERE
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -350,7 +350,7 @@ SELECT DISTINCT ON(nid)
   wid,
   nid,
   ST_AsText(geom),
-  highways.tags->'highway'
+  highways.highway
 FROM (
   SELECT
     wid,
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -450,12 +450,12 @@ Ensure that `service=drive-through` is the correct tag.''')),
         self.callback22 = lambda res: {
             "class": 2,
             "data": [self.way_full, self.node_full, self.positionAsText],
-            "text": {"en": "Highway tag: {}".format(res[3])}
+            "text": {"en": "Unconnected highway: {}".format(res[3])}
 }
         self.callback43 = lambda res: {
             "class": 3,
             "data": [self.way_full, self.node_full, self.positionAsText],
-            "text": {"en": "Highway tag: {}".format(res[3])}
+            "text": {"en": "Oneway inaccessible: {}".format(res[3])}
 }
     def analyser_osmosis_common(self):
         boundary_relation = self.config.polygon_id # Either a number, None or (number, number, ...)
--- a/analysers/analyser_osmosis_highway_deadend.py
+++ b/analysers/analyser_osmosis_highway_deadend.py
@@ -446,16 +446,8 @@ Ensure that `service=drive-through` is the correct tag.''')),
             resource = 'https://wiki.openstreetmap.org/wiki/Tag:service%3Ddrive-through')
 
         self.callback21 = lambda res: {"class": 1, "data": [self.way_full, self.node_full, self.positionAsText]}
-        self.callback22 = lambda res: {
-            "class": 2,
-            "data": [self.way_full, self.node_full, self.positionAsText],
-            "text": T_("Unconnected highway: {0}", res[3])
-        }
-        self.callback43 = lambda res: {
-            "class": 3,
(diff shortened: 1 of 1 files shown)


Famlam:
Hi,

A few comments:
- For class 2: column `highway_tag` = column `highway`, no need to duplicate
- For class 3: it is a (partially) recursive check, so adding the highway tag to the message will not add much in my opinion as in a loop it can pick any way
- The message "Highway tag: *" isn't really user friendly. For class 3 I wouldn't do it (showing the current title is much more informative in the list view than just the highway tag), but for class 2 I would just make it "Unconnected {the tag}". (In addition, if you're looking at item 1210 altogether without filtering on the class, you cannot distinguish classes 2 and 3 if both have the same text)

ComradeRamen:
> * For class 2: column `highway_tag` = column `highway`, no need to duplicate
> * For class 3: it is a (partially) recursive check, so adding the highway tag to the message will not add much in my opinion as in a loop it can pick any way

I'm sorry, I don't understand these comments. Could you clarify please?

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -58,7 +58,8 @@
     MIN(way_ends.id) AS wid,
     nodes.id AS nid,
     nodes.geom,
-    MIN(way_ends.highway) AS highway
+    MIN(way_ends.highway) AS highway,
+    MIN(highways.tags->'highway') AS highway_tag
```
Famlam, line 62:
```suggestion
    MIN(way_ends.highway) AS highway
```

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -97,7 +98,8 @@
 SELECT
     wid,
     nid,
-    ST_AsText(geom)
+    ST_AsText(geom),
+    highway_tag
```
Famlam, line 102:
```suggestion
    highway
```

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -116,7 +118,8 @@
   END AS nodes,
   linestring,
   array_length(nodes, 1) AS length,
-  generate_subscripts(nodes, 1) AS nid_index
+  generate_subscripts(nodes, 1) AS nid_index,
+  tags->'highway' AS highway_tag
```
Famlam, line 122:
```suggestion
  highway
```

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -348,7 +351,8 @@
 SELECT DISTINCT ON(nid)
   wid,
   nid,
-  ST_AsText(geom)
+  ST_AsText(geom),
+  oneway_highway.highway_tag
```
Famlam, line 355:
```suggestion
  oneway_highway.highway
```

Famlam (changes requested):
Untested, but hopefully these suggestions clarify what I mean with the first point
The column `highway` in table `highways` (and `highways_ends`) contains the tag value already, so adding another column `highway_tag`* is not needed.

*(which should have been highway_value by the way, it's just the value not the tag ;) )

The second point means that I *personally* don't see too much value in having the highway listed for class 3 as class 3 operates (partly) recursively over the road network, so the value can be different every time the database changes in an area. But that's just my opinion, I'll leave it to frodrigo and others to decide

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -442,8 +447,16 @@ def __init__(self, config, logger = None):
             resource = 'https://wiki.openstreetmap.org/wiki/Tag:service%3Ddrive-through')
 
         self.callback21 = lambda res: {"class": 1, "data": [self.way_full, self.node_full, self.positionAsText]}
-        self.callback22 = lambda res: {"class": 2, "data": [self.way_full, self.node_full, self.positionAsText]}
-
+        self.callback22 = lambda res: {
+            "class": 2,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": {"en": "Unconnected highway: {}".format(res[3])}
+}
```
Famlam, line 454: I think some indentation is missing here by the way? Same at line 459. There's zero spaces at the `}` while the callback function is indented by ~8 spaces

ComradeRamen: lint doesn't like it indented

ComradeRamen:
> The second point means that I _personally_ don't see too much value in having the highway listed for class 3 as class 3 operates (partly) recursively over the road network, so the value can be different every time the database changes in an area. But that's just my opinion, I'll leave it to frodrigo and others to decide

Still don't understand this fully but I want the highway value in class 3 so it can be sorted on the website since most of the results in the US are service roads and this should make it easier to prioritize issues.

analysers/analyser_osmosis_highway_deadend.py
Famlam: I think lint won't complain if it's at the same indent as the s from self.callback*

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -442,8 +446,16 @@ def __init__(self, config, logger = None):
             resource = 'https://wiki.openstreetmap.org/wiki/Tag:service%3Ddrive-through')
 
         self.callback21 = lambda res: {"class": 1, "data": [self.way_full, self.node_full, self.positionAsText]}
-        self.callback22 = lambda res: {"class": 2, "data": [self.way_full, self.node_full, self.positionAsText]}
-
+        self.callback22 = lambda res: {
+            "class": 2,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": {"en": "Unconnected highway: {}".format(res[3])}
```
frodrigo, line 452: Please, make the string translatable with `T_()`.

frodrigo, line 452: As "Unconnected highway" is already the label of the class. You just set the highway value.
```suggestion
            "text": {"en": res[3]}
```

Famlam (commented):
One thing to be checked on a larger extract: the id of oneway_highway isn't indexed, so it *may* be faster to use table highways instead to JOIN, which already has an index (without needing to build another index, which also costs time). Not sure if I'll have time to run this tomorrow..

By the way, that query 43 is now inconsistent with prefixing nid and wid. It's unnecessary to add `results_final.` prefixes, fine with me either way, but then also do the same in the SELECT at the very top with wid, nid and geom please. Or revert the prefixes, your call ;)

analysers/analyser_osmosis_highway_deadend.py
Famlam: Oh, this was actually my suggestion 🙈. It makes the list view more readable (because nothing is wrong with the highway value itself)

Famlam: @frodrigo just tagging to make sure you saw the comment as it was applied immediately.

Now there will be a strange text "service" or so, while there's nothing wrong with the highway=service tag

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -442,8 +446,16 @@ def __init__(self, config, logger = None):
...
-
+        self.callback22 = lambda res: {
+            "class": 2,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": {"en": res[3]}
+        }
+        self.callback43 = lambda res: {
+            "class": 3,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": {"en": "Oneway inaccessible: {}".format(res[3])}
```
Famlam, line 457: Should either become translatable (see the function `T_(string, parameter1, parameter2, ...)`, but lets await what frodrigo says about my comment on line 452

analysers/analyser_osmosis_highway_deadend.py
frodrigo: Too much notif, I do not read it.

As opposite, the text should use translation ( with `T_()`). And be translated by translator, removing the text was to avoid translation work. `text` is also know as "subtitle", and should come with the title.

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -366,6 +369,7 @@
     JOIN nodes ON
       nodes.id = results_recursive.nid
 ) AS t
+JOIN oneway_highway ON wid = oneway_highway.id
```
Famlam, line 372: > One thing to be checked on a larger extract: the id of oneway_highway isn't indexed, so it may be faster to use table highways instead to JOIN, which already has an index (without needing to build another index, which also costs time).

Current:
```
['Unique  (cost=9046767.12..9078631.33 rows=200 width=80) (actual time=431.099..431.337 rows=203 loops=1)']
['  ->  Sort  (cost=9046767.12..9062699.23 rows=6372842 width=80) (actual time=431.098..431.180 rows=1634 loops=1)']
['        Sort Key: results_straightforward.nid, results_straightforward.wid']
['        Sort Method: quicksort  Memory: 278kB']
['        ->  Merge Join  (cost=82826.05..8144485.78 rows=6372842 width=80) (actual time=370.268..430.595 rows=1634 loops=1)']
['              Merge Cond: (results_straightforward.wid = oneway_highway.id)']
['              ->  Sort  (cost=6857.00..6864.30 rows=2920 width=48) (actual time=1.434..1.455 rows=203 loops=1)']
['                    Sort Key: results_straightforward.wid']
['                    Sort Method: quicksort  Memory: 40kB']
['                    ->  Append  (cost=0.00..6688.93 rows=2920 width=48) (actual time=0.027..1.398 rows=203 loops=1)']
['                          ->  Seq Scan on results_straightforward  (cost=0.00..20.70 rows=1070 width=48) (actual time=0.027..0.048 rows=198 loops=1)']
['                          ->  Nested Loop  (cost=0.56..6635.13 rows=1850 width=48) (actual time=0.017..1.332 rows=5 loops=1)']
['                                ->  Seq Scan on results_recursive  (cost=0.00..28.50 rows=1850 width=16) (actual time=0.001..0.003 rows=5 loops=1)']
['                                ->  Index Scan using pk_nodes on nodes  (cost=0.56..3.57 rows=1 width=40) (actual time=0.264..0.264 rows=1 loops=5)']
['                                      Index Cond: (id = results_recursive.nid)']
['              ->  Sort  (cost=75969.05..77060.29 rows=436496 width=40) (actual time=368.341..389.638 rows=436151 loops=1)']
['                    Sort Key: oneway_highway.id']
['                    Sort Method: quicksort  Memory: 41850kB']
['                    ->  Seq Scan on oneway_highway  (cost=0.00..35078.96 rows=436496 width=40) (actual time=0.024..248.841 rows=436496 loops=1)']
['Planning Time: 0.252 ms']
['Execution Time: 435.084 ms']
```

Using `highways` instead of `oneway_highway` (here and on line 354):
```
['Unique  (cost=19632.60..19647.20 rows=200 width=58) (actual time=78.626..78.679 rows=203 loops=1)']
['  ->  Sort  (cost=19632.60..19639.90 rows=2920 width=58) (actual time=78.625..78.638 rows=203 loops=1)']
['        Sort Key: results_straightforward.nid, results_straightforward.wid']
['        Sort Method: quicksort  Memory: 53kB']
['        ->  Nested Loop  (cost=0.43..19464.53 rows=2920 width=58) (actual time=0.868..78.426 rows=203 loops=1)']
['              ->  Append  (cost=0.00..6688.93 rows=2920 width=48) (actual time=0.028..0.505 rows=203 loops=1)']
['                    ->  Seq Scan on results_straightforward  (cost=0.00..20.70 rows=1070 width=48) (actual time=0.027..0.108 rows=198 loops=1)']
['                    ->  Nested Loop  (cost=0.56..6635.13 rows=1850 width=48) (actual time=0.024..0.366 rows=5 loops=1)']
['                          ->  Seq Scan on results_recursive  (cost=0.00..28.50 rows=1850 width=16) (actual time=0.009..0.011 rows=5 loops=1)']
['                          ->  Index Scan using pk_nodes on nodes  (cost=0.56..3.57 rows=1 width=40) (actual time=0.069..0.069 rows=1 loops=5)']
['                                Index Cond: (id = results_recursive.nid)']
['              ->  Index Scan using idx_highways_id on highways  (cost=0.43..3.12 rows=1 width=18) (actual time=0.380..0.380 rows=1 loops=203)']
['                    Index Cond: (id = results_straightforward.wid)']
['Planning Time: 0.272 ms']
['Execution Time: 78.761 ms']
```

Seems to be more efficient to use table `highways`, using the existing index

analysers/analyser_osmosis_highway_deadend.py
Famlam: Hmm, ok. Still not my preference for the list view as it doesn't give information about the actual issue.

analysers/analyser_osmosis_highway_deadend.py
ComradeRamen: Like this https://github.com/osm-fr/osmose-backend/pull/2176/commits/813e0c9110ee1926173949151aca758f7001b687 ?

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -348,7 +349,8 @@
 SELECT DISTINCT ON(nid)
   wid,
   nid,
-  ST_AsText(geom)
+  ST_AsText(geom),
+  highways.tags->'highway'
```
Famlam, line 353:
```suggestion
  highways.highway
```

analysers/analyser_osmosis_highway_deadend.py
Famlam: Just use `highways.highway` instead of `highways.tags->'highway'` :)
See my suggestion

analysers/analyser_osmosis_highway_deadend.py
ComradeRamen: Not sure how to implement this.

Famlam: My own preference would be:
`"text": T_("Oneway inaccessible: {0}", res[3])`
Just to avoid confusion that something would be wrong with the highway tag.

But it might be that @frodrigo also prefers to use the same here:
`"text": {"en": res[3]}`

@frodrigo you choose? You didn't comment on this one so I'm not sure if you preferred it for both or just for the first one?

Famlam (approved):

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -366,6 +368,7 @@
     JOIN nodes ON
       nodes.id = results_recursive.nid
 ) AS t
+JOIN highways ON wid = id
```
frodrigo, line 371: Code format
```suggestion
  JOIN highways ON
    wid = id
```

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -442,8 +445,16 @@ def __init__(self, config, logger = None):
...
-
+        self.callback22 = lambda res: {
+            "class": 2,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": {"en": res[3]}
+        }
+        self.callback43 = lambda res: {
+            "class": 3,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": T_("Oneway inaccessible: {0}", res[3])
```
frodrigo, line 456: Raw data for both, or text with translation for both.

ComradeRamen: >or text with translation for both.

done

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -442,8 +446,16 @@ def __init__(self, config, logger = None):
...
+        self.callback22 = lambda res: {
+            "class": 2,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": T_("Unconnected highway: {0}", res[3])
+        }
+        self.callback43 = lambda res: {
+            "class": 3,
+            "data": [self.way_full, self.node_full, self.positionAsText],
+            "text": T_("Oneway inaccessible: {0}", res[3])
+        }
```
frodrigo, line 458: self.callback22 = lambda res: {"class": 2, "data": [self.way_full, self.node_full, self.positionAsText]}

```suggestion
        self.callback22 = lambda res: {"class": 2, "data": [self.way_full, self.node_full, self.positionAsText], "text": T_("Unconnected highway: {0}", res[3])}

```

analysers/analyser_osmosis_highway_deadend.py
```diff
@@ -465,7 +477,7 @@ def analyser_osmosis_common(self):
         self.run(sql40)
         self.run(sql41)
         self.run(sql42)
-        self.run(sql43, lambda res: {"class":3, "data":[self.way_full, self.node_full, self.positionAsText]})
+        self.run(sql43, self.callback43)
```
frodrigo, line 480:
```suggestion
        self.run(sql43, lambda res: {"class":3, "data":[self.way_full, self.node_full, self.positionAsText], "text": T_("Oneway inaccessible: {0}", res[3])})
```

frodrigo:
Merged, thank you all.

PR #2184, opened 2024-03-30
Detect more invalid multipolygons

Famlam:
Adds warnings for invalid multipolygons that can be build by table `multipolygons` (in an invalid way).
These are primarily cases where the inner ways cross or overlap the outer ways.

It does not overlap with the existing analyser (sql21), the two sql queries provide complementary results.

Also includes some minor clean-ups (second commit)


frodrigo:
Thank you.

PR #2188, opened 2024-03-31
Warn if ST_AsText was forgotten

Famlam:
It has happened quite often to me that I forgot to add `ST_AsText` (or similar) to the position indicator, resulting in no output for matching elements in the XML files and costing quite some debugging time.

Hopefully adding this little warning is fine?

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -651,7 +651,11 @@ WHERE
         if res is None:
             self.logger.warn("NULL location provided")
             return []
-        for loc in self.get_points(res):
+        points = self.get_points(res)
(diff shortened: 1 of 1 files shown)


frodrigo:
Why not.

But infact the best move here it to remove the usage of ST_AsText, by reading the binary geometry.

We can merge this anyway.

Famlam:
> But infact the best move here it to remove the usage of ST_AsText, by reading the binary geometry.

I agree, but this would be quite a big change (affecting all analysers), so maybe better for a future PR.
Should be doable though: https://stackoverflow.com/questions/74398628/wkb-string-to-wkt-geojson-in-python

PR #2194, opened 2024-04-09
Fix French graceTHD analyser for power poles and add Brittany and Jura

flacombe:
Two fixes for common French GraceTHD analyzer:
* Add `height` field out of GraceTHD `pt_a_haut` when defined and > 6 meters
* Default GraceTHD's `pt_gest` with `pt_prop` as public authority sometimes confuses owners and operators (and GraceTHD summarizes them with common identifiers).

Two supplementary regions were recently added on data.gouv:
* Brittany (24 000 poles)
* Jura (20 000 poles)


Famlam:
Hi, I didn't check it in full detail, but shouldn't the new files be added to the config?

flacombe:
Oh that's right... I did it again :p

PR #2198, opened 2024-04-13
Enable multipolygon intersection checks

Famlam:
See the Matrix chat earlier this week.

`natural=water` *ways* intersecting highways are dealt with in `osmosis_highway_vs_building`, which doesn't support multipolygons. It is a special case there. On the other hand, `osmosis_polygon_intersects` deals with all polygons *except* `natural=water`, for highways. (It already deals with railways etc intersecting `natural=water`). However, `osmosis_polygon_intersects` is restricted to important highways (to avoid spamming) and takes a small buffer where intersections are permitted.

Therefore, porting the check directly to osmosis_polygon_intersects without losing existing results would be very difficult (and expensive) as natural=water vs. highways would become an exception nearly everywhere in the code: all other polygons should remain checked vs. important highways (& railways etc) only, and with the 2m buffer.

Rather than adding multipolygon support to `osmosis_highway_vs_building` (which is already a 'heavy' analyser), make the special case more specific so that multipolygon support is added for the cases that would also be found by osmosis_polygon_intersects. That way the results remain unchanged AND multipolygon support is added.

For the user it probably won't be noticed: both analysers report to item 1070

### DIFF
--- a/analysers/analyser_osmosis_polygon_intersects.py
+++ b/analysers/analyser_osmosis_polygon_intersects.py
@@ -77,7 +77,7 @@ sql11 = """
 -- Collect all landuse=* and natural=*, closed ways and multipolygons
 CREATE TEMP TABLE landusage AS
 SELECT
-  'W' || id AS id,
+  'W' || id AS tid,
   ST_MakePolygon(linestring) AS poly_full,
   ST_Subdivide(ST_Buffer(ST_Transform(ST_MakePolygon(linestring), {proj}), -2.0), 1000) AS poly_proj_buffer_fragment,
   CASE
@@ -93,7 +93,7 @@ WHERE
   (tags?'natural' OR tags?'landuse')
 UNION ALL
 SELECT
-  'R' || id AS id,
+  'R' || id AS tid,
   poly AS poly_full,
   ST_Subdivide(ST_Buffer(poly_proj, -2.0), 1000) AS poly_proj_buffer_fragment,
   CASE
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -167,14 +167,14 @@
     landusekey = 'natural' AND
     landusevalue IN ('bay', 'cliff', 'scrub', 'shrubbery', 'sinkhole', 'tree_group', 'wetland', 'wood') OR
     (
-      -- Special case as highway=* vs. natural=water is already included in item 1070 class 4/5
+      -- Special case as highway=* vs. natural=water _ways_ are already included in item 1070 class 4
       landusevalue = 'water' AND
-      mobility_way_type != 'highway'
+      (mobility_way_type != 'highway' OR SUBSTRING(landusage_tid,1,1) = 'R')
```
frodrigo, line 172:
```suggestion
      (mobility_way_type != 'highway' OR substring(landusage_tid,1,1) = 'R')
```

analysers/analyser_osmosis_polygon_intersects.py
```diff
@@ -167,14 +167,14 @@
     landusekey = 'natural' AND
     landusevalue IN ('bay', 'cliff', 'scrub', 'shrubbery', 'sinkhole', 'tree_group', 'wetland', 'wood') OR
     (
-      -- Special case as highway=* vs. natural=water is already included in item 1070 class 4/5
+      -- Special case as highway=* vs. natural=water _ways_ are already included in item 1070 class 4
       landusevalue = 'water' AND
-      mobility_way_type != 'highway'
+      (mobility_way_type != 'highway' OR substring(landusage_tid,1,1) = 'R')
```
frodrigo, line 172:
```suggestion
      (mobility_way_type != 'highway' OR substring(landusage_tid, 1, 1) = 'R')
```

frodrigo:
Thank you for the analysis.

PR #2207, opened 2024-04-27
Fix typo

ComradeRamen:


frodrigo:
@jocelynj There is a typo in an extract name.

frodrigo:
@ComradeRamen Bu the issue is still here. Can you tell us again on witch name the problem is ?

The issue is that not only to be fixed on Osmose side, but also on download.openstreetmap.fr

ComradeRamen:
Didn't mean to close. I'll find it in the morning

ComradeRamen:
https://github.com/osm-fr/osmose-backend/pull/2245

PR #2208, opened 2024-04-27
Remove workaround for kOpeningHours bug

### LINKED ISSUE #2136: Update kOpeningHours when released, remove workaround
Now that [this PR at kOpeningHours](https://invent.kde.org/libraries/kopeninghours/-/merge_requests/103) was merged, we can remove the [workaround](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/ConditionalRestrictions.py#L42) as soon as a new kOpeningHours release has been made.
Filing for future reference :)

Edit: not included yet in [v24.02.1](https://invent.kde.org/libraries/kopeninghours/-/commits/v24.02.1)

Famlam:
The fix for the workaround has been included in v24.04.80
Fixes #2136

Although the workaround caught most issues, it still failed if there was another issue simultaneously.
For example, the fix suggestion for:
`2024 Mar 02-2024 Apr 08 Vr-Zo` (note: using Vr and Zo instead of Fr and Su) suggested:
`2024 Mar 02-Apr 08 Fr-Su` instead of
`2024 Mar 02-2024 Apr 08 Fr-Su`

This was because the workaround essentially just converted `2024 Mar 02-2024 Apr 08 Vr-Zo` to `2024 Mar 02-Apr 08 Vr-Zo` before feeding it to pyKOpeningHours, so that pyKOpeningHours wouldn't detect the (false positive) issue with the years. However, if there was a second issue simultaneously, this also meant that the fix suggestion included the bad year conversion.

### DIFF
--- a/plugins/ConditionalRestrictions.py
+++ b/plugins/ConditionalRestrictions.py
@@ -39,8 +39,6 @@ class ConditionalRestrictions(Plugin):
     self.ReWeekdayMonthOpeningH = re.compile(r'\b[A-Z][a-z]+') # i.e. Mar or Mo
     self.ReMonthDayOpeningH = re.compile(r'\w\w\w[\s-]\d') # i.e. sep 1
     self.ReTimeOpeningH = re.compile(r'\d\D[\d-]|sun[sr][ei][ts]') # i.e. 5:30 or 5h30 or 5h-8h
-    # Workaround https://bugs.kde.org/show_bug.cgi?id=452236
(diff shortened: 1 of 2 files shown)


Famlam:
@frodrigo, did this deploy well?
I'm surprised to find issues like https://osmose.openstreetmap.fr/nl/issue/7f4b0d18-1f5b-5a54-c987-1feade40d7d1

However, if I test them locally they don't appear (and they shouldn't, because the update of kOpeningHours should have resolved this). It seems the deployment of the python file worked well, but the update of kOpeningHours didn't.

This is likely also what you see in the graph, which suddenly jumped https://osmose.openstreetmap.fr/nl/issues/graph.png?item=3350&class=33504&start_date=2024 while a steady line was expected.

Issue #2213, opened 2024-05-06
Minor clarification for `missing oneway`


opk12:
[This example](https://osmose.openstreetmap.fr/en/issue/e9e0aaaa-cf24-029f-2aff-a605b863c71e) of item 2030's entry `2. Missing oneway` is probably a false positive, from a quick look at the aerial imagery. What about mentioning `oneway=no` in the warning? The wiki [allows](https://wiki.openstreetmap.org/wiki/Key:oneway#oneway_=_no) the value `no` explicitly.

In particular, the warning says `After verifying that it is an access road to the roundabout and they
were well oriented, set the tag oneway=yes on the two segments.` ([source code](https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_roundabout_level.py#L278)). It could say something like `If the road is two-way, set oneway=no or mark as a false positive`.

Famlam:
Solved :)

PR #2216, opened 2024-05-09
Exclude the node shared with the roundabout

Famlam:
See #2213 - fixes incorrectly reported "should be oneway" cases if ways joined at the roundabout.

The current behavior included the node on the roundabout itself for ways where the final node was the roundabout way.
Consequently, completely unrelated ways that only intersected at the roundabout were detected, thus giving a false warning about a required oneway tag.

It was already correct for ways in the reverse direction (e.g. line 162 in the code)

Asides from the way in the issue, it also fixes other very similar cases, like
https://www.openstreetmap.org/way/384256584
https://www.openstreetmap.org/way/141816760
https://www.openstreetmap.org/way/88823059

(And it adds cases where the intersection occurs at the now newly included node)

---

@frodrigo I find it odd that also the node is given as output geometry to the user (e.g. node 1630563193 in https://osmose.openstreetmap.fr/en/issue/e9e0aaaa-cf24-029f-2aff-a605b863c71e). I think it should only report the way, not? The node seems completely random.

---

As a future improvement, we should probably check for intersections in a certain way length, rather than at the first 3 nodes after the roundabout. Using the first 3 nodes makes it highly dependent on the level of mapping accuracy. If very accurate, 3 nodes may miss the intersection; if very poorly mapped, it may find cases where the ways are completely unrelated but simply join by coincident at some point.

### DIFF
--- a/analysers/analyser_osmosis_roundabout_level.py
+++ b/analysers/analyser_osmosis_roundabout_level.py
@@ -160,7 +160,7 @@ SELECT
     ways.id AS a_id,
     CASE
         WHEN ways.nodes[1] = ANY (roundabout.nodes) THEN ARRAY[ways.nodes[2], ways.nodes[3], ways.nodes[4]]
-        WHEN ways.nodes[array_length(ways.nodes,1)] = ANY (roundabout.nodes) THEN ARRAY[ways.nodes[array_length(ways.nodes,1)], ways.nodes[array_length(ways.nodes,1)-1], ways.nodes[array_length(ways.nodes,1)-2]]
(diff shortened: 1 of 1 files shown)


frodrigo:
- Ideally the now would the one the of intersection with the other insertion way.
- Yes a check of distance from roundabout will be an improvement. Maybe check intersection distance from roundabout or way length.

Famlam:
Filed #2217 for this :)

> Ideally the now would the one the of intersection with the other insertion way.

For the marker position I can see this, but how is the node itself relevant for the user? I would actually propose to remove the node from the output as it's only the way that needs to be adjusted.

Issue #2217, opened 2024-05-09
Improve osmosis_roundabout_level oneway check: use a certain way length, not nodes


Famlam:
See https://github.com/osm-fr/osmose-backend/pull/2216 (Original message starting from the horizontal line, plus first comment)

Item 2030 class 2 https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_osmosis_roundabout_level.py

The current approach looks at the first 3 nodes after the roundabout, but this is highly dependent on mapping quality. It would for instance not find very detailed mapping approaches like in the first picture (intersection only occurs after 7 nodes). However, in cases of low mapping quality (second picture) it incorrectly finds ways that aren't related to eachother.
![image](https://github.com/osm-fr/osmose-backend/assets/1315520/439c9101-0426-4faf-82fa-674ea63cbb52) ![image](https://github.com/osm-fr/osmose-backend/assets/1315520/062e4d86-7277-46b0-b74b-d0e5a4920fb3)

Possible solutions:
1. Through-space distance from roundabout to intersection < treshold*
2. Restricting the way length** < treshold
3. Absolute way distance from roundabout to intersection < treshold

*This is more restricting on the distance we can check on. If hypothetically 100m highway length is the optimum for option 3 to avoid the false positives from image 2, then option 1 would be restricted to ~70m (assuming square street blocks), but this would favor ways with bends.
**Although 2 is the easiest, it's prone to this case that I've seen quite a few times, where the road isn't necessarily split at the intersection. Example (using different road types to make it more clear how the 3 roads go) 
![image](https://github.com/osm-fr/osmose-backend/assets/1315520/ae66a815-6f43-4d09-8ad7-2d0e72ff31c9)

PR #2218, opened 2024-05-11
Use (true) length rather than a fixed number of nodes as limit for roundabout connections

Famlam:
#2217 

Changes:
- If the road network is mapped in low detail, reduce the chance of suggesting to add oneway where not applicable. Example:   
  - https://www.openstreetmap.org/way/848800395 + https://www.openstreetmap.org/way/862468870
  - https://www.openstreetmap.org/way/177855997 + https://www.openstreetmap.org/way/177855984
- If the road network is mapped in high detail, find connections without oneway if they have more than 4 nodes. Examples:
  - https://www.openstreetmap.org/way/1236336280 / https://www.openstreetmap.org/way/185168321 (this one would need to be split)
  - https://www.openstreetmap.org/way/76579954 / https://www.openstreetmap.org/way/76579970
- Place the node on the intersection of the ways

The cutoff of 100m was chosen because it contains the vast majority of all currently found connecting highways (>95%) - and many more which were previously not found due to the node count limit - yet is short enough to make it unlikely that roads connect by coincidence (see first point).
Note that no cut-off will be perfect, unfortunately. With the current length, for example, [way 420552742](https://www.openstreetmap.org/way/420552742) is very likely a false positive, while [way 402048470](https://www.openstreetmap.org/way/402048470) is probably a false negative.

**Note regarding commit "Simplify"**
I _think_ these statements are equal (linesubstring takes fractions and I suspect that this won't change between the default projection and the extract projection), but I'm not 100% certain about this. Please let me know if my assumption is correct or not.

### DIFF
--- a/analysers/analyser_osmosis_roundabout_level.py
+++ b/analysers/analyser_osmosis_roundabout_level.py
@@ -159,9 +159,12 @@ SELECT
     roundabout.id AS ra_id,
     ways.id AS a_id,
     CASE
-        WHEN ways.nodes[1] = ANY (roundabout.nodes) THEN ARRAY[ways.nodes[2], ways.nodes[3], ways.nodes[4]]
-        WHEN ways.nodes[array_length(ways.nodes,1)] = ANY (roundabout.nodes) THEN ARRAY[ways.nodes[array_length(ways.nodes,1)-1], ways.nodes[array_length(ways.nodes,1)-2], ways.nodes[array_length(ways.nodes,1)-3]]
-    END AS n_ids,
+        -- Get the (up to) 100m of linestring directly after the roundabout, excluding the connection itself (0.1m)
+        WHEN ways.nodes[1] = ANY(roundabout.nodes) THEN
+            ST_Transform(ST_LineSubString(ways.linestring_proj, 0.1/ST_Length(ways.linestring_proj), LEAST(1, 100/ST_Length(ways.linestring_proj))), 4326)
+        WHEN ways.nodes[array_length(ways.nodes,1)] = ANY(roundabout.nodes) THEN
+            ST_Transform(ST_LineSubString(ways.linestring_proj, 1-LEAST(1, 100/ST_Length(ways.linestring_proj)), 1-0.1/ST_Length(ways.linestring_proj)), 4326)
+    END AS connection_sublinestring,
     ways.is_oneway
 FROM
     roundabout
@@ -171,7 +174,8 @@ FROM
         roundabout.id != ways.id AND
         NOT ways.is_construction
 WHERE
-    ways.highway IN ('primary', 'secondary', 'tertiary', 'unclassified', 'residential', 'road')
+    ways.highway IN ('primary', 'secondary', 'tertiary', 'unclassified', 'residential', 'road') AND
+    ST_Length(ways.linestring_proj) > 0.1
 """
 
 sql21 = """
@@ -181,19 +185,15 @@ CREATE INDEX roundabout_access_idx ON roundabout_access(ra_id)
 sql22 = """
 SELECT
     ra1.a_id,
-    COALESCE(ra1.n_ids[2], ra1.n_ids[1]),
-    COALESCE(ra1.n_ids[2], ra1.n_ids[1])
+    ST_AsText(ST_Intersection(ra1.connection_sublinestring, ra2.connection_sublinestring))
 FROM
     roundabout_access AS ra1
     JOIN roundabout_access AS ra2 ON
         ra1.ra_id = ra2.ra_id AND
         ra1.a_id != ra2.a_id
 WHERE
-    ra1.n_ids && ra2.n_ids AND
+    ST_Intersects(ra1.connection_sublinestring, ra2.connection_sublinestring) AND
     NOT ra1.is_oneway
-GROUP BY
-    ra1.a_id,
-    COALESCE(ra1.n_ids[2], ra1.n_ids[1])
 """
 
 sql30 = """
@@ -258,6 +258,9 @@ class Analyser_Osmosis_Roundabout_Level(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
+        if not "proj" in self.config.options:
+            return
+
         self.classs[1] = self.def_class(item = 3010, level = 2, tags = ['highway', 'roundabout', 'fix:chair'],
             title = T_('Wrong highway on roundabout'),
             detail = T_(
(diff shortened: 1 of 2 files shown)


Famlam:
> > Note regarding commit "Simplify"
> 
> As length as short comparing to earth size, that right.

Ah ok, so there is thus a difference between the statements (at least for non-horizontal linestrings), even though its small?
I was not sure whether the relative proportions of a linestring would change under different projections. do you have a preference to use or not use the changes from commit "Simplify"?

---

> There is no initially, but what do you think about adding and index ?

I'm not sure if I understand correctly, but both sql20 and sql22 use an index.

For `italy_emilia_romagna`:

_sql20_
```
['Nested Loop  (cost=1745.54..138432.39 rows=18051 width=49) (actual time=17.137..2871.915 rows=14305 loops=1)']
['  ->  Bitmap Heap Scan on highways ways  (cost=1745.39..69473.71 rows=66984 width=489) (actual time=16.991..207.735 rows=199507 loops=1)']
["        Recheck Cond: (highway = ANY ('{primary,secondary,tertiary,unclassified,residential,road}'::text[]))"]
["        Filter: ((NOT is_construction) AND (st_length(linestring_proj) > '0.1'::double precision))"]
['        Heap Blocks: exact=32827']
['        ->  Bitmap Index Scan on idx_highways_highway  (cost=0.00..1728.64 rows=201347 width=0) (actual time=12.097..12.097 rows=199507 loops=1)']
["              Index Cond: (highway = ANY ('{primary,secondary,tertiary,unclassified,residential,road}'::text[]))"]
['  ->  Index Scan using roundabout_linestring_idx on roundabout  (cost=0.14..0.20 rows=1 width=72) (actual time=0.013..0.013 rows=0 loops=199507)']
['        Index Cond: (linestring && ways.linestring)']
['        Filter: ((id <> ways.id) AND (((ways.nodes)[1] = ANY (nodes)) OR ((ways.nodes)[array_length(ways.nodes, 1)] = ANY (nodes))))']
['        Rows Removed by Filter: 0']
['Planning Time: 0.505 ms']
['Execution Time: 2873.245 ms']
```

_sql22_ 
```
['Sort  (cost=12900201.40..12900202.68 rows=509 width=40) (actual time=19.761..19.763 rows=28 loops=1)']
['  Sort Key: ra1.a_id']
['  Sort Method: quicksort  Memory: 27kB']
['  ->  Nested Loop  (cost=0.29..12900178.52 rows=509 width=40) (actual time=0.446..19.727 rows=28 loops=1)']
['        ->  Seq Scan on roundabout_access ra1  (cost=0.00..396.05 rows=7152 width=48) (actual time=0.008..1.895 rows=3148 loops=1)']
['              Filter: (NOT is_oneway)']
['              Rows Removed by Filter: 11157']
['        ->  Index Scan using roundabout_access_idx on roundabout_access ra2  (cost=0.29..1801.78 rows=1 width=48) (actual time=0.005..0.005 rows=0 loops=3148)']
['              Index Cond: (ra_id = ra1.ra_id)']
['              Filter: ((ra1.a_id <> a_id) AND st_intersects(ra1.connection_sublinestring, connection_sublinestring))']
['              Rows Removed by Filter: 4']
['Planning Time: 0.381 ms']
['Execution Time: 19.816 ms']
```


However, maybe you'd prefer sql22 to use an index on connection_sublinestring? This would give:
```
['Sort  (cost=194390.92..194392.20 rows=509 width=40) (actual time=115.274..115.277 rows=28 loops=1)']
['  Sort Key: ra1.a_id']
['  Sort Method: quicksort  Memory: 27kB']
['  ->  Nested Loop  (cost=0.15..194368.04 rows=509 width=40) (actual time=1.352..115.227 rows=28 loops=1)']
['        ->  Seq Scan on roundabout_access ra1  (cost=0.00..396.05 rows=7152 width=48) (actual time=0.011..3.892 rows=3148 loops=1)']
['              Filter: (NOT is_oneway)']
['              Rows Removed by Filter: 11157']
['        ->  Index Scan using roundabout_access_sublinestring_idx on roundabout_access ra2  (cost=0.15..25.24 rows=1 width=48) (actual time=0.033..0.033 rows=0 loops=3148)']
['              Index Cond: (connection_sublinestring && ra1.connection_sublinestring)']
['              Filter: ((ra1.a_id <> a_id) AND (ra1.ra_id = ra_id) AND st_intersects(ra1.connection_sublinestring, connection_sublinestring))']
['              Rows Removed by Filter: 1']
['Planning Time: 1.405 ms']
['Execution Time: 115.367 ms']
```

(For e.g. Iran it would use both indices, but I'm not convinced that this is worth the extra time to build two indices)
```
['Sort  (cost=1521409.48..1521445.07 rows=14235 width=40) (actual time=926.438..926.453 rows=270 loops=1)']
['  Sort Key: ra1.a_id']
['  Sort Method: quicksort  Memory: 49kB']
['  ->  Nested Loop  (cost=3.74..1520427.47 rows=14235 width=40) (actual time=4.474..926.103 rows=270 loops=1)']
['        ->  Seq Scan on roundabout_access ra1  (cost=0.00..2028.49 rows=37824 width=48) (actual time=0.018..22.484 rows=22507 loops=1)']
['              Filter: (NOT is_oneway)']
['              Rows Removed by Filter: 53142']
['        ->  Bitmap Heap Scan on roundabout_access ra2  (cost=3.74..30.25 rows=1 width=48) (actual time=0.039..0.039 rows=0 loops=22507)']
['              Recheck Cond: (ra_id = ra1.ra_id)']
['              Filter: ((ra1.a_id <> a_id) AND st_intersects(ra1.connection_sublinestring, connection_sublinestring))']
['              Rows Removed by Filter: 1']
['              Heap Blocks: exact=25792']
['              ->  BitmapAnd  (cost=3.74..3.74 rows=1 width=0) (actual time=0.031..0.031 rows=0 loops=22507)']
['                    ->  Bitmap Index Scan on roundabout_access_sublinestring_idx  (cost=0.00..0.36 rows=8 width=0) (actual time=0.028..0.028 rows=1 loops=22507)']
['                          Index Cond: (connection_sublinestring && ra1.connection_sublinestring)']
['                    ->  Bitmap Index Scan on roundabout_access_idx  (cost=0.00..3.13 rows=378 width=0) (actual time=0.003..0.003 rows=5 loops=22507)']
['                          Index Cond: (ra_id = ra1.ra_id)']
['Planning Time: 0.464 ms']
['Execution Time: 926.529 ms']
```

frodrigo:
> Ah ok, so there is thus a difference between the statements (at least for non-horizontal linestrings), even though its small?
I was not sure whether the relative proportions of a linestring would change under different projections. do you have a preference to use or not use the changes from commit "Simplify"?

I was wrong, as we want meters and not proportional distance. This simplification cannot be done.

>  index on connection_sublinestring

Yes , this one.

> but I'm not convinced that this is worth the extra time to build

As you want. That right, the run is already quick.

Unless when query do only one scan, it worth to add an index. On small temp table like that, it is all in memory.

Famlam:
Ok, I've reverted "Simplify".

Regarding the indices, you can choose :) . The (in that case to be added) index on the sublinestrings gives the lowest cost score, but the (already existing) index on the id of the roundabout that the ways are connected to seems to be the fastest for countries with many roundabouts (tested on iran, italy_emilia_romagna, and netherlands_gelderland).

frodrigo:
Ok, thank you. Let merge it as is, as there is alreadyan index used.

Issue #2219, opened 2024-05-11
Roundabout crossing road only detected if it has `access=*` (`*`!=`no`, `private` or `psv`)


Famlam:
sql40 of [analyser_osmosis_roundabout_level](https://github.com/osm-fr/osmose-backend/blob/dev/analysers/analyser_osmosis_roundabout_level.py#L237) seems to be missing a `NOT ways.tags?'access'` check. Presently it only matches ways that *do* have an explicit `access` key (other than those with excluded values). This seems to have been introduced [9 years ago](https://osmose.openstreetmap.fr/nl/issues/graph.png?item=3010&class=4) [in this commit](https://github.com/osm-fr/osmose-backend/commit/27a170ec70ca8cea9732c82e31fb6fc88a02eb9f).

Todo:
1. Change `ways.tags->'access' NOT IN ('no', 'psv', 'private')` to `(NOT ways.tags?'access' OR ways.tags->'access' NOT IN ('no', 'psv', 'private'))`
2. Add `self.check_err(cl="4", elems=[("way", "1000"), ("way", "1010")])` to the tests, and increase the expected error count by 1. (Requires 2218 to be merged first).
3. Probably the relatively new `highway=busway` (and `bus_guideway`) should be excluded, similar to how access=psv is excluded. And probably `access=emergency` too; maybe other cases that didn't exist yet in 2015?

(I didn't want to mix it with PR 2218 as it's unrelated)

PR #2221, opened 2024-05-14
Enable check on ways without `access` tag

Famlam:
#2219

Previously this check only ran on ways explicitly tagged `access=yes` or similar `access=*`-tagged (`*`!= `psv`/`no`/`private`) highways. This was probably by mistake (nothing to suggest it was intended in [the commit which added it](https://github.com/osm-fr/osmose-backend/commit/27a170ec70ca8cea9732c82e31fb6fc88a02eb9f)). Hence enable for ways without explicit `access` tag.

Furthermore, whitelist `busway` and `bus_guideway` (access=no by default; I can understand they'd be permitted to cross roundabouts sometimes; and consistent with access=psv being allowed). Also whitelist `highway=path` (like `footway`, which is already permitted) as parks may exist within large roundabouts. Example with path: https://www.openstreetmap.org/way/16059651

### DIFF
--- a/analysers/analyser_osmosis_roundabout_level.py
+++ b/analysers/analyser_osmosis_roundabout_level.py
@@ -247,8 +247,8 @@ FROM
         roundabout.nodes && ways.nodes[2:array_length(ways.nodes,1)-1]
 WHERE
     NOT ways.is_construction AND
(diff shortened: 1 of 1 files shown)


frodrigo:
Thank you.

PR #2222, opened 2024-05-16
bump database to postgresql 15/postgis 3.4

stephankn:
I tested it with my set of countries and did not detect any problem on execution.


frodrigo:
Thank you.

Issue #2200, opened 2024-04-14
Check for trees that are too big


ivanbranco:
According to the [Guinness World Records](https://www.guinnessworldrecords.com/world-records/living-tree-with-greatest-girth#:~:text=Generally%20speaking%2C%20however%2C%20the%20largest,m%20(141%20ft)%20recorded), the greatest circumference for a tree is 43m. I suggest implementing a check for trees with circumferences exceeding this value, as they are likely to be errors.

For instance, we currently have 2001 nodes in the database with a circumference of 45: https://overpass-turbo.eu/s/1JUU. Osmose could suggest to check them and 
1. delete the value if plainly wrong or
2. fix 'circumference=45' to 'circumference=0.45' or 'circumference=45 cm'

Another check could be on height. Trees taller than [Hyperion](https://en.wikipedia.org/wiki/Hyperion_(tree)) (tallest tree in the world with 116m) are likely to be errors.

Famlam:
https://overpass-turbo.eu/s/1JWd
62638 trees worldwide affected by the circumference limit of 43m.

I suspect this might be interesting for JOSM, for their numerical validation rules (which we synchronize with Osmose). That way it benefits more people ànd people who add those trees get a warning prior to uploading it in the first place.

Could you report it to them please?
If they decline, we can still add it to Osmose directly :)

sekerob:
Clicked on one where curiously the crown was given a lower value of 35 v 45. The JOSM tool with the measured values plugin gives good indication of crown size, seen many exceeding 150-200 m2. Maybe add a crown/trunk ratio test to home in further, venture to think if a mapper goes into mapping these, both will be tagged.

![image](https://github.com/osm-fr/osmose-backend/assets/23308504/f6daa488-c5fb-4f17-b99b-9fb48b9116b8)

Famlam:
I'm not a biologist, so I'm not sure if there's trees where it could be true (a cactus perhaps?). I played around a bit to get some stats though:

This mapcss rule that compares `circumference` with `diameter_crown` and warns if the latter is smaller than the former gives 62489 results, of which 17620 with a circumference <= 43m:
```css
node[circumference][diameter_crown][to_float(tag("circumference")) > to_float(tag("diameter_crown"))] {
  throwWarning: tr("circumference larger than diameter_crown");
}
```

This mapcss rule finds (I think) all 73369 trees with a circumference > 43 m. Still excluding trees measured in feets, inches & other uncommon units of course.
```css
[...]

PR #2224, opened 2024-05-17
Implement tree-vs-wiki check

Famlam:
Implements the request at https://github.com/osm-fr/osmose-backend/issues/2200#issuecomment-2072199324. Note: this is only part of the issue, it doesn't close it.

1) Check for missing `leaf_type` or `leaf_cycle` tags based on the tag `species`, using the [wiki table](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dtree/List_of_Species)
2) Check for mismatching `leaf_type` or `leaf_cycle` tags versus the tag `species`, using the [wiki table](https://wiki.openstreetmap.org/wiki/Tag:natural%3Dtree/List_of_Species)

Notes:
- As it would lead to excessive reports, and a different analyser already handles wikidata, the `species:wikidata` value isn't actively used for the checks; instead, it's just added in the fix suggestion.
- I've only enabled it for `natural=tree` and `natural=tree_row`, because I consider it quite unlikely that an entire forest (or so) is of a single species.
- Note, the "missing properties" class will have about 1M entries worldwide. If they're too concentrated in certain countries (e.g. added in imports or so) we might have to disable this class in these countries.

@ivanbranco I have one question for you, which I started wondering about while testing:
Is a tree that is e.g. `deciduous` or `evergreen`, always `deciduous` or `evergreen` worldwide, or should we (for a value of deciduous) also permit the tree being tagged by `semi_deciduous` (and the same for `semi_evergreen`)? E.g., is the `semi_` climate-dependent or truly established on a per-species basis?

### DIFF
--- a/plugins/TagFix_Tree.py
+++ b/plugins/TagFix_Tree.py
@@ -1,0 +1,149 @@
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights osmose project 2024                                        ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from modules.OsmoseTranslation import T_
+from plugins.Plugin import Plugin
+from modules.downloader import urlread
+
+
+class TagFix_Tree(Plugin):
+
+    def _read_leaf_properties_table(self):
+        # The documented values, excluding:
+        # - mixed leaf types/cycles: not compatible with `species`
+        # - leaf_cycle = semi_*: might be climate dependent and unclear difference between semi_evergreen and semi_deciduous, see #2224 first comment
+        allowed_leaf_type = ("broadleaved", "needleleaved", "leafless")
+        allowed_leaf_cycle = ("evergreen", "deciduous")
+
+        data = urlread(u"https://wiki.openstreetmap.org/w/index.php?title=Tag:natural%3Dtree/List_of_Species&action=raw", 1)
+        data = list(map(lambda x: list(filter(lambda z: len(z) > 0, map(lambda y: y.strip(), x.split("|")))), data.split("|-")[1:-1]))
+        species_map = {}
+        for row in data: # data: list of [species, species:wikidata, leaf_cycle, leaf_type]
+            this_species = {}
+            if row[2] in allowed_leaf_cycle:
+                this_species['leaf_cycle'] = row[2]
+            if row[3] in allowed_leaf_type:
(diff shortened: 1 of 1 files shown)


ivanbranco:
> @ivanbranco I have one question for you, which I started wondering about while testing:
> Is a tree that is e.g. deciduous or evergreen, always deciduous or evergreen worldwide, or should we (for a value of deciduous) also permit the tree being tagged by semi_deciduous (and the same for semi_evergreen)?

In my knowledge is species-related, but now you make my doubt.
It's also difficult to tell using OSM data, the species=* with most leaf_cycle=semi_deciduous occurences in the database is "Tipuana tipu", so I took a look at Overpass:
image
image
In yellow semi_deciduous, in red deciduous. I doubt the climate in Barcelona differs that much, so I'm more prone to say it's an error in this case.

I'm not knowledgeable in this matter. I also do not fully understand the difference between semi-deciduous and semi-evergreen tbh, even Wikipedia lists both terms on the same page: [Semi-deciduous](https://en.wikipedia.org/wiki/Semi-deciduous) ("Semi-deciduous or semi-evergreen is a botanical term which refers to..."). I think these distinctions are difficult for a typical user to evaluate. Usually, people determine the tag by simply seeing if the tree has leaves in winter or not. They might not notice if the tree loses its leaves only for a short period or loses most but not all of them, etc.

If the matter is not cleared the validator could just ignore the species with semi- values occurrences? Or allow semi_deciduous on deciduous species and semi_evergreen to evergreen species? There are only 8 species with more than >10 semi_deciduous occurrences. The rest of them are likely errors, example: https://www.openstreetmap.org/node/11346613681 (I doubt this is the only semi-deciduous Acer platanoides among the +80.000 mapped, in Berlin climate).

Famlam:
Ok, thank you. Indeed, it's better to be safe than sorry, so I've modified it so it'll no longer touches the `leaf_cycle` tag if it's set to `semi_*` (or if the wiki suggests to make it `semi_*`; currently 0 cases)

> I also do not fully understand the difference between semi-deciduous and semi-evergreen tbh

I agree it's not really clear. The OSM wiki distinguishes it as:
["_Plants that lose their foliage for a very short period, when old leaves fall off and new foliage growth is starting._"](https://wiki.openstreetmap.org/wiki/Tag%3Aleaf_cycle%3Dsemi_deciduous) vs
["_Plants that lose most, but not all, of their foliage for a fraction of the year._"](https://wiki.openstreetmap.org/wiki/Tag%3Aleaf_cycle%3Dsemi_evergreen)
However, that sounds relatively vague (define "most", define "very short") and I can imagine this will differ from winter to winter.

> The rest of them are likely errors, example: https://www.openstreetmap.org/node/11346613681 (I doubt this is the only semi-deciduous Acer platanoides among the +80.000 mapped, in Berlin climate).

It may be located under a warm street light :)

frodrigo:
@Famlam what about the quantity of raised issues by the class 31201 ?

Famlam:
> @Famlam what about the quantity of raised issues by the class 31201 ?

It's indeed a concern of mine, as there's about 1 million worldwide (based on the [TagInfo counts](https://taginfo.openstreetmap.org/keys/species#combinations) for `species` and the two `leaf_*` keys). Hence my remark:
> Note, the "missing properties" class will have about 1M entries worldwide. If they're too concentrated in certain countries (e.g. added in imports or so) we might have to disable this class in these countries.

My thought was: likely it'll be concentrated in a few countries where trees are added with automatic imports.
If this is true, we should disable this class (but not 31202) for those countries.
If this is false, class 31201 should (unfortunately) be removed for being too spammy.

It would be great if it is possible to run this analyser once under the useDevItem flag? I'm afraid that running it on osm110 for all countries will take several days.

frodrigo:
I am i favour of disabling 31201. Because it is about improving the data and not fixing issues and too numerous.

Famlam:
Ok, I've removed the check for missing properties, so now it only checks for mismatches

PR #2226, opened 2024-05-23
add unsigned_ref to highway_without_ref tags

emersonveenstra:
While uncommon, it is possible for a motorway to only have an `unsigned_ref` https://wiki.openstreetmap.org/wiki/Key:unsigned_ref, see https://www.openstreetmap.org/relation/1648091 and https://www.openstreetmap.org/relation/282102 for examples

### DIFF
--- a/analysers/analyser_osmosis_highway_without_ref.py
+++ b/analysers/analyser_osmosis_highway_without_ref.py
@@ -34,10 +34,10 @@ FROM
     relation_members.member_id = ways.id
   LEFT JOIN relations ON
     relations.id = relation_members.relation_id AND
-    relations.tags ?| ARRAY['noref', 'ref', 'nat_ref', 'int_ref']
+    relations.tags ?| ARRAY['noref', 'ref', 'nat_ref', 'int_ref', 'unsigned_ref']
 WHERE
   ways.highway = 'motorway' AND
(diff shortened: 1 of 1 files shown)


Famlam:
It's a very strange tag... 🤔 

Whether a ref is a ref or an unsigned_ref, seems to be determined by the frequency and size of the marker, not by whether its a ref or not?
I mean, the polish example is clearly signposted, if I'd call the emergency services to help me on a street with that number they'll probably know exactly which road I'm on, and it's verifiable on the ground, yet it is supposed to be tagged as unsigned_ref...
That sounds to me like building=house vs building=house_behind_tree

emersonveenstra:
I think the example photos are the small signposts because it is hard to photograph an example of a road with no signage :) But all the examples I've seen don't have any signage at all.

Famlam:
I could argue for such cases, but the wiki is very clear that it also includes roads with signs that are just small. And the examples are pretty - in the phrasing of the wiki - prominently ~sign~posted, also showing examples with huge signs.
(Ironically, one of the examples is also used on the wiki for official_ref. Which would make more sense to me already)

But I'll leave it to the others to decide

frodrigo:
I close. Ask for reopen if not agree.

PR #2230, opened 2024-06-01
Fix more than one location on overlapping highways

Famlam:
Attempts to fix the issue reported by Tomas.

Cause:
- https://www.openstreetmap.org/way/145265140 and https://www.openstreetmap.org/way/1175247716 fully overlap
- This results in `ST_Intersection` to output a non-point geometry (can be linestring for complete overlaps, but also multipoint for ways that cross each-other more than at the connecting node)
- This gives multiple locations in the output:
```xml
<error class="2">
<way id="145265140" version="14" user="">
<tag k="lit" v="yes" />
<tag k="foot" v="private" />
<tag k="name" v="Injektorweg" />
<tag k="access" v="private" />
<tag k="source" v="gps" />
<tag k="highway" v="unclassified" />
<tag k="surface" v="asphalt" />
<tag k="maxspeed" v="5" />
<tag k="lane_markings" v="no" />
</way>
<location lat="53.57444925572931" lon="9.879802518076215" />
<location lat="53.574370900000005" lon="9.879850600000001" />
</error>
```

### DIFF
--- a/analysers/analyser_osmosis_roundabout_level.py
+++ b/analysers/analyser_osmosis_roundabout_level.py
@@ -193,6 +193,7 @@ FROM
         ra1.a_id != ra2.a_id
 WHERE
     ST_Intersects(ra1.connection_sublinestring, ra2.connection_sublinestring) AND
+    ST_NPoints(ST_Intersection(ra1.connection_sublinestring, ra2.connection_sublinestring)) = 1 AND
     NOT ra1.is_oneway
 ORDER BY
     ra1.a_id,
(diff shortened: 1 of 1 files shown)


frodrigo:
Is not better to extract only one point from the intersection than only report intersection limited to one point ?

Famlam:
Well, I thought that the majority of the intersections with more than one intersecting point probably required a different fix (like removing the overlap). 

For linestring-dimension overlaps this is certainly the case. Two (non-area) highways should never overlap. (Hence I filed an issue to add an analyser to detect these cases)

For multipoint intersections it would become difficult to determine which of the intersection points is the right one. And I expect such cases to be extremely rare (given the 100m length limit).

Would you prefer that I get the first point, regardless of whether that's the most appropriate one?
(If so, do you have any suggestion for efficient ways to deal with all possible geometries (point, multipoint, linestring, multilinestring))?

Famlam:
Updated the PR. I kept the change as a separate commit in case you'd prefer the original implementation after all: for linestring overlaps it's very likely a false positive, especially if one completely covers the other linestring. (I could add a check for this, if desired).

frodrigo:
Looks good to me. Thank you.

PR #2234, opened 2024-06-12
Fix mapcss `regexp_match` behavior

Famlam:
The current implementation of `regexp_match` deviates from the expected behavior in three aspects:
1) JOSM requires a full match of the string, while the current implementation of Osmose allows for partial matches.
(This hasn't lead to issues because JOSM is consistent with the usage of `^` and `$`, effectively giving the same behavior, but I spotted it while fixing the others)

2) Osmose returned `[]` upon a non-match, JOSM `null`. Hence, the mapcss `any` function behaved different if combined with regexp_match: `any(None, "X") == "X"`, while `any([], "X") == []`

3) Osmose failed to get the correct capture group contents in case there's only one capture group.
This was due to the `flatten(z)` function, which would receive e.g. `z = ['string']` , which it changed to `['s', 't', 'r', 'i', 'n', 'g']`

### DIFF
--- a/mapcss/mapcss_lib.py
+++ b/mapcss/mapcss_lib.py
@@ -172,9 +172,6 @@ class str_value_(str):
 
 None_value = str_value(None)
 
-def flatten(z):
-    return [x for y in z for x in y]
-
 uncapture_param_re = re.compile(r'\{([0-9]+\.[a-z]+)\}')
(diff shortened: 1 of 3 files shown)


frodrigo:
Lot of not previously seen issues. Thank you for the fixes !

PR #2235, opened 2024-06-12
Fix 2233 - allow K or nm colors

Famlam:
Allows colors to be expressed in kelvin or nanometer units per the wiki for light:colour.
To keep updating the regex simple, I've changed it for all keys, so the regex is the same everywhere too.

For review's purposes: the actual change is at the end and middle of the regex, where I replaced `|#[^;]*` by `|#[^;]+|[1-9][0-9]*(\.[0-9]+)? (K|nm)`


Famlam:
Hmm, ok, the 'backslash' issue seems to be here too. And I now see that they are only permitted in the JOSM files because they're [excluded from pylama](https://github.com/osm-fr/osmose-backend/blob/dev/pylama.ini#L2C1-L3C1)

Filed #2236 for this

frodrigo:
Waiting for #2236 fixed before merge this PR.

Famlam:
Ok, marking as draft for now then (as I'd need to regenerate the python file after that's merged)

frodrigo:
Now ready for review ?

Famlam:
Yes, sorry, forgot to mark it as such

PR #2240, opened 2024-06-15
Implement `to_short, to_int, to_long, to_float, to_double` mapcss functions

Famlam:
Implement `to_short, to_int, to_long, to_float, to_double` mapcss functions

Thereby I'm neglecting the precision, since Python doesn't really differentiate between short/int/long or float/double.

Note that JOSM throws an exception rather than None (null) for the numeric conversions when they are impossible. E.g. `to_int(3.14)` in JOSM gives `Cannot add MapCSS rule: java.lang.NumberFormatException: For input string: "3.14"`.

Although not in use by rules in Osmose right now, it's for example used in https://josm.openstreetmap.de/wiki/Rules/Runways and I proposed it (without realizing it wasn't implemented) in one of the comments of issue 2200.

### DIFF
--- a/mapcss/mapcss_lib.py
+++ b/mapcss/mapcss_lib.py
@@ -585,18 +585,43 @@ def tag_regex(tags, regex):
 
 #to_short(str)
 #    returns the string argument as a short [since r16110]
+def to_short(string):
+    try:
+        return int(string)
+    except:
+        return None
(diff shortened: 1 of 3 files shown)


frodrigo:
Ok. Good to me, thank you.

PR #2245, opened 2024-06-22
typo: Rename US Georgia subdivision northweast to northwest

ComradeRamen:
https://github.com/julien-noblet/download-geofabrik/pull/323
https://github.com/osm-fr/osm-extract-replication/pull/1


jocelynj:
Could you change the commit message by something like this?
```
typo: Rename US Georgia subdivision northweast to northwest
```

I’ve already updated the diffs generation here: http://download.openstreetmap.fr/replication/north-america/us-south/georgia/, keeping the old name until all Osmose backends are updated.

jocelynj:
@frodrigo : how do you want to proceed for the update, as we need to update the frontend database to keep history?

frodrigo:
I guess we have first to rename in the frontend and then in the backend.
The frontend will block upload until the backend send data with the new name.

frodrigo:
Rename in frontend db, and backend.

@jocelynj now it has to be renamed in buildbot.

Issue #2058, opened 2023-10-31
geometry error for PL - osmosis_building_in_polygon


frodrigo:
Reported from mail of @tkas 

for few days I got an error for analyser_osmosis_building_in_polygon for poland_lubelskie. Seems to have something in the data the library for parsing is not able to handle:

(current whole log on https://osm.fit.vutbr.cz/osmose/poland_lubelskie.log)

SELECT DISTINCT ON (poly_landuse.type_id)
  poly_landuse.type_id,
  buildings.id,
  -- Use the building location to point out the error, even though the landuse should be changed
  ST_AsText(way_locate(buildings.linestring)),
  buildings.tags->'building',
  poly_landuse.landuse
FROM
  buildings
  JOIN poly_landuse ON
    ST_Contains(poly_landuse.poly_proj, buildings.polygon_proj)
WHERE
  -- ignore e.g. small utility buildings that happen to be on the farmland or are fully surrounded by farmland
  buildings.area > 36 AND
  (NOT buildings.tags?'location' OR buildings.tags->'location' != 'underground')
ORDER BY
  poly_landuse.type_id
ESC[0m
2023-10-30 09:15:28   ESC[0;31merror: error on analyse osmosis_building_in_polygon...ESC[0m
2023-10-30 09:15:28     Traceback (most recent call last):
2023-10-30 09:15:28       File "/home/osmose/osmose-backend.git/./osmose_run.py", line 275, in execc
2023-10-30 09:15:28         analyser_obj.analyser()
2023-10-30 09:15:28       File "/home/osmose/osmose-backend.git/analysers/Analyser_Osmosis.py", line 290, in analyser
2023-10-30 09:15:28         self.analyser_osmosis_common()
2023-10-30 09:15:28       File "/home/osmose/osmose-backend.git/analysers/analyser_osmosis_building_in_polygon.py", line 102, in analyser_osmosis_common
2023-10-30 09:15:28         self.run(sql11, lambda res: {
2023-10-30 09:15:28       File "/home/osmose/osmose-backend.git/analysers/Analyser_Osmosis.py", line 583, in run
2023-10-30 09:15:28         self.run00(sql, callback_package)
2023-10-30 09:15:28       File "/home/osmose/osmose-backend.git/analysers/Analyser_Osmosis.py", line 535, in run00
2023-10-30 09:15:28         self.giscurs.execute(sql)
[...]

PR #2255, opened 2024-07-03
Add common table polygons

frodrigo:
Combine multipolygones table with polygons from way table.

First step to support polygons type in merge analysers (see POC #2256)

Can also be used in SQL analyser.

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -206,6 +206,39 @@ CREATE INDEX idx_multipolygons_tags ON {0}.multipolygons USING gist (tags);
 ANALYZE {0}.multipolygons;
 """
 
+    sql_create_polygons = """
+CREATE UNLOGGED TABLE {0}.polygons AS
+SELECT
+    'R' AS type,
+    id,
+    tags,
+    poly,
+    poly_proj
+FROM
+    {0}.multipolygons
+WHERE
+    is_valid
+UNION ALL
+SELECT
+    'W' AS type,
+    id,
+    tags,
+    ST_MakePolygon(linestring) AS poly,
+    ST_MakePolygon(ST_Transform(linestring, {1})) AS poly_proj
+FROM
+    ways
+WHERE
+    tags != ''::hstore AND
+    is_polygon AND
+    ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1})))
+;
+
+CREATE INDEX idx_polygons ON {0}.polygons USING gist(poly);
+CREATE INDEX idx_polygons_proj ON {0}.polygons USING gist(poly_proj);
+CREATE INDEX idx_polygons_tags ON {0}.polygons USING gist(tags);
+ANALYZE {0}.polygons;
+"""
+
     sql_create_buildings = """
 CREATE UNLOGGED TABLE {0}.buildings AS
 SELECT
@@ -368,6 +401,15 @@ ANALYZE {0}.buildings;
                 elif table == 'touched_multipolygons':
                     self.requires_tables_build(['multipolygons'])
                     self.create_view_touched('multipolygons', 'R')
+                elif table == 'polygons':
+                    self.requires_tables_build(["multipolygons"])
+                    self.giscurs.execute(self.sql_create_polygons.format(self.config.db_schema.split(',')[0], self.config.options.get("proj")))
+                elif table == 'touched_polygons':
+                    self.requires_tables_build(["polygons"])
+                    self.create_view_touched('polygons', ['W', 'R'])
+                elif table == 'not_touched_polygons':
+                    self.requires_tables_build(["polygons"])
+                    self.create_view_not_touched('polygons', ['W', 'R'])
                 elif table == 'buildings':
                     self.giscurs.execute(self.sql_create_buildings.format(self.config.db_schema.split(',')[0], self.config.options.get("proj")))
                 elif table == 'touched_buildings':
@@ -497,10 +539,11 @@ SELECT
 FROM
     {0}
     JOIN transitive_touched ON
-        transitive_touched.data_type = '{1}' AND
-        {0}.{2} = transitive_touched.id
+        transitive_touched.data_type = ANY(%s) AND
+        {0}.{1} = transitive_touched.id
 """
-        self.giscurs.execute(sql.format(table, type, id))
+        type = type if isinstance(type, (list, tuple)) else [type]
+        self.giscurs.execute(sql.format(table, id), (type, ))
 
     def create_view_not_touched(self, table, type, id = 'id'):
         """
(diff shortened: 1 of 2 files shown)


analysers/Analyser_Osmosis.py
```diff
@@ -204,6 +204,34 @@ class Analyser_Osmosis(Analyser):
...
+    'R' AS type,
+    *
+FROM
+    {0}.multipolygons
+UNION ALL
+SELECT
+    'W' AS type,
+    id,
+    tags,
+    CASE WHEN ST_IsValid(linestring) = 't' AND ST_IsSimple(linestring) = 't' AND ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(linestring) ELSE NULL END AS poly,
```
Famlam, line 221:
```suggestion
    ST_MakePolygon(linestring) AS poly,
```
the checks for validity of the linestring are already included in the definition of `is_polygon`
https://github.com/osm-fr/osmose-backend/blob/dev/osmosis/ImportDatabase_Prepare.sql#L15-L17
(I suspect the ST_Transform was unintended here; they can differ, see https://github.com/osm-fr/osmose-backend/issues/2058#issuecomment-1788769091)

analysers/Analyser_Osmosis.py
```diff
@@ -204,6 +204,34 @@ class Analyser_Osmosis(Analyser):
...
+    CASE WHEN ST_IsValid(linestring) = 't' AND ST_IsSimple(linestring) = 't' AND ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(ST_Transform(linestring, {1})) ELSE NULL END AS poly_proj,
```
Famlam, line 222:
```suggestion
    CASE WHEN ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(ST_Transform(linestring, {1})) ELSE NULL END AS poly_proj,
```
Same as above, except that here the validity of the poly_proj has to be checked as only the non-projected version is already checked for validity

analysers/Analyser_Osmosis.py
```diff
@@ -368,6 +396,12 @@ def requires_tables_build(self, tables):
                 elif table == 'touched_multipolygons':
                     self.requires_tables_build(['multipolygons'])
                     self.create_view_touched('multipolygons', 'R')
+                elif table == 'polygons':
+                    self.requires_tables_build(["multipolygons"])
+                    self.giscurs.execute(self.sql_create_polygons.format(self.config.db_schema.split(',')[0], self.config.options.get("proj")))
+                elif table == 'touched_polygons':
```
Famlam, line 402: Although not used yet, it may be useful to already add the `not_touched_polygons` table? It's not created anyway until it's requested by an analyser.

analysers/Analyser_Osmosis.py
```diff
@@ -204,6 +204,34 @@ class Analyser_Osmosis(Analyser):
...
+FROM
+    {0}.multipolygons
+UNION ALL
+SELECT
+    'W' AS type,
+    id,
+    tags,
+    CASE WHEN ST_IsValid(linestring) = 't' AND ST_IsSimple(linestring) = 't' AND ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(linestring) ELSE NULL END AS poly,
+    CASE WHEN ST_IsValid(linestring) = 't' AND ST_IsSimple(linestring) = 't' AND ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(ST_Transform(linestring, {1})) ELSE NULL END AS poly_proj,
+    true AS is_valid
```
Famlam, line 223: More a question than feedback, but: 
if the way-polygons are always valid, why not simply restrict the multipolygon-polygons to valid ones, and drop the is_valid output here?

Also this is a bit confusing because is_valid may be true, while `poly_proj` can be `NULL` due to being invalid (in rare cases)

Famlam (commented):
Just went through it very quickly, so I might have missed things, but these are my very first thoughts

frodrigo:
@Famlam thank you for review, I fixed all according to you.

doc/3-SQL-basics.md
```diff
@@ -173,6 +173,7 @@ The available common tables are (see full definition in [Analyser_Osmosis.py](ht
 * `highway_ends`: the start and ends of all highways ways.
 * `buildings`: from ways (and not from multipolygon relations), with tags normalization and re-projected in local country _projection_ as _polygons_.
 * `multipolygons`: multipolygon relations with relation id, relation tags, geometry and projected geometry, and whether the multipolygon is valid.
+* `polygons`: polygon from relations and ways with osm object type, id, tags, geometry and projected geometry, and whether the multipolygon is valid.
```
Famlam, line 176:
```suggestion
* `polygons`: polygon from valid relations and ways with osm object type, id, tags, geometry and projected geometry
```

analysers/Analyser_Osmosis.py
```diff
@@ -204,6 +204,38 @@ class Analyser_Osmosis(Analyser):
...
+    {0}.multipolygons
+WHERE
+    is_valid
+UNION ALL
+SELECT
+    'W' AS type,
+    id,
+    tags,
+    ST_MakePolygon(linestring) AS poly,
+    CASE WHEN ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(ST_Transform(linestring, {1})) ELSE NULL END AS poly_proj
```
Famlam, line 227: To avoid having to do a NULL check for the very rare case where it isn't true, perhaps `ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1})))`can just be in the WHERE

Famlam:
Thanks!
Asides from a documentation thing and a suggestion, it looks good to me.

I do not know how `self.create_view_touched('polygons', 'W')` and `self.create_view_not_touched('polygons', 'W')` respond to the cases where type != `W` (e.g. the multipolygons), but I assume you've tested that that works fine.

analysers/Analyser_Osmosis.py
Famlam: I think this was resolved based on the commit I see below, so I'll close this

frodrigo:
> I do not know how self.create_view_touched('polygons', 'W') and self.create_view_not_touched('polygons', 'W') respond to the cases where type != W (e.g. the multipolygons), but I assume you've tested that that works fine.

I fixed that.

Famlam:
I'm not familiar with how the `giscurs.execute` function handles tuples/arrays in the second argument (can't even find the documentation/definition of that function) so I'll just assume it works.

Looks good to me :). 

You might want to update the `@param type in 'N', 'W', 'R'` line above both `create_view_*touched` functions to state that it can also be a list of those.

PR #2257, opened 2024-07-04
add noexit=yes as valid oneway endpoint

ComradeRamen:


Famlam:
How can a oneway have no exit?

ComradeRamen:
These come up quite often
image

jdhoek:
> How can a oneway have no exit?

When the oneway is not in (current) use and has `access=no` (or equivalent), but otherwise has all the trappings and road-legal signs of a oneway road. And of course there are real world examples of actual oneway streets with no exit due to bad signage or other half-assed road planning.

Fortunately, `noexit=yes` exists to inform QA-tools that the mapped situation is correct. It should be valid for these cases too.

Famlam (approved):

frodrigo:
I am prefer to exclude way with `access=no` from the check.

A no-exit oneway could exist with bad signage, but no in the usage. And I think we map the usage in OSM. Nevertheless such case could just be marked as false positive.

Famlam:
@frodrigo Although I see your point, I fear "just" excluding `access=no` would also whitelist for example any busway tagged explicitly with `access=no + bus=yes` or so, and thus prevent errors on such ways from being detected. (It could work if all other (common) access keys are checked to not be present)

ComradeRamen:
Most of these I come across have noexit=yes tagged. I believe this is also the method geofabrik osminspector routing view uses.

frodrigo:
https://osmose.openstreetmap.fr/en/issues/false-positive?item=1210&class=3

Most are false positive are in poland_mazowieckie  and morocco. It is your area ?

I think it may not have good way to do this. One way with no exist, is also a ground issue.

jdhoek:
> I think it may not have good way to do this. One way with no exist, is also a ground issue.

The example linked to by @ComradeRamen is just one valid example of oneway streets ending in a dead end. They exist (and there is no ground issue there). Mappers can indicate that this is correct with `noexit=yes`.

frodrigo:
From the wiki

> Its purpose is to either confirm a way ends when this way has limited visibility on satellite imagery (e.g. blocked by trees), or to inform quality assurance software or a human reader that an otherwise suspicious tag or **road layout preventing passing further than the end of a road is perfectly intentional** 

> Mappers can indicate that this is correct with noexit=yes.

@jdhoek ok, you are right.

I am not fully convinced, but ok let's merge this.

Issue #2254, opened 2024-07-02
False positive in test "Abbreviation of number should be in lowercase"


muralito:
I don't know if osmose can handle rules based on the location of the analyzed object, but in spanish names (at least in Argentina and Uruguay) is common use, for the name of schools, to abbreviate "número" (number) in upper case.

See https://osmose.openstreetmap.fr/en/issue/c98241f9-2e49-bc90-88c5-ad7162690447
and https://osmose.openstreetmap.fr/en/issues/open?item=5010&class=905

The name of this school https://www.openstreetmap.org/node/3841949274 is correct, so this kind of situations shouldn't be reported as an issue, and similar issues closed as false positives if possible.

frodrigo:
In Spain there is not just Spanish. But look ok anyway.

frodrigo:
PR merged. It will require manual clearing in other Spanish speaking countries.

frodrigo:
I deleted all issues 5010/905. Too hard to select. Let's the right one be back.

muralito:
> I deleted all issues 5010/905. Too hard to select. Let's the right one be back.

Thanks.

Famlam:
I think this is resolved? Only 8 left :)

PR #2260, opened 2024-07-05
Limit Spanish to Iberian Spain (=the country Spain)

Famlam:
See #2254

Limit the check for the Spanish language to Iberian Spain only.
This also matches what [Wikipedia](https://en.wikipedia.org/wiki/Numero_sign#Spanish) says for the symbol in the Spanish language

This patch changes it from:
`languages=fr,es,it,pt`
to 
`languages=fr,it,pt` or `country=ES`

(Untested)
It'll probably require manual clearing of the results for South-America


frodrigo:
In Spain there is not just Spanish. But look ok anyway.

Issue #2251, opened 2024-07-02
Allow litre (l) as an explicit unit for capacity if a volume is specified (e.g. for amenity=waste_disposal or amenity=grit_bin)


jajajaneeneenee:
The value for capacity of a volume is normally specified in m³, but litre ("l") is also allowed (= documented) as an explicit specification (see [https://wiki.openstreetmap.org/wiki/Map_features/Units](https://wiki.openstreetmap.org/wiki/Map_features/Units)).

The integer capacity value check now reports an error if, for example, "capacity=350 l" is specified for amenity=grit_bin ("capacity" value "350 l" is not an integer).

I think, it should not throw an error. And not with m³ either. (Other volume units: I am not sure ... see below. But at least  "l" seems to be established and it is documented).

And, for example, for grit bins, the content is usually specified in liters and not in m³ on the container and is therefore probably preferred to be specified that way. I mention grit bin because I first encountered the problem with it – although there are currently only 147 cases of grit_bin + capacity (but growing ...) – but almost all of them with "l" (exact numbers see below). So this combination now has a nearly 100% Osmose error rate – I would say false positives. Therefore a special case ... But it also affects other tags (see below).

For waste disposals, there seem to be some incorrect information such as capacity=3200 (e.g. https://www.openstreetmap.org/node/10210997092), which would have to be a pretty large container... (capacity=800, etc. also occurs frequently). I think the unit "l" for litre is missing here – maybe because of incorrect validator messages people are using an integer only? Hoping someone will guess the right unit? Not a good idea in my opinion ... This is also why I am writing this issue.

I spontaneously think of these cases:

- amenity=waste_disposal (> 10 k uses with capacity=*, zero cases with "l", 360 cases with values > 100 – probably "l" missing)
- amenity=waste_basket (approx. 1600x used with capacity=*, 3 cases with "l" and 1335x with "litres")
- amenity=recycling (approx. 7700x used with capacity=*, 172 cases with "l")
- man_made=storage_tank (approx. 2000x used with capacity=*, 12 cases with "l", other cases with "gal" etc.)
- amenity=grit_bin (147x used with capacity=*, 143x with "l", 3x "l" missing => **>99% of the values in litre**)
- water=reservoir (> 100x used with capacity=*; not checked any further, but megalitres or hm3 seem to be used in some cases)
- emergency=water_tank (not checked any further)

[...]

PR #2261, opened 2024-07-06
Implement `siunit_length` mapcss function

Famlam:
Implements the new `siunit_length` mapcss function.

Rules using this function were [added in the JOSM mapcss rules today](https://josm.openstreetmap.de/changeset?old_path=%2Fjosm&old=19129&new_path=%2Fjosm&new=19129).
(If desired I can update the mapcss rules in this PR)

I'm using Pint for the unit conversions. Pint would also be useful for #2251 (as suggested by @jocelynj) and probably other places as well (plugin Number.py for example).

### DIFF
--- a/plugins/tests/test_mapcss_parsing_evaluation.validator.mapcss
+++ b/plugins/tests/test_mapcss_parsing_evaluation.validator.mapcss
@@ -561,3 +561,27 @@ node[to_float("3.14e1") = 31.4][to_double("3.1415e1") = 31.415] {
   throwWarning: "test";
   assertMatch: "node x=y";
 }
+
+node[tag_length][siunit_length(tag(tag_length)) == "1.2"] {
+  throwWarning: tr("test unit conversion {0} -> {1}", "{0.value}", siunit_length(tag(tag_length)));
+  assertMatch: "node tag_length=1.2";
+}
+node[tag_length][siunit_length(tag(tag_length)) == tag(b)] {
+  throwWarning: tr("test unit conversion {0} -> {1}", "{0.value}", siunit_length(tag(tag_length)));
+  assertMatch: "node tag_length=\"  1,3 m \" b=1.3";
+  assertMatch: "node tag_length=1.4m         b=1.4";
+  assertMatch: "node tag_length=150cm        b=1.5";
+  assertMatch: "node tag_length=1600.0mm     b=1.6";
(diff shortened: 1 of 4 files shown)


plugins/tests/test_mapcss_parsing_evaluation.validator.mapcss
```diff
@@ -561,3 +561,27 @@ node[to_float("3.14e1") = 31.4][to_double("3.1415e1") = 31.415] {
...
+  assertMatch: "node tag_length=\"10ft\"     b=3.048";
+  assertMatch: "node tag_length=\"20'\"      b=6.096";
+  assertMatch: "node tag_length=\"100in\"    b=2.54";
+  assertMatch: "node tag_length=\"200\"\"    b=5.08";
+  assertMatch: "node tag_length=\"1nmi\"     b=1852";
+  assertMatch: "node tag_length=\"1mi\"      b=1609.344";
+  assertMatch: "node tag_length=\"10ft 1in\" b=3.0734";
+  assertMatch: "node tag_length=\"20'2\"\"   b=6.1468";
+  assertMatch: "node tag_length=\"-20'2\"\"  b=-6.1468";
+  assertNoMatch: "node tag_length=\"Hello\"  b=0";
```
frodrigo, line 586: Should really be 0 here ?

Famlam: No, it's just to test that it doesn't error, but there shouldn't be any match (hence the assertNoMatch). I can also add a test like this if you prefer:
`node[any(siunit_length("Hello"), "x") == "x"]`
This would literally ensure the output is None.

frodrigo: Ok, it is NoMatch, so it is ok for me.

Famlam:
I spotted two different (minor) bugs:
1. JOSM always converts to Double, I didn't, so our output differed from JOSM by a `.0` for an int -> int conversion
2. In the test string I was still using `tag(a)` while I had renamed that later to `tag_length`

Also removed a lot of unnecessary `\"`

Famlam:
Hmm, I just discovered that pint also does mathematics... E.g. input `4-5` (a range) becomes -1 rather than PintError...
This complicates it's use quite a bit...

Famlam:
New implementation in #2263

PR #2263, opened 2024-07-10
Implement `siunit_length(str)` mapcss function

Famlam:
Implements the new siunit_length mapcss function. Rules using this function are now in the JOSM mapcss files and can be updated after this PR is merged.

This PR implements it without using Pint or any other external package due to extra complications (see #2261)
I think the current implementation should also be easy to adjust in case we have to add another unit (or even another type of unit, like volume)

It also provides a separate string-parsing function which would be useful for #2251 and in plugin Number.py

### DIFF
--- a/plugins/modules/units.py
+++ b/plugins/modules/units.py
@@ -1,0 +1,110 @@
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Osmose project 2024                                        ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+
+# This module file contains functions to read and convert numbers with units
+
+import re
+
+_ftin_re = re.compile("^(-?)(?:(\\d+(?:\\.\\d+)?)(ft|')) ?(?:(\\d+(?:\\.\\d+)?)(in|\"))$") # Regex for combined feet/inch measures
+_numunit_re = re.compile(r"^(-?\d+(?:\.\d+)?) ?(\D.*)?$") # Regex for any number followed by an optional unit
+_si_prefixes = {
+    "n": 1E-9, "u": 1E-6, "µ": 1E-6, "m": 0.001, "c": 0.01, "d": 0.1, "da": 10, "h": 100, "k": 1000, "M": 1E6, "G": 1E9, "T": 1E12,
+    "nano": 1E-9, "micro": 1E-6, "milli": 0.001, "centi": 0.01, "deci": 0.1, "deca": 10, "hecto": 100, "kilo": 1000, "mega": 1E6, "giga": 1E9, "tera": 1E12,
+}
+
+
+# Converts strings that contain a number with a unit to a dict of {"value": float, "unit": string}
+# Input:
+#   string: the string to parse, e.g. "22.4 km"
+#   defaultUnit: the unit if not specified (default: None)
+# Returns:
+#   None if conversion isn't possible
+#   Otherwise a dict with keys:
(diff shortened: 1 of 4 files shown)


plugins/tests/test_mapcss_parsing_evaluation.validator.mapcss
```diff
@@ -561,3 +561,34 @@ node[to_float("3.14e1") = 31.4][to_double("3.1415e1") = 31.415] {
...
+  assertMatch: "node tag_length=-1.8km       b=-1800.0";
+  assertMatch: "node tag_length=10ft         b=3.048";
+  assertMatch: "node tag_length=20'          b=6.096";
+  assertMatch: "node tag_length=100in        b=2.54";
+  assertMatch: "node tag_length=200\"        b=5.08";
+  assertMatch: "node tag_length=1nmi         b=1852.0";
+  assertMatch: "node tag_length=1mi          b=1609.344";
+  assertMatch: "node tag_length=\"10ft 1in\" b=3.0734";
+  assertMatch: "node tag_length=20'2\"       b=6.1468";
+  assertMatch: "node tag_length=-20'2\"      b=-6.1468";
```
jocelynj, line 587: Do you know if '1,600.0m' is supposed to be supported too?
Or '1 600 m'.

Famlam: The first one is not supposed to be supported: JOSM [replaces](https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/tools/Utils.java#L2034) all `,` by `.` so that would become an invalid number (`1.600.0m`). In case they update their code to only do this replacement if there's no `.` in the string yet, we should follow of course, for the mapcss part.

Regarding `1 600 m`: good point, I've updated the code of the mapcss function to match JOSMs behavior here too.

(I didn't do this in the functions in /units.py as I also want to be able to use those for e.g. validation purposes)

jocelynj: I see for '1,600.0m', and it makes sense. This looks a bit dangerous for '1,600m', which might mean '1600 m', but it doesn't matter for this mapcss function - it is more a matter of OSM data that should follow guidelines.

Famlam: Yep, well, also with mapcss it can result in false positives if used in selectors (e.g. adding a warning about an object that's too small when the comma was a true 1000-separator) but in that case it also needs to be fixed by JOSM. And in case of such a false positive, at least it'll hopefully make the user think another time about the value...

Of course we can choose to remove this replacal ourselves if desired.

jocelynj:
This looks nice, thanks.

I'm letting @frodrigo finish the review.

plugins/modules/units.py
```diff
@@ -0,0 +1,110 @@
...
+#   string: the string to parse, e.g. "22.4 km"
+#   defaultUnit: the unit if not specified (default: None)
+# Returns:
+#   None if conversion isn't possible
+#   Otherwise a dict with keys:
+#       value [float] - the number in the string
+#       unit [string or None] - the unit
+# For a string with multiple numbers/units, it converts it to a float of the largest unit (e.g. 3'4" becomes 3.33 ft)
+def parseNumberUnitString(string, defaultUnit = None):
+    if not string:
```
frodrigo, line 46: Add a check to check string is as string ?

plugins/modules/units.py
```diff
@@ -0,0 +1,110 @@
...
+    if isinstance(x, str):
+        # String input, extract value and unit
+        x = parseNumberUnitString(x, convertTo)
+    if not x:
+        return None
+    if x["unit"] == convertTo:
+        return x["value"]
+
+    # Length based conversions
+    if convertTo == "m": # default for range, height, length
```
frodrigo, line 88: Can we use pint once the parsing is done ?

Famlam: It would still require a lot more pre-parsing from our side than I initially expected. Pint for instance doesn't natively support things like `m3` (for the fix of 2251) or `"` (inch, as symbol, for siunit_length). So we'd still need a dictionary mapping for instance `'` to `ft`, `m3` to `m³` etc.

In addition, feeding pint with arguments like `value=123.45, units='m'` gives issues because Pint then complains about the use of scaling factors (it seems to assume it's milli on first glance), so it'd have to become something like `x["value"] * ureg("1" + x["unit"])` to use it (with `ureg` the function from Pint)
Lastly, it still does mathematics if there's anything in the unit. E.g. if the unit contains `30m*3` pint will simply multiply the value by 3 instead of complaining about invalid units.

So personally, after noticing these things I think I would prefer to have a function of our own until for instance Pint has an option to disable math or we need many more conversion functions. There are benefits to each, but...

frodrigo: Ok. Thank you for the explanation.

plugins/modules/units.py
Famlam: Done, good catch. I've also adjusted the other function

PR #2279, opened 2024-07-30
Update analyser_osmosis_node_like_way.py

didier2020:
false positive for parking's airplane


Famlam:
If I read the wiki correctly, only the way should have the tag, not?

didier2020:
il n'est pas indiqué dans le wiki que la description doit etre uniquement un node ou un way
dans toutes les modélisations, le way est utilisé pour décrire le chemin et le node la position de la roue avant.

https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Dparking_position

it is not indicated in the wiki that the description must only be a node or a way
in all modeling, the way is used to describe the path and the node the position of the front wheel.

Famlam:
Hmm, I read the wiki as an way _or_ node, like most other wiki's with multiple tagging options.
And in case of a way it says to tag the way (with the final node on the nose wheel position). It doesn't say to also tag that node.

Issue #2282, opened 2024-08-01
new analyser_merge_milestone for guyane, mayotte and Saint-Pierre-et-Miquelon


didier2020:
i've update analyser_merge_milestone_FR_metropole at https://github.com/osm-fr/osmose-backend/pull/2281

i want to create new analyser(s) for other milestone but
i don't know how to do for RRN_2024_973-Guyane_shp, RRN_2024_975-Saint-Pierre-et-Miquelon_shp and RRN_2024_976-Mayotte_shp because of their spécific "projection"  set in BORNAGE_TOUT.prj

frodrigo:
If the the shape file contains the project, and it is usually the case, there is nothing to do in particular about that.

didier2020:
thanks, it's done but pull requests are not ok (?)

frodrigo:
I see only the resource id change over the areas. Let's use only one analyser and just change the id according the the area.

You reuse (en rename) the existing `analyser_merge_milestone_FR_metropole.py`.

Look at this 
https://github.com/osm-fr/osmose-backend/blob/09e2bbfd30246de926d731c99bca6778e8c4f1f3/analysers/analyser_merge_healthcare_FR_finess.py#L47
https://github.com/osm-fr/osmose-backend/blob/09e2bbfd30246de926d731c99bca6778e8c4f1f3/analysers/analyser_merge_road_FR.py#L34

didier2020:
something like

```python
        if config.db_schema == 'france_guyane':
            resource = 'e9288a97-faf1-4638-b3df-571ceae401f8'
        elif config.db_schema == 'france_saintpierreetmiquelon':
            resource = '9425939b-13f5-460d-a635-fcde0e05ade5'
        elif config.db_schema == 'france_mayotte':
            resource = '65060ce7-acc5-4546-a1ce-b9b47b40383f'
        else:
            resource = '90a65602-3ca4-41d7-bf7c-23d435c916e1'

...
            SHP(SourceDataGouv(
                attribution="data.gouv.fr:Ministère de la Transition écologique",
                dataset="57a837e2c751df5b90bb5dd4",
                resource=resource),
                zip='BORNAGE_TOUT.shp'),
```

frodrigo:
> something like

Yes

frodrigo:
Use `dep_code`, it is better.

PR #2288, opened 2024-08-02
Support oversea torritories in analyser_merge_milestone_FR #2282

frodrigo:
@didier2020 I made it works. But you need to review the content of the `is_milestone()` function.

### DIFF
--- a/analysers/analyser_merge_milestone_FR.py
+++ b/analysers/analyser_merge_milestone_FR.py
@@ -22,9 +22,19 @@
 from modules.OsmoseTranslation import T_
 from .Analyser_Merge import Analyser_Merge_Point, SourceDataGouv, SHP, LoadGeomCentroid, Conflate, Select, Mapping
 
-class Analyser_Merge_Milestone_FR_metropole(Analyser_Merge_Point):
+class Analyser_Merge_Milestone_FR(Analyser_Merge_Point):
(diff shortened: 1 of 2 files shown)


analysers/analyser_merge_milestone_FR.py
didier2020, line 1: as dep can be on 3 digit 
to Filter only real milestone (not logical as DRD, FRG, CS etc...)
change 
            return row['nom_plo'][2:4] == 'PR'
by
            return (row['nom_plo'][2:4] == 'PR' or row['nom_plo'][2:5]  == 'PR')

didier2020: guadeloupe is not in the data it's guyanne FR-GF with resource e9288a97-faf1-4638-b3df-571ceae401f8

frodrigo: Both fixed.

frodrigo:
I also see a N1 in FR-PM, but excluded by is_milestone. Is it expected ?

didier2020:
is_milestone() function is ok.
#I also see a N1 in FR-PM, but excluded by is_milestone. Is it expected ?
there is no wrong filter on "route" and "nom_plo". N1 is in "lib_rte"

frodrigo:
Ok, thank you, merged.

PR #2291, opened 2024-08-03
Allow highway=proposed with footway=sidewalk

ziemek99:


Famlam:
Looks good to me. We'll still have to regenerate the python code afterwards

frodrigo:
> We'll still have to regenerate the python code afterwards

Done.

PR #2299, opened 2024-08-06
Exclude parking=layby in analyser_osmosis_parking_highway.py

rskedgell:
Add parking=layby to parking=* types which do not require an access highway (parking=street_side and parking=lane).

The wiki page for [parking=layby](https://wiki.openstreetmap.org/wiki/Tag:parking%3Dlayby) describes it as being similar to parking=street_side and also states: "There is often no physical barrier between the road and the lay-by. If there is then consider using highway=rest_area instead."

Encouraging users to add a non-existent access highway to connect a layby for routing would be unhelpful.

### DIFF
--- a/analysers/analyser_osmosis_parking_highway.py
+++ b/analysers/analyser_osmosis_parking_highway.py
@@ -67,7 +67,7 @@ FROM (
 WHERE
   tags?'amenity' AND
   tags->'amenity' = 'parking' AND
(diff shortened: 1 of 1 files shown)


frodrigo:
Right, thank you.

Famlam:
Note that this commit was merged into master, not dev

frodrigo:
> Note that this commit was merged into master, not dev.

Right. Force pushed the dev branch.

Issue #2249, opened 2024-06-28
Add check for bridges connected to roads they cross


ComradeRamen:
Bridges often get accidentally connected to the ways they cross. Examples: [307875034](https://www.openstreetmap.org/node/307875034),[8713078340](https://www.openstreetmap.org/node/8713078340),[11251420292](https://www.openstreetmap.org/node/11251420292),[2930285459](https://www.openstreetmap.org/node/2930285459) 9,000 nodes found globally when checking all highway tags

I found these by checking for nodes with connections within bridge and non-bridge ways. To avoid false positives, I checked that multiple bridge connections have the same highway tags and multiple non-bridge connections have the same highway tags, assigning single connections to nodes at the ends and double connections to nodes in the middle of a way.

I'm not sure where this analyzer would fit best

ComradeRamen:
Highway Class Counts:
motorway: 443
motorway_link: 223
trunk: 855
trunk_link: 198
primary: 1327
primary_link: 70
secondary: 1131
secondary_link: 71
tertiary: 1507
tertiary_link: 44
unclassified: 901
residential: 1171
other: 10333

Famlam:
Inconveniently it seems someone has fixed all your referred false positives.

I guess for connections with non-end nodes of the "bridge-way" this should be fairly fail-safe (and same for tunnels). E.g., this should always be bad (two uninterrupted ways of which one is a bridge, connected in a middle node)
![image](https://github.com/osm-fr/osmose-backend/assets/1315520/590f9dea-cfea-409d-9fc0-daf1d1d9b6f8)

It becomes more tricky with end nodes. E.g. a steps connecting to a bridge is not per se bad. And obviously the end node of the bridge way can be connected to anything.

Could you share the (probably) overpass query that you used to get those statistics?

ComradeRamen:
Mappers will sometimes add turn restrictions as a "fix" which splits the ways. Others have used roads as members in boundary relations and will split the roads accordingly.
![image](https://github.com/osm-fr/osmose-backend/assets/150104144/a7c2bcec-4303-475d-ac92-0aabf80e6b53)

ComradeRamen:
using this query in a python script
```
[out:xml];
way["bridge"]({south},{west},{north},{east});
>;
way(bn)["highway"];
(._;>;);
out meta;
```

Famlam:
[...]

PR #2301, opened 2024-08-08
Detect incorrect bridge/tunnel-highway connections

Famlam:
Implements a check to detect (non-bridge/tunnel) highways that are connected to the (non-start/end) nodes of tunnels or bridges.

For tunnels I only enabled it for 'higher priority' ways due to many false positives with e.g. footways etc.

For bridges I disabled boardwalks as hiking routes apparently sometimes require people to step off a boardwalk (which is typically not too high). If there's still too many false positives, we may have to disable it for 'low priority' ways too.

In all cases I excluded steps and anything that may be indoor/building related

#2249

### DIFF
--- a/analysers/analyser_osmosis_highway_tunnel_bridge.py
+++ b/analysers/analyser_osmosis_highway_tunnel_bridge.py
@@ -101,7 +101,57 @@ HAVING
     0 = SUM(CASE WHEN layer IS NULL THEN 0 ELSE 1 END) + CASE WHEN blayer IS NULL THEN 0 ELSE 1 END
 """
 
+sql40 = """
+SELECT
+    nodes.id,
+    bt_ways.id,
+    bt_connections.id,
+    ST_AsText(nodes.geom),
+    CASE
+        WHEN bt_ways.tags?'bridge' AND bt_ways.tags->'bridge'!='no' THEN 'bridge'
+        ELSE 'tunnel'
+    END,
+    nodes.id IN (bt_connections.nodes[1], bt_connections.nodes[array_length(bt_connections.nodes,1)]) OR bt_connections.is_area
+FROM
+    {0}highways AS bt_ways
+    JOIN {1}highways AS bt_connections ON
+        bt_connections.linestring && bt_ways.linestring AND
+        bt_connections.nodes && bt_ways.nodes AND
+        bt_connections.id != bt_ways.id
+    JOIN nodes ON
+        nodes.geom && bt_connections.linestring AND nodes.geom && bt_ways.linestring AND -- One is redundant, but let the planner choose
+        nodes.id = ANY(bt_ways.nodes) AND
+        nodes.id = ANY(bt_connections.nodes) AND
+        nodes.id != bt_ways.nodes[1] AND
+        nodes.id != bt_ways.nodes[array_length(bt_ways.nodes,1)]
+WHERE
+    (
+        (
+            bt_ways.highway NOT IN ('steps') AND
+            bt_connections.highway NOT IN ('steps') AND
+            bt_ways.tags?'bridge' AND bt_ways.tags->'bridge' NOT IN ('no', 'boardwalk') AND
+            (NOT bt_connections.tags?'bridge' OR bt_connections.tags->'bridge' = 'no') AND
+            (NOT bt_connections.tags?'man_made' OR bt_connections.tags->'man_made' != 'pier')
+        ) OR (
+            -- Tunnels for 'low level' highways give many false positives, hence only enable for crossing 'car roads'
+            bt_ways.level <= 4 AND bt_connections.level <= 4 AND
+            bt_ways.tags?'tunnel' AND bt_ways.tags->'tunnel' NOT IN ('no', 'avalanche_protector') AND
+            (NOT bt_connections.tags?'tunnel' OR bt_connections.tags->'tunnel' = 'no') AND
+            (NOT bt_connections.tags?'covered' OR bt_connections.tags->'covered' = 'no')
+        )
+    ) AND
+    NOT bt_ways.is_construction AND NOT bt_connections.is_construction AND
+    -- Below: filter all cases where one would for instance walk from a building directly onto a bridge
+    (NOT bt_connections.tags?'indoor' OR bt_connections.tags->'indoor' = 'no') AND
+    NOT bt_connections.tags?'location' AND
+    NOT bt_connections.tags?'level'
+"""
+
+
 class Analyser_Osmosis_Highway_Tunnel_Bridge(Analyser_Osmosis):
+
+    requires_tables_full = ['highways']
+    requires_tables_diff = ['highways', 'touched_highways', 'not_touched_highways']
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_highway_tunnel_bridge.py
```diff
@@ -101,8 +101,57 @@
...
+FROM
+    {0}highways AS bt_ways
+    JOIN {1}highways AS bt_connections ON
+        bt_connections.linestring && bt_ways.linestring AND
+        bt_connections.nodes && bt_ways.nodes AND
+        bt_connections.id != bt_ways.id
+    JOIN nodes ON
+        nodes.geom && bt_connections.linestring AND nodes.geom && bt_ways.linestring AND -- One is redundant, but let the planner choose
+        nodes.id = ANY(bt_ways.nodes) AND
+        nodes.id = ANY(bt_connections.nodes) AND
```
frodrigo, line 123: To be checked. Byt may use `<@` operator in place.

ComradeRamen:
I found 33 incorrect connections between bridge layers. https://overpass-turbo.eu/s/1Ptt 
Idk if this analyzer should incorporate layers or if its not enough cases to worry about.
image

tests/osmosis_highway_tunnel_bridge.osm
```diff
@@ -0,0 +1,215 @@
...
+  <node id='22' timestamp='2014-03-31T22:00:00Z' version='1' lat='51.86660901291' lon='5.88531909079'>
+    <tag k='note' v='No match class 4/5' />
+  </node>
+  <node id='23' timestamp='2014-03-31T22:00:00Z' version='1' lat='51.86664717616' lon='5.88547259351' />
+  <node id='24' timestamp='2014-03-31T22:00:00Z' version='1' lat='51.86657266235' lon='5.88531964268'>
+    <tag k='note' v='No match class 4/5' />
+  </node>
+  <node id='25' timestamp='2014-03-31T22:00:00Z' version='1' lat='51.86657090427' lon='5.88519467722' />
+  <node id='26' timestamp='2014-03-31T22:00:00Z' version='1' lat='51.86651016597' lon='5.88532059152'>
+    <tag k='note' v='Bad node class 4/5' />
```
ComradeRamen, line 49: Imo this being an error would result in too many false positives.

Famlam: Can you explain why? It matches your original request as far as I know (see the comment about split ways)?

Also I didn't see too many false positives (tested on 12 extracts, [results](https://github.com/user-attachments/files/16559967/bridgetunnel.zip)), most of them are very close to the bridge heads where it's debatable where exactly the bridge starts, but this isn't limited to end or non-end nodes. Same as that this isn't related to having 1 or 2 non-bridge ways connected to a node because often it involves paths leading off of the bridge which is frequently symmetrical.

More than happy to exclude all end nodes if you can explain why, but the vast majority of valid cases are also like this so it may make the analyser pretty useless :)

analysers/analyser_osmosis_highway_tunnel_bridge.py
```diff
@@ -117,16 +166,40 @@ def __init__(self, config, logger = None):
...
+
+If the highway is truely connected to the bridge or tunnel, it may only be by a short section of this highway.
+If so, you may have to split the connecting way, and adding bridge or tunnel tags only on the relevant part.'''),
+            trap = T_(
+'''There might be bad detections with connections at the bridge heads or tunnel entrances.'''))
 
         self.callback10 = lambda res: {"class":1, "data":[self.way_full, self.positionAsText], "fix":[{"+":{"bridge:structure":"beam"}}, {"+":{"bridge:structure":"suspension"}}] }
         #self.callback20 = lambda res: {"class":2, "data":[self.way_full, self.way_full, self.positionAsText] }
         #self.callback30 = lambda res: {"class":3, "data":[self.way_full, self.positionAsText] }
+        self.callback40 = lambda res: {"class": 4 if res[4] == 'bridge' else 5, "data": [self.node_full, self.way, self.way_full, self.positionAsText] }
```
Famlam, line 195: Note to self: forgot to remove `_full` for the non-bridge way, and distinct isn't necessary anymore with the ways in the output

Famlam:
> I found 89 incorrect connections between bridge layers on tertiary and higher classes. https://overpass-turbo.eu/s/1PtL Idk if this analyzer should incorporate layers or if its not enough cases to make it worth. 

I don't think I can detect such cases as being "bad" connections, especially cases like [this node](https://www.openstreetmap.org/node/31221551). The layer tag only indicates a relative height separation for everything the way crosses, e.g. layer=3 connecting to a layer=1 way is fine as it may imply that the layer=3 bridge crosses for instance a river and two other (bridge) ways, while the layer=1 way maybe only crosses a single way. In some cases it may probably be possible to "optimize" the layer tag, but that's a different (low-priority IMO) issue. See for example the first example on https://wiki.openstreetmap.org/wiki/Key:layer

analysers/analyser_osmosis_highway_tunnel_bridge.py
Famlam: It seems it's not defined for this data type, I get this warning:
`operator does not exist: bigint <@ bigint[]`

tests/osmosis_highway_tunnel_bridge.osm
ComradeRamen: My idea was that it would only count as an error if like node 39 it had multiple end connections where both ways have the same highway tag. 

If this makes it too complicated I think just checking middle nodes would be good enough.

ComradeRamen: The bt_ways only using middle nodes `nodes.id != bt_ways.nodes[1] AND
        nodes.id != bt_ways.nodes[array_length(bt_ways.nodes, 1)]` would not detect the specific case I mentioned anyway since the bridge way was split and not connected via a middle node.

Famlam: It would mostly make it more computationally intensive (always have to run it in full mode). And if a single end node would cause false positives, I have to consider that bridges are often symmetric, so then I would indeed also exclude double end nodes and only use middle nodes.

However, I'm primarily trying to understand why you expect the current method would give  too many false positives? Maybe there's a specific thing you have in mind, which may have other solutions (e.g. if it's footways, we can also just exclude those). Can you maybe give examples (see the results file shared earlier)? Or draw a hypothetical one?

Except at bridge heads, I cannot think of a way how a road can connect, via an end node, somewhere in the middle of the bridge, except if the bridge is interrupted (but then the bridge should be split in [bridge][regular][bridge]) or if the connecting way is at least partially a bridge too, even if just for a few meters until it reaches the existing bridge. Maybe very exotic constructions?

More than happy to make this change, but I'd like to understand first why it's needed :)

ComradeRamen: They are not so much false positives but they're usually [incorrect bridge tags](https://www.openstreetmap.org/node/583110103) or [incorrect connections near the end of a bridge](https://www.openstreetmap.org/node/6336921071)  [(2)](https://www.openstreetmap.org/node/7891591583) which are not navigation breaking like a bridge being accidentally glued to another way. 

I don't mind if these are included if they're put in a different class or severity.

Famlam: Ok, now I understand why you requested it. Done. 
I went for the end node vs not end-node distinction because this allows us to run the analyser in diff mode, which is much more efficient.

frodrigo:
@Famlam @ComradeRamen it is ok for both of yours, can we merge this PR ?

Famlam:
:+1: for me, it's ready for review (and thus merging)

ComradeRamen (approved):

Issue #2116, opened 2023-12-31
Soundex reported for statue despite being for streets


matkoniecz:
https://osmose.openstreetmap.fr/en/issue/f3e4887c-6a2d-1fa1-f7cd-597f64e1b435

```
source 	8423
item 	5050
class 	1
lat lon 	50.0614682 19.9379946 
```

![screen01](https://github.com/osm-fr/osmose-backend/assets/899988/78d29b0e-7b42-47c0-944f-ca1ccc2d57cf)

> A street name "sounds" similar to that of another street but is not spelled the same way.

triggered by https://github.com/Helium314/SCEE/issues/504#issuecomment-1872950553

matkoniecz:
also, I have no idea what and why is supposed to be wrong here

Famlam:
I must say I don't 100% understand how the code of the soundex test works, but to my knowledge it looks for names that "sound" similar. Thereby for instance detecting typo's like James Bond vs James Bont. However, this also means that it'll inherently have some false positives (like in Dutch Kennedylaan vs Kennedybaan: Kennedy lane vs Kennedy track). Human verification is very important with this one.

I don't know if it was supposed to be restricted to streets (in which case a `highway`-check is missing) or not (in which case the text should be updated)

frodrigo:
There is name in street but common with other objects and it is better to have more data for self comparing.

See also #813.

> also, I have no idea what and why is supposed to be wrong here

Yes the title is bad, it should say what is wrong, not how it works.

mnalis:
> also, I have no idea what and why is supposed to be wrong here

yes, it would be nice if the test mentioned from which element it got that other/similar name. Currently it is guessing work (and it may well be that it is the _other_ element that needs editing, not the highlighted one).

> Yes the title is bad, it should say what is wrong, not how it works.

Perhaps title=_"Possibly misspelled name"_ would fit better then _"Soundex test"_ ?

> There is name in street but common with other objects and it is better to have more data for self comparing.

So if that is intentional, perhaps detail=_'A street name "sounds" similar to that of another street<strong>/place</strong> but is not spelled the same way.'_ would be less confusing then? (bold text added).

Famlam:
The improved texts are now live

PR #2305, opened 2024-08-09
Update texts of soundex test

Famlam:
#2116

- Clarify its not for streets (only) but for everything, e.g. in NL there's a `leisure=sports_centre` and a `healthcare=clinic` detected too
- Improve the title (used the suggested one by mnalis)
- Add some extra clarification text (my personal wishlist, sometimes the change is so small, e.g. i vs í, that it looks like it's proposing to add the same...)
- Fix a typo


frodrigo:
Nice, thank you.

PR #2309, opened 2024-08-13
Add ref:Aléop exception

waterced:
Corresponding to the public transportation system of Pays-de-la-Loire region
https://aleop.paysdelaloire.fr/


frodrigo:
Better to fix this key to match the recommended rules. The key should be in ASCII.

frodrigo:
I close it. Ask for reopen if required.

PR #2313, opened 2024-08-17
Enable more tags to trigger construction warnings

Famlam:
Instead of enabling only `highway|landuse|building=construction` (+ everything with `construction:*` and `construction=*`), enable it for all tags that have `*=construction` (unless explicitly blacklisted), for example railways.

Also adds `operational_status=(under_)construction`. An uncommon key, so I'm also happy to ignore it if preferred.

Additionally, fix a typo

(Note my actual use case didn't work yet due to an issue regarding timestamps, but I thought it could still be worth to send in the PR already)

### DIFF
--- a/plugins/Construction.py
+++ b/plugins/Construction.py
@@ -38,7 +38,10 @@ class Construction(Plugin):
 been in construction for more than two years or opening data is
 exceeded.'''))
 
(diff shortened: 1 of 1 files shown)


frodrigo:
Nice, thank you.

Issue #2314, opened 2024-08-18
Timestamp of data not provided to plugins


Famlam:
### Summary
The plugin [Construction](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/Construction.py) uses the timestamp of OSM data. Timestamp data from the pbf file is however not provided to the plugins.

### More in detail
_(Summarized from external discussion)_
- [Way 958176876](https://www.openstreetmap.org/way/958176876) doesn't get flagged by [plugins/Construction.py](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/Construction.py) (item = class = 4070) despite being >3 years untouched as `railway=construction`
- The plugin [uses the `data["timestamp"]` property](https://github.com/osm-fr/osmose-backend/blob/dev/plugins/Construction.py#L93) to determine if a construction is likely completed. However, the only keys in `data` for that way are `['id', 'tag', 'nd']`
- The pbf file (spain_andalucia_malaga from OSMFR) does contain the timestamp property: 
```
famlam@osm110:~$ osmium getid -r -f xml /data/work/famlam/extracts/spain_andalucia_malaga.osm.pbf w958176876
<?xml version='1.0' encoding='UTF-8'?>
<osm version="0.6" generator="osmium/1.13.1">
[======================================================================] 100%
  <node id="8865772378" version="2" timestamp="2023-08-21T21:47:03Z" uid="6493119" user="JAR2004" changeset="140201770" lat="37.0292581" lon="-4.5597221">
    <tag k="railway" v="buffer_stop"/>
  </node>
  <node id="8865772379" version="1" timestamp="2021-06-25T11:45:17Z" uid="13517460" user="ALF_00" changeset="106959804" lat="37.0292166" lon="-4.5563174"/>
  <way id="958176876" version="1" timestamp="2021-06-25T11:45:17Z" uid="13517460" user="ALF_00" changeset="106959804">
    <nd ref="8865772378"/>
    <nd ref="8865772379"/>
    <tag k="railway" v="construction"/>
  </way>
</osm>
```
- @jocelynj proposed a possible cause: "if i’m not mistaken, we had an issue with the pbf reader, that doesn’t output timestamp in its default version". "it looks like that timestamp is available in [modules/osm_pbf_parser/osm_pbf_parser.cc](https://github.com/osm-fr/osmose-backend/blob/dev/modules/osm_pbf_parser/osm_pbf_parser.cc#L93-L99), but not propagated to python code"
[...]

PR #2315, opened 2024-08-18
Fix timestamp check

Famlam:
1. Propagate timestamp to python (as int)
2. Convert Construction.py to use int (seconds from epoch) rather than string dates
3. Add a log warning to ensure a change to the pbf reader won't go unnoticed again

#2314

### DIFF
--- a/modules/OsmPbf_libosmbf.py
+++ b/modules/OsmPbf_libosmbf.py
@@ -85,35 +85,35 @@ class OsmPbfReader(OsmReader, osm_pbf_parser.Visitor):
         osm_pbf_parser.read_osm_pbf(self._pbf_file, self)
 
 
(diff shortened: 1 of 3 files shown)


Famlam:
The error to modules/Polygon.py seems unrelated to this PR

frodrigo:
I pick only the first commit on dev branch. Than ask all backend maintenair to rebuild the pbf parser.

Famlam:
I'm afraid that the current partial merge of my PR will crash when deployed: the old plugin cannot handle numeric timestamps, which means that python will try to evaluate `end_date = datetime.datetime.strptime(data["timestamp"][0:10], "%Y-%m-%d") + self.date_limit` with a number as `data["timestamp"]`
I think it's better to merge the full PR and comment out the line that prints the warning temporarily?

PR #2319, opened 2024-08-24
Implement method for special country codes

### LINKED ISSUE #2318: Support `FX` country code
In [this modification](https://josm.openstreetmap.de/wiki/Rules/FranceSpecificRules?action=diff&version=94&old_version=93) the France-specific mapcss rules switched some rules from `[inside("FR")]` (all of France) to `[inside("FX")]` (only mainland France + nearby islands)

This is valid per https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation#Territoryselector
Also defined here: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Exceptional_reservations

However, when updated those rules unfortunately won't match anymore within Osmose as FX is not a known country code in the configuration. (Nor is there a 'dictionary' of country code aliases in the mapcss parser)

Famlam:
- Implement method to convert special country codes to Osmose-codes
  - `FX` is already implemented
  - Others from [the special ones](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#exceptional-reservations) will trigger a console warning, to be implemented when needed
- Add test cases

Fixes #2318

### DIFF
--- a/mapcss/mapcss2osmose.py
+++ b/mapcss/mapcss2osmose.py
@@ -344,6 +344,34 @@ def functionExpression_rule_flags(t, c):
         c['flags'].append('relational')
     return t
 
+specialCountryMap = {
+    # https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#exceptional-reservations
+    # None here simply means 'not yet implemented by us'
+    'AC': None,
+    'CP': None,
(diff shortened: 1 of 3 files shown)


frodrigo:
Thank you.

PR #2324, opened 2024-09-01
Add `charge`-related mapcss rules

Famlam:
I noticed that the `charge`/`charge:conditional` key very frequently doesn't follow the right structure as defined on the wiki so I made a mapcss rule set for it.
https://josm.openstreetmap.de/wiki/Rules/ChargeRules

Gives the following results:
- bad time unit: 754x
- no currency specified (like `charge=5`): 2371x
- use of symbols instead of the ISO code (`charge=€5`): 8020x
- invalid currency codes (`charge=5 dollars`): 2524x
- `,` instead of `.`: 1054x
- other issues: 5327x

---

P.s. I could also add this as part of item 3091 (numerical tag issues) if desired.

### DIFF
--- a/mapcss/item_map.py
+++ b/mapcss/item_map.py
@@ -1,6 +1,11 @@
 #-*- coding: utf-8 -*-
 item_map = \
-{'DutchSpecific': {'class': {None: 0,
+{'Charge': {'class': {'Invalid value of charge': 30916},
+            'prefix': 'Josm_',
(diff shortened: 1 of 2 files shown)


frodrigo:
> P.s. I could also add this as part of item 3091 (numerical tag issues) if desired.

Yes, please.

Famlam:
Done :)

PR #2325, opened 2024-09-01
Fix #2323

### LINKED ISSUE #2323: Google Open Buildings / google_africa_buildings is a valid source
item | 3020
class | 706

https://www.openstreetmap.org/way/1069265880
source=esri/Google_Africa_Buildings
or
https://www.openstreetmap.org/way/1069265603
source=esri/google_africa_buildings

but [Google Open Buildings has the ODbL as one of its licences](https://sites.research.google/open-buildings/), which makes it compatible with osm

Famlam:
See #2323

Allow Google Open Buildings which is ODbL
Also remove some unnecessary unicode-strings

### DIFF
--- a/plugins/Source.py
+++ b/plugins/Source.py
@@ -51,14 +51,14 @@ the origin.'''),
 `source=*`.'''))
 
     def check(self, tags):
-        if u"source" not in tags and u"ref" not in tags:
+        if "source" not in tags and "ref" not in tags:
             return
 
-        for tag in (u"source", u"ref"):
+        for tag in ("source", "ref"):
             if tag in tags:
                 value = tags[tag].lower()
-                if u"google" in value:
-                    return {"class": 706, "subclass": 2, "text": {"en": u"Google {0}={1}".format(tag, value)}}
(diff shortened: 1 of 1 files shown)


plugins/Source.py
```diff
@@ -51,14 +51,14 @@ def init(self, logger):
...
+        if "source" not in tags and "ref" not in tags:
             return
 
-        for tag in (u"source", u"ref"):
+        for tag in ("source", "ref"):
             if tag in tags:
                 value = tags[tag].lower()
-                if u"google" in value:
-                    return {"class": 706, "subclass": 2, "text": {"en": u"Google {0}={1}".format(tag, value)}}
+                if any("google" in v and not v.endswith("buildings") for v in value.split(';')):
```
frodrigo, line 60: Maybe add a `.strip()`?

Famlam: Sure, done :) .
According to TagInfo there's no cases of `source=google_africa_buildings ; something`, thus with a space between `buildings` and `;` but it's more future proof indeed.

PR #2328, opened 2024-09-09
Include way tags in errors

ComradeRamen:
Not sure if this is all that's needed.


ComradeRamen:
I use way tags to sort errors offline after downloading them.

PR #2335, opened 2024-09-14
[minor update] Following an official name change: Curtit→Del Fabbro

AntoninDelFabbro:


Famlam:
Isn't this automaticaly updated once you translate again with your new name using the same account? (And then synchronized)

frodrigo:
That complicates things. The data in the .po files are synchronized from Transifex. So editing it here on Github will have no effect: it'll be restored to the current state every time we synchronize the translations. The change should be done on the Transifex side with the contributor account. And even here, I'm not sure you can change old contrib.

PR #2350, opened 2024-10-06
charging_station_FR: add support for output power tags

guiohm:
This will use the newly aggregated data from https://github.com/Jungle-Bus/ref-EU-EVSE/pull/15.

We may want to also edit the `trap` text:
The data published is the output of one point of delivery which may have multiple sockets, so it is assumed that a charger able to output 100 kW on a combo-CCS socket will be able to "max out" a Type 2 socket rated at 43 kW (in my experience that's the case). In this scenario the `socket:type2:output` is "deduced" and we may inform the users to check it.

There's also a related conversation on this topic with @nlehuby here > https://github.com/Jungle-Bus/ref-EU-EVSE/pull/14


frodrigo:
I close. Please reopen if required.

Issue #2345, opened 2024-09-24
TagWatchFrViPofm.py doesn't manage entry with template


Marc-marc-marc:
item=3030
little by little, some templates are added and the pipe causes the regex to fail, for ex 
```
|altitude||{{tag|ele}} || tag ||
```
related code https://github.com/osm-fr/osmose-backend/blob/master/plugins/TagWatchFrViPofm.py#L55

Marc-marc-marc:
improved regex
`reline = re.compile(r"^\|\s*((?:[^|{}]*\s*|\s*\{\{.*?\}\})\s*\|\|\s*){3}((?:[^|{}]*\s*|\s*\{\{.*?\}\})\s*).*")
`

Famlam:
Oof, now that all previously skipped/unparsable entries are detected again, the (worldwide) results exploded:
https://osmose.openstreetmap.fr/nl/issues/graph.png?item=3030&start_date=2024

But it seems to be ok. If certain entries are too spammy, the wiki can be adjusted.

PR #2352, opened 2024-10-10
TagWatchFrViPofm.py : improve regexp

Marc-marc-marc:
related to #2345

### DIFF
--- a/plugins/TagWatchFrViPofm.py
+++ b/plugins/TagWatchFrViPofm.py
@@ -52,7 +52,7 @@ class TagWatchFrViPofm(Plugin):
         self._update_ks_vr = defaultdict(dict)
         self._update_kr_vr = defaultdict(dict)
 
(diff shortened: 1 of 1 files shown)


Famlam:
I think that, using the example you provided: `|altitude||{{tag|ele}} || tag ||`
The capture groups would only be [`"tag ||"`, None]
Instead of: [`"altitude"`, `"{{tag|ele}}"`, `"tag"`, None]

Perhaps you can add a test case as well?

Famlam:
The fix in #2358 should also resolve this issue and all other issues listed in #2345
I hope you don't mind that this PR would become obsolete in that case?

Famlam:
Implemented via #2358

Marc-marc-marc:
> I hope you don't mind that this PR would become obsolete in that case?

that's fine, thanks

Issue #2357, opened 2024-10-11
Tree vs wiki plugin is (also) broken / universal wiki parser solution? 


Famlam:
At this moment the tree vs wiki check is broken because:
1. ~the table is a mix of multi- and single-line entries (nevermind, probably this still works)~
2. Someone added the wikidata entries as link (thus adding a `|`)

Both break the parser. (Although it'll work for a few more days due to the cached version) 
cc @ivanbranco, see also #2224

This raises two questions for me:
1. What's the future of this table (because it's said to be intended for QA tools), 
2. Since we have a lot of wikidata parsers in various plugins (and corresponding issues every time someone adds a template, hyperlink, switches multi/single-lines..., is there a python package which can do this for us instead?
If not, I'll probably make a module with some universal functions like `remove_templates()` and `link_to_displaylink()`

Similar issues with tables: #2345 #2351

ivanbranco:
That page was originally created by me with QA tools in mind, so if you confirm the Wikidata links are causing issues, I'd suggest reverting it to the previous state. It might also be helpful to add a note on the page, encouraging discussion before making significant changes to the table, especially since the table is actively used by at least one QA tool (this one).

Famlam:
I found two libraries to parse wiki tables. I'll give that a try first :)

Famlam:
The new wiki table parser seems to work, results are back :)

Famlam:
@ivanbranco As a heads-up: someone again modified the wiki by adding a new column so everything is broken again, results will disappear as soon as the cache expires :(
I've added a warning to the wiki... and will look into fixing the code again.

Issue #2351, opened 2024-10-09
Incorrect error message Invalid postcode


woodyBu:
Hi,

the error messages “Invalid street level postcode 53844 for country code DE” are incorrect
https://osmose.openstreetmap.fr/de/map/#level=1%2C2%2C3&loc=17/50.77821/7.100168&item=3190
the zip codes are all correct here.

woodyBu

frodrigo:
Date come from 
https://en.wikipedia.org/wiki/List_of_postal_codes

Maybe we have a parsing issue with historic data for DE.

Famlam:
Could it be related to the change to the wiki on Oct 7?
It _should_ be fine for Germany as the table is read top-to-bottom (see #2202) and the newer entries overwrite the older?

Famlam:
Solved :)

PR #2358, opened 2024-10-13
Implement universal Wiki reader

Famlam:
Implemented an initial wiki-reader based on the `wikitextparser` package and required functions
Adds support for reading tables, lists and templates, and adds a function to convert `{{Tag|x|y}}` to `x=y`

Plugins to convert (see #2345 #2351 #2357)
-  TagFix_Tree (table-based)
-  TagWatchFrViPofm (table-based)
-  TagFix_Postcode (table-based)
-  ~TagFix_Tree_Lang_fr (list-based)~ skipped
(The benefits are small at the moment, as it still requires a lot of post-parsing, making this PR unnecessary big. Also see 2361)
-  TagFix_Deprecated (template-based)

### DIFF
--- a/plugins/modules/wikiReader.py
+++ b/plugins/modules/wikiReader.py
@@ -23,19 +23,20 @@
 # This module file contains functions to read MediaWiki markup tables, templates, lists, ...
 
 import wikitextparser
+from typing import Union, Optional
 
 # Get a list of lists containing all cells of a table.
 # Parameters:
-#   wikitext (str) - the text of a wikipedia page
-#   tab_index (int) - the index of the table (if there's multiple tables on the wiki)
-#   keep_markup (bool) - if False, everything (except Templates) will be converted to plain text
-#   skip_headers (bool) - if True, header rows are removed. Assumes all headers are on top
+#   wikitext - the text of a wikipedia page
+#   tab_index - the index of the table (if there's multiple tables on the wiki)
+#   keep_markup - if False, everything (except Templates) will be converted to plain text
+#   skip_headers - if True, header rows are removed. Assumes all headers are on top
 # Returns:
 #   The cell contents, specified as a list in a list.
 #   The outer list is the rows, the inner list are the cells in that row
 # Throws:
 #   If the table at the specified index isn't found
-def read_wiki_table(wikitext, tab_index = 0, keep_markup = False, skip_headers = True):
+def read_wiki_table(wikitext: str, tab_index: int = 0, keep_markup: bool = False, skip_headers: bool = True) -> list[list[Optional[str]]]:
     # Drops all markup, such as italics, hyperlinks, ...
     if not keep_markup:
         wikitext = wikitextparser.remove_markup(wikitext, replace_tables=False, replace_templates=False)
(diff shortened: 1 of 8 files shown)


plugins/modules/wikiReader.py
```diff
@@ -0,0 +1,138 @@
...
+# See read_wiki_list for details (excluding list_index)
+def read_all_wiki_lists(wikitext, keep_markup = False, include_sublists = False):
+    res = []
+    if not keep_markup:
+        wikitext = wikitextparser.remove_markup(wikitext, replace_templates=False)
+
+    try:
+        list_index = 0
+        while True:
+            res = res + read_wiki_list(wikitext, list_index=list_index, keep_markup=True, include_sublists=include_sublists)
```
frodrigo, line 114: res.append() ?

plugins/modules/wikiReader.py
```diff
@@ -0,0 +1,138 @@
...
+    if not keep_markup:
+        wikitext = wikitextparser.remove_markup(wikitext, replace_templates=False)
+
+    # Get all templates that match the filter
+    template_objects = list(filter(lambda t: t.name.strip().lower() in template_name, wikitextparser.parse(wikitext).templates))
+
+    ret = []
+    for t in template_objects:
+        ret.append([t.string, t.name.strip()] + [str(a)[1:].strip() for a in t.arguments])
+    return ret
```
frodrigo, line 79: Why not use a map like in other functions ?

plugins/modules/wikiReader.py
```diff
@@ -0,0 +1,138 @@
...
+#   wikitext (str) - the text of a wikipedia page
+#   tab_index (int) - the index of the table (if there's multiple tables on the wiki)
+#   keep_markup (bool) - if False, everything (except Templates) will be converted to plain text
+#   skip_headers (bool) - if True, header rows are removed. Assumes all headers are on top
+# Returns:
+#   The cell contents, specified as a list in a list.
+#   The outer list is the rows, the inner list are the cells in that row
+# Throws:
+#   If the table at the specified index isn't found
+def read_wiki_table(wikitext, tab_index = 0, keep_markup = False, skip_headers = True):
```
frodrigo, line 38: As it is a in modules. Maybe type the function rather than explaining the typing.

Famlam: Done. It seems that typing is only checked by mypy if the calling function is also typed though

plugins/modules/wikiReader.py
```diff
@@ -0,0 +1,136 @@
...
+#   wikitext - the text of a wikipedia page
+#   tab_index - the index of the table (if there's multiple tables on the wiki)
+#   keep_markup - if False, everything (except Templates) will be converted to plain text
+#   skip_headers - if True, header rows are removed. Assumes all headers are on top
+# Returns:
+#   The cell contents, specified as a list in a list.
+#   The outer list is the rows, the inner list are the cells in that row
+# Throws:
+#   If the table at the specified index isn't found
+def read_wiki_table(wikitext: str, tab_index: int = 0, keep_markup: bool = False, skip_headers: bool = True) -> list[list[Union[str | None]]]:
```
frodrigo, line 39:
```suggestion
def read_wiki_table(wikitext: str, tab_index: int = 0, keep_markup: bool = False, skip_headers: bool = True) -> list[list[Optional[str]]]:
```

plugins/modules/wikiReader.py
frodrigo: > Done. It seems that typing is only checked by mypy if the calling function is also typed though

Thank. Yes, but you can also force to check it.

Famlam: You mean with `check_untyped_defs = True`?
Because that gives a lot of errors `Found 928 errors in 183 files (checked 432 source files)`

frodrigo:
Thank you !

Issue #2341, opened 2024-09-20
Different values of phone and of contact:phone (only spaces differ)


Marc-marc-marc:
item 3092
class 3097
contact:phone=+33 5 56 22 74 03
phone=+33556227403
it's the same value exempt space, it could be usefull to avoit detecting space change

Famlam:
I can see your point here that this warning doesn't benefit much, since it can be interpret without issues in either format and it doesn't point to an actual conflict. Changing the rule to the following would fix this issue.
```css
*[phone][contact:phone][replace(tag("contact:phone"), " ", "")!=replace(tag("phone"), " ", "")],
*[fax][contact:fax][replace(tag("contact:fax"), " ", "")!=replace(tag("fax"), " ", "")],
*[email][contact:email][contact:email!=*email],
*[website][contact:website][contact:website!=*website] {
    group: tr("Different value of tag contact:* and *");
    throwWarning: tr("Different values of {0} and of {1}", "{0.key}", "{1.key}");
    -osmoseItemClassLevel: "3092/3097/2";

    assertMatch: "node phone=1 contact:phone=2";
    assertNoMatch: "node website=1 contact:website=1";
    assertNoMatch: "node phone=\"+31 123 456 789\" contact:phone=\"+31123456789\""
}
```

I didn't do it as a PR, perhaps some people might see issues with "differently-spaced phone numbers" asides from just looking nice

Famlam:
The fix is deployed, spaces and dashes are now ignored. I can't find the original case from the first post on OSM so I assume that object was modified

PR #2365, opened 2024-10-19
Fix 2341: phone vs contact:phone with only spaces/dashes as difference

Famlam:
See #2341

Disable the warning of different `phone` and `contact:phone` numbers if the only difference are spaces.
Although the ideal case is that the phone numbers are formatted, I don't think this has any benefit other than just 'looking nicer'. So if both are specified, once with and once without spaces, this shouldn't raise a warning that they're different phone numbers.

Also adds three very popular keys that also frequently occur as `contact:X` or as `X`

(The error in the tests is unrelated, that's fixed by the PR in 2358 )

### DIFF
--- a/plugins/Phone2.py
+++ b/plugins/Phone2.py
@@ -23,19 +23,34 @@ class Phone2(PluginMapCSS):
         err = []
 
 
-        # *[phone][contact:phone][contact:phone!=*phone]
-        # *[fax][contact:fax][contact:fax!=*fax]
+        # *[phone][contact:phone][replace(replace(tag("contact:phone"),"-","")," ","")!=replace(replace(tag("phone"),"-","")," ","")]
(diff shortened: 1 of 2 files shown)


frodrigo:
Maybe also remove "-" and "()" ?

Famlam:
It's unfortunate that mapcss doesn't have a regexp_replace function (or similar), so for checking for equality I'd need something very ugly like `replace(replace(replace(replace(tag("X"), " ", ""), "-", ""), "(", ""), ")", "")`.

I'll do it for `-`, good suggestion! For `()` I don't think we should do this since the common notation +31 (0) 6 12345678 isn't the same as +31 0 6 12345678. The first means +31612345678 outside NL or 0612345678 when calling from inside NL, the second is invalid.

PR #2369, opened 2024-10-23
remove motorway and trunk exclusions from almost junction check.

ComradeRamen:
I'm not sure why these were initially excluded and would be helpful to find accidentally disconnected roads. 
If there's a different check I should use or another reason for the exclusion let me know.


frodrigo:
Yes. Let's try.

Issue #2368, opened 2024-10-23
Name with uppercase on highway=bus_stop


Gafigglethorp:
In Greece, the community consensus is that bus stop names are rendered in all caps.

This is the standard way to render them in all user facing applications: Transit maps, [the stop itself](https://thesstrans.com/wp-content/uploads/2022/02/IMG_1327-scaled-e1644251158437.jpg), bus route diagrams, digital maps inside the buses themselves or [online](https://telematics.oasth.gr), [transit applications](https://www.oasa.gr/%CE%B5%CE%BE%CF%85%CF%80%CE%B7%CF%81%CE%AD%CF%84%CE%B7%CF%83%CE%B7-%CE%B5%CF%80%CE%B9%CE%B2%CE%B1%CF%84%CF%8E%CE%BD/%CE%B5%CF%81%CE%B3%CE%B1%CE%BB%CE%B5%CE%AF%CE%B1/%CF%84%CE%B7%CE%BB%CE%B5%CE%BC%CE%B1%CF%84%CE%B9%CE%BA%CE%AE-app/), and more.

It is also documented in the wiki [here](https://wiki.openstreetmap.org/wiki/Greece/Public_Transport).

This has caused literally (tens of?) thousands of false positives across Greece.

frodrigo:
https://osmose.openstreetmap.fr/el/issue/079f2f57-5b21-4d9d-38c1-0841344aa21b

source 	7642
item 	5010
class 	803

Famlam:
Fixed :)

PR #2370, opened 2024-10-27
Whitelist uppercase PT infrastructure in Greece

Famlam:
See #2368


plugins/Name_UpperCase.py
```diff
@@ -46,21 +46,26 @@ def init(self, logger):
...
+                                       whitelist_from_nsi(self.country.lower())))
             self.whitelist.update(nsi_whitelist)
         else:
             self.whitelist = set()
 
     def node(self, data, tags):
         err = []
-        if u"name" in tags:
+        if "name" in tags:
+            # Whitelist bus stops in Greece, see 2368
```
frodrigo, line 63:
```suggestion
            # Whitelist bus stops in Greece, see #2368
```

Famlam: Thanks!

PR #2386, opened 2024-11-10
Make NSI country filter uniform

Famlam:
- Let all plugins use the same function, `nsi_rule_applies`, to evaluate the locationSet of NSI
- Allow country codes with subcodes (like "NL-GE") instead of only the country ("NL")

This doesn't fix 2358 completely yet, but makes a fix easier. 
Also, for cases where the geojson file has the same code as the Osmose extract (such as Iowa, Quebec, ...) this makes it work again.

For canada it also fixes some cases that never worked, e.g. [this entry](https://nsi.guide/index.html?t=operators&k=amenity&v=post_depot&tt=canada%20post&cc=ca) previously wasn't parsed well due to including `ca` but excluding `ca-qc`. We previously didn't support such "subregion-exclusions".

### DIFF
--- a/plugins/Name_Script.py
+++ b/plugins/Name_Script.py
@@ -120,7 +120,7 @@ appropriate.'''),
 
         self.whitelist_names = set()
         if country:
-            self.whitelist_names = whitelist_from_nsi(country[:2].lower())
+            self.whitelist_names = whitelist_from_nsi(country)
 
     def node(self, data, tags):
         err = []
--- a/plugins/Name_UpperCase.py
+++ b/plugins/Name_UpperCase.py
@@ -49,10 +49,10 @@ class Name_UpperCase(Plugin):
         self.country = None
 
         if "country" in self.father.config.options:
-            self.country = self.father.config.options.get("country")[:2]
-            self.whitelist = set(UpperCase_WhiteList.get(self.country, []))
+            self.country = self.father.config.options.get("country")
+            self.whitelist = set(UpperCase_WhiteList.get(self.country.split("-")[0], []))
             nsi_whitelist = set(filter(lambda name: self.UpperTitleCase.match(name) and not self.RomanNumber.match(name),
-                                       whitelist_from_nsi(self.country.lower())))
+                                       whitelist_from_nsi(self.country)))
             self.whitelist.update(nsi_whitelist)
         else:
             self.whitelist = set()
@@ -61,7 +61,7 @@ class Name_UpperCase(Plugin):
         err = []
         if "name" in tags:
             # Whitelist bus stops in Greece, see #2368
-            if self.country and self.country == "GR" and "public_transport" in tags and tags["public_transport"] in ("stop_position", "platform", "station"):
+            if self.country and self.country.split("-")[0] == "GR" and "public_transport" in tags and tags["public_transport"] in ("stop_position", "platform", "station"):
                 return err
 
             # first check if the name *might* match
--- a/plugins/TagFix_Brand.py
+++ b/plugins/TagFix_Brand.py
@@ -22,7 +22,7 @@
 from modules.OsmoseTranslation import T_
 from plugins.Plugin import TestPluginCommon
 from plugins.Plugin import Plugin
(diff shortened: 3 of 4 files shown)


frodrigo:
Sub country code was not done before because of not matching Osmose definition.

Famlam:
You mean that the same country code (like `US-IA` for Iowa) has different borders in Osmose or NSI? That would be quite surprising to me. Would you happen to know where (which region) this causes issues?

Currently, by not supporting it we also have false positives, e.g. [this detection](https://osmose.openstreetmap.fr/nl/issue/b47ccea1-19c5-9f8e-3475-794b11a46b42) shouldn't exist because [it's rule](https://nsi.guide/index.html?t=operators&k=amenity&v=post_box&tt=canada%20post) is excluding CA-QC. However, since we (without this patch) don't check for the `ca-qc` it's still detected.
(And also false negatives, e.g. [this node](https://www.openstreetmap.org/node/6773120886) isn't detected by [this rule](https://nsi.guide/index.html?t=brands&k=amenity&v=fast_food&tt=poulet), without this patch)

Of course this would only affect cases where our division is the same (US-IA) or smaller (CA-QC-*) than theirs, and the naming scheme is similar. For France for instance we use FR-[number] rather than FR-[letters]-[number], so that would really require a map and/or third party library. On the other hand, if they use a smaller subdivision than Osmose, it's unfixable at the moment (unless we start comparing node coordinates rather than extracts). However, those cases wouldn't be affected by this patch, would they?

frodrigo:
> You mean that the same country code (like US-IA for Iowa) has different borders in Osmose or NSI? That would be quite surprising to me. Would you happen to know where (which region) this causes issues?

Not the boundaries, but the code itself.

The issue is about making the map of the codes.

Maybe we can change some code on Osmose side, to match national or supra national code (if it can help to match NSI code).

Famlam:
I don't understand. 
If the codes match (e.g. CA-QC, US-IA) this will work to add support. 
If the codes do not match (e.g. FR) this PR will not change anything, right?

frodrigo:
I reread the code. Today the country is always the 2 chars country code, even for sub country area. If I'm not wrong, when a rule exclude a subpart, the 2-chars only country code make the rule excluded, doesn't ?

Concerning France sub areas, It looks like to me subarea codes are not used, but always geojson.

Famlam:
I'm not sure if I understand what you mean.
1. For France, nothing will change by this PR because of the different divisions (FR-[number] in Osmose vs FR-[lettercode] in NSI)
Maybe in a follow-up PR, either via a hardcoded map (where possible) or via the library (see issue). The library would be favorable for maintenance, but would be a big PR since it would need to download all geojson files etc. Hence I didn't do this yet.
2. For many other countries with subdivisions (AT, AU, BR, CA, CH, (part of) DE, ES, IT, MX, (part of) PH, (part of) US) it will make the subdivisions work. If I use **CA-QC** as example:
a) [this rule](https://nsi.guide/index.html?t=operators&k=amenity&v=post_box&tt=canada%20post) has `{"include":["ca"],"exclude":["ca-qc.geojson"]}`
b) What this PR does is that it strips away the `.geojson` (and normalizes the case if needed). For many countries (see list above and in the issue) this will work since the file is just called [country-region-code].geojson
c) Then, since Osmose divides CA-QC in subdivisions, example "CA-QC-CQC", it first checks `CA-QC-CQC` -> no match. So it checks `CA-QC` and finds this in de `exclude`. Hence it detects that this rule is excluded for CA-CQ.

Without this PR, it would only check for `CA`. It finds `CA` in the `include`, hence it (incorrectly) thinks this rule is included.

This PR doesn't do anything if the geojson isn't named the same as the Osmose subdivisions, e.g. France is unaffected, Denmark is unaffected, ... (see issue for full list)

frodrigo:
> I reread the code. Today the country is always the 2 chars country code, even for sub country area. If I'm not wrong, when a rule exclude a subpart, the 2-chars only country code make the rule excluded, doesn't ?

I was thinking the code was doing a smart thing with exclusion... but it does not.

Sorry, yes this PR will improve the results without introduce more false positive.

PR #2388, opened 2024-11-15
Cosmectic improvement on TagFix_Housenumber.py

FredericBourgeon:
Putting Number between `` characters will format it with the <code> tag. A simple cosmetic improvement that will hopefully break nothing.

### DIFF
--- a/plugins/TagFix_Housenumber.py
+++ b/plugins/TagFix_Housenumber.py
@@ -41,7 +41,7 @@ addr:inclusion=*.'''))
             title = T_('Invalid addr:interpolation or addr:inclusion value'),
             detail = T_(
 '''* Tag `addr:interpolation=*` is only valid with values: `even`, `odd`,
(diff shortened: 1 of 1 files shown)


Famlam:
I'm afraid this would suggest to people that they should put literally "Number" there, rather than 1, 2, 3 etc. So I wouldn't do this.

(Grammar-wise I think it should be "a number", but without these `-marks)

FredericBourgeon:
Giving it a second thought and agreeing with your comments, I'd suggest "or a number" rather than "and" which could suggest something such as "even3" as being valid.

Famlam:
Agreed, would you like to make this edit?

FredericBourgeon:
> Agreed, would you like to make this edit?

I hope I've done it the right way on the second time. All through Github web interface, which is quite nifty.

Famlam (approved):
Thank you, it looks correct to me :)

PR #2394, opened 2024-11-19
Add merge analyser for French power poles in Vendee

flacombe:
Vendée Numérique is a local authority in Vendée (France) who manage very high speed Internet broadband rollout.

They recently published power poles points used for fibre roll out that could be useful to complete power poles there.

This new analysis comes with a brand new gracethd2 abstract analyser as they still use Grace THD v2 instead of usual v3 which was already available.
As a consequence, current abstract gracethd analyser was moved to gracethd3 as to prevent any confusion.

### DIFF
--- a/analysers/Analyser_Merge_power_pole_FR_gracethd2.py
+++ b/analysers/Analyser_Merge_power_pole_FR_gracethd2.py
@@ -49,7 +49,7 @@ class Analyser_Merge_power_pole_FR_gracethd2 (Analyser_Merge_Point):
                     static2 = {'source': self.source},
                     mapping1 = {
                         'material': lambda res: self.extract_material.get(res['modele']),
-                        'operator': lambda res: extract_operator.get(res['gestionnai'],  extract_operator.get(res['proprietai'])) ,
+                        'operator': lambda res: extract_operator.get(res['gestionnai'],  extract_operator.get(res['proprietai'])),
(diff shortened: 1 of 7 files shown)


frodrigo:
Please, add it the configuration.

analysers/Analyser_Merge_power_pole_FR_gracethd2.py
```diff
@@ -0,0 +1,61 @@
...
+                select = Select(
+                    types = ['nodes'],
+                    tags = {'power': 'pole'}),
+                conflationDistance = conflationDistance,
+                mapping = Mapping(
+                    static1 = {'power': 'pole'},
+                    static2 = {'source': self.source},
+                    mapping1 = {
+                        'material': lambda res: self.extract_material.get(res['modele']),
+                        'operator': lambda res: extract_operator.get(res['gestionnai']) if res['gestionnai'] and extract_operator.get(res['gestionnai']) else extract_operator.get(res['proprietai']) if res['proprietai'] and extract_operator.get(res['proprietai']) else None,
```
frodrigo, line 52: Check if this works
```suggestion
                        'operator': lambda res: extract_operator.get(res['gestionnai'],  extract_operator.get(res['proprietai'])) ,
```

flacombe: Yes it works thanks. Also used in gracethd3 analyzers

PR #2398, opened 2024-11-25
Enable amenity=weighbridge on highway=toll_gantry

urbalazs:
Not only `highway=service` can have the tag `amenity=weighbridge`, but `highway=toll_gantry`, too. See [this street view image](https://www.google.hu/maps/@47.6535607,17.4166038,3a,64.3y,238.2h,78.66t/data=!3m7!1e1!3m5!1sRIZH9NcFGyUfsT2K3qMDlA!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fcb_client%3Dmaps_sv.tactile%26w%3D900%26h%3D600%26pitch%3D11.341690730724679%26panoid%3DRIZH9NcFGyUfsT2K3qMDlA%26yaw%3D238.20018444895564!7i16384!8i8192?entry=ttu&g_ep=EgoyMDI0MTExOS4yIKXMDSoASAFQAw%3D%3D) as an example.


Famlam:
It's a common tag combination in Hungary (weigh-in-motion systems), but I'm not sure if this is a toll system? It looks more like an enforcement situation to detect overweight vehicles, based on what I can find online?

p.s. you can ignore the failing test, that's unrelated

Yogurt4:
Well, a bit of history.

The Hungarian electronic toll system for HGVs started in 2013. It includes not only highways and motor roads but all primary roads, too. Owning to this, there are no entrance/exit gates (like in Germany). One has to pay toll for each section [typically a few kilometres], and for every use, in each direction. To enforce this, a hundred of fixed cameras were installed in addition to mobile control vehicles.

Skip forward a couple of years. Some enforcement cameras were upgraded with in-motion weighing, especially near border crossings to filter out overloaded trucks (typically from the Balkans). They use the same infrastructure, e.g., the existing electricity, mobile network, and the existing plate number recognition system.

The above StreetView photo shows the camera gantry, the speed and axle number detection loops in the foreground. The gray strips are the later added weight sensors.

The situation now: AFAIK all e-toll cameras were upgraded with in-motion weighing, along with a hundred new additional checkpoints. The country-wide weight enforcement system starts in this December.

urbalazs:
> I'm not sure if this is a toll system

Here is the [office website](http://tengelysulymeres.hu/tsm-system) of the National Axle Weight Measurement System (TSM). You can get localized PDF in many languages in the top of the page. From the [English version](http://tengelysulymeres.hu/documents/TSM_Kisokos_ENG.pdf):

> How Does It Work?
> The TSM system is built onto the existing elements of the toll enforcement system (UD system).

Famlam:
Ok on the PR.
However, can you please document this on the wiki somewhere and/or give a link to a forum discussion?
There probably was a discussion somewhere about whether this is surveillance or toll, two separate objects (the physical weighing part and the toll/enforcing part) or a single object I would assume?

Famlam:
> However, can you please document this on the wiki somewhere and/or give a link to a forum discussion?

@urbalazs @Yogurt4

Yogurt4:
> > However, can you please document this on the wiki somewhere and/or give a link to a forum discussion?

Hi. Exactly what do you want to be documented and on which Wiki?
And why would you need a forum discussion (we use Matrix, by the way) for a factual thing?

Famlam:
Just to make sure there's a community consensus on this tagging method. (Rather than for instance enforcement relations and/or 2 separate objects for the weighbridge and the camera part)

For example, the wiki of highway=toll_gantry also shows an exception for German toll systems:
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtoll_gantry
> _This should not be used in the case of the German toll enforcement gantries. For such cases, please refer to the [enforcement relation page](https://wiki.openstreetmap.org/wiki/Relation:enforcement#Tags). More info can be found on [...]_

Yogurt4:
I've added a paragraph to the highway=toll_gantry Wiki page.

> Just to make sure there's a community consensus on this tagging method. (Rather than for instance enforcement relations and/or 2 separate objects for the weighbridge and the camera part)

As for the enforcement relations, they are a bit tricky due to these Hungarian specialities. We have 3k+ toll sections and <200 fixed enforcement cameras. (There are also mobile units on cars.) For these cameras, they are part of the corresponding enforcement relation in "device" role. The remaining 3k sections are without a "device", triggering a warning from QA tools all the time.

Yes, there could've been a 2-device sub-relation connecting the gantry (cameras) and the weighbridge (road surface). Since they share the same hardware, everyone felt it's one device.

Famlam (approved):
Thank you

frodrigo:
Thank you all.

PR #2405, opened 2024-12-08
Fix some country codes

Famlam:
- Colorado used US-CA, but that's California
- Cambodia used the 3-letter code rather than the 2 letter code like all other countries
- Kiribati used KL instead of KI

Not sure if this requires any other changes on the server or whether this is a simple change?


frodrigo:
Nice to spot error in the config !

I think nothing more is required.

cc @jocelynj

jocelynj:
There should be no impact on the server, thanks for the change.

Issue #2408, opened 2024-12-10
False positive unmapped max speed limit in US


smichae:
This may be my lack of understanding about this validation, but many Mapillary-detected speed limits in my region seem to always be detected as unmapped in my region, but an adjacent highway way does have the maxspeed.  (This may also be a problem primarily where survey:date is added on the way - I have seen a few where survey:date is not on a way, and Osmose isn't flagging an issue). I have closed this particular detection (and a few on this same stretch of road) at least once and it has reappeared.

https://osmose.openstreetmap.fr/en/issue/3c2a70b7-759e-8358-c399-f6390c4ccef8

cycleway:both=no
highway=tertiary
lanes=3
lanes:backward=2
lanes:forward=1
lit=yes
mapillary=312881763836487
maxspeed=40 mph
name=South Morris Avenue
shoulder=no
sidewalk=right
surface=concrete
survey:date=2021-06-12
tiger:cfcc=A41
tiger:county=McLean, IL
tiger:name_base=Morris
tiger:name_direction_prefix=S
tiger:name_type=Ave
tiger:reviewed=no
tiger:zip_left=61704
tiger:zip_left_1=61704
tiger:zip_right=61704
tiger:zip_right_1=61704
turn:lanes:backward=left|through;right

https://www.openstreetmap.org/edit#map=19/40.449326/-89.005384

![image](https://github.com/user-attachments/assets/3abac42d-d06e-4b74-be72-094be3158cf2)

Famlam:
On a very inspection it seems Osmose is mixing up mph and km/h.
Just to be sure, the full state of Illinois uses mph units for all speed signs, correct?

smichae:
Correct, the whole state does use mph - I believe all other U.S. states do as well.

Famlam:
Solved :)

Issue #1716, opened 2022-12-28
Split USA-Virginia


frodrigo:
- State Police, 7 named slices  https://en.wikipedia.org/wiki/Virginia_State_Police

frodrigo:
```
richmond,1149986,1149988,1633329,206598,206600,2534164,2534169,2534170,2534172,2534175,2534176,2534179,2534182,2534186,2534187,2534194,2534196,2534198,2534200,2534202,3070666,3070728,3070795,3864712,8623880
culpeper,1149987,1633325,1633326,1633327,1633328,1633330,1633331,1633332,206857,207037,2534173,2534189,2534197,2534201,2534203,2534204
appomattox,206651,206666,2504948,2532613,2532619,2532627,2534165,2534166,2534168,2534171,2534174,2534177,2534180,2534188,2534191,2534193,2534199,3864617
wytheville,1913188,2532616,2532618,2532620,2532622,2532626,2532632,2532635,2532636,2532637,2532638,2532639,2532640,2532641,3864677,3864683,7508928
chesapeake,1477365,206528,206670,206671,206672,206673,206807,207036,2534167,2534178,2534181,2534184,2534185,2534190,2534192,2534195,2534205,2534206,2534207,3864687,3864703,3864704,3864705,3864706
salem,node/317004614,206561,206856,206868,2532612,2532614,2532615,2532617,2532621,2532623,2532624,2532625,2532628,2532629,2532630,2532631,2532633,2532634,2534183,3864647,3864686,5884638,7495916,9260621
fairfax,1149984,206609,206637,206642,206870,206874,944948,945043,962190
```

jocelynj:
Diffs are now generated. Note that I've removed "node/317004614" from salem line.

Issue #1712, opened 2022-12-27
Split USA-Colorado


frodrigo:
Need a layout
* State patrol, 6 unnamed slices, https://en.wikipedia.org/wiki/Colorado_State_Patrol#/media/File:CO_-_CSP_Districts.jpg
* Trasportation, 5 slices, https://www.codot.gov/topcontent/assets/cdotregionmap.pdf

frodrigo:
```
1,1411327,1411333,1411337,1411338,1411339,1411346,1411351,442310
2,1411317,1411320,1411328,1411331,1411335,1411343,1411347,1411349,415767,415769,415781,415785,417649,439376,
3,1411315,1411329,1411332,1411340,1411341,1411352,1411353,1411354,415167,415168,416905,416907,421142,441008,443690
4,1411314,1411318,1411319,1411322,1411323,1411324,1411326,1411330,1411334,1411336,1411337,1411344,1411348
5,1411313,1411316,1411325,1411342,1411345,1411350,1411355,416515,416516,416521,416522,416906,421120,443692
```

jocelynj:
Diffs are now generated on http://download.openstreetmap.fr/replication/north-america/us-west/colorado/.

Issue #1711, opened 2022-12-27
Split USA-Michigan


frodrigo:
Need a layout
* Police, 8 slices, https://www.michigan.gov/-/media/Project/Websites/msp/EMHSD/documents/EMD_Map2_51205.pdf?rev=59b37c6f00264fe8b4132af064c8c89e

frodrigo:
```
[out:xml];
relation(165789);
map_to_area -> .a;
relation(area.a)[admin_level=6];
out body;
>;
out skel qt;
```

```
cat mi.geojson | jq -r "[.features[].properties | select(.r)] | unique | group_by(.r) | map({name: .[0].r, ids: [.[] | .\"@id\"] | @csv }) | .[] | [.name, .ids] | @csv" | tr -d '"' | sed -e "s_relation/__g"
```

```
1,1582241,1582242,1801749,1801750,1898721,1898723,1907314,359765,582246,901500,901501
2,359761,359762,359766,359767,407497,574424
3,1907303,1907305,1907324,1907331,1907335,359763,359764,571344,572130,572624,572689,574448,574462,582247,611155
5,1801747,1801748,1801751,1907304,1907307,1907309,1907317,1907340,407519
6,1907312,1907315,1907316,1907319,1907320,1907322,1907323,1907326,1907328,1907329,1907330,1907333,1907337
7,1586055,1586056,1907305,1907306,1907308,1907310,1907311,1907313,1907318,1907321,1907325,1907327,1907334,1907336,1907338,1907339,1907341
8,1798737,1798740,1798741,1798743,1798744,1798745,1898716,1898717,1898718,1898719,1898720,1898722,912579
```

jocelynj:
Could we name the regions by looking at these regions? https://en.wikipedia.org/wiki/List_of_regions_of_the_United_States#Michigan

1 - southeast
2 - detroit_metro
3 - central
5 - southwest
6 - west
7 - northern_lower
8 - upper_peninsula

Or we could use these named regions: https://www.michigan.gov/mdot/about/regions

frodrigo:
```
1 - southeast
2 - detroit_metro
3 - central
5 - southwest
6 - west
7 - northern_lower
8 - upper_peninsula
```
I think this is ok

jocelynj:
Diff are generated with the following names:
```
southeast 1582241,1582242,1801749,1801750,1898721,1898723,1907314,359765,582246,901500,901501
[...]

PR #2410, opened 2024-12-11
Add missing speed_limit_unit='mph' for several US states

Famlam:
They seem to have disappeared in:
e21fa09 (CA)
#1611 (TX)
#1677 (NY)
#1707 (FL, GA, IL, NC)
#1711 (MI)
#1712 (CO)
#1716 (VA)

Before those commits these used `us_state` (which has `speed_limit_unit='mph'` set). Since the commits listed above were only to split the states up in smaller extracts, I assumed they were all unintentional.

Untested, but probably the fix for #2408


frodrigo:
Yes, right. Thank you.

Issue #2419, opened 2024-12-23
ref:refuges.info shouldn't be deprecated


Nekzuris:
Osmose report "ref:refuges.info" as deprecated but it's not.
https://wiki.openstreetmap.org/wiki/FR:Key:ref:refuges.info

Famlam:
This issue originates from JOSMs rules ([here](https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/deprecated.mapcss#L709)). We only import these rules.  Can you report it to them please?

frodrigo:
May be it was already removed from JSOM. I do not locate the rule about that.

Famlam:
Line 709 of the file I pointed to :)

frodrigo:
dot "." should not be present in the key.

Nekzuris:
Why is this a problem? I didn't find this rule in the spec.
Then should the tag change of just the validator?

Famlam:
If this is an established tag, it's probably best to contact JOSM as it's their rule, not ours.
However, see https://wiki.openstreetmap.org/wiki/Any_tags_you_like#Syntactic_conventions_for_new_tags
> Some characters have special meaning in various context and using these characters in keys will likely cause all sorts of issues. The following characters are among ones discouraged to be used in keys: =, +, /, &, <, > ;, ', ", ?, %, #, @, \, . and ,.

---

@frodrigo we might want to consider blacklisting this rule anyway because it overlaps with our https://github.com/osm-fr/osmose-backend/blob/dev/plugins/TagFix_BadKey.py plugin. What do you think?

Famlam:
@frodrigo can you clear [class 9002011](https://osmose.openstreetmap.fr/en/issues/open?item=9002&class=9002011) on the server?

frodrigo:
> @frodrigo can you clear [class 9002011](https://osmose.openstreetmap.fr/en/issues/open?item=9002&class=9002011) on the server?

Done

Famlam:
Great

PR #2424, opened 2024-12-28
Remove duplicate rule for key with uncommon characters

Famlam:
The `key with uncommon character` rule from JOSM is less strict than our TagFix_BadKey plugin, but it always overlaps (except for the whitelists we have). Hence I'd propose removing it, to stop duplicate warnings.

This PR:
- Improves the texts of TagFix_BadKey (mostly grammar)
  - I also changed "Bad key" to "Key with uncommon character" as I think that's more descriptive.
- Blacklists the overlapping rule from JOSM

Note: as `fix = T_('''Check the key tag, and correct.'''),` doesn't add much, we might want to consider removing this fix text instead?

After merging this would probably require manual clearing of class 9002011.
See also #2419


frodrigo:
Ok, thank you.

PR #2426, opened 2025-01-01
Error with empty value + warning

Famlam:
Surprisingly this is the first time I'm encountering them, so I always thought they didn't exist, but the following taught me otherwise:
```
2024-12-31 10:39:59       File "/data/project/osmose/backend/plugins/TagFix_Maxspeed.py", line 99, in way
2024-12-31 10:39:59         maxspeed_tags = list(filter(lambda t: t.startswith('maxspeed') and tags[t][0] in "0123456789", tags))
2024-12-31 10:39:59                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-12-31 10:39:59       File "/data/project/osmose/backend/plugins/TagFix_Maxspeed.py", line 99, in <lambda>
2024-12-31 10:39:59         maxspeed_tags = list(filter(lambda t: t.startswith('maxspeed') and tags[t][0] in "0123456789", tags))
2024-12-31 10:39:59                                                                            ~~~~~~~^^^
2024-12-31 10:39:59     IndexError: string index out of range
```

### Tags without value

Some examples from this extract: (this extract has 7 of them)
- https://www.openstreetmap.org/way/474159364
- https://www.openstreetmap.org/node/11141071205

This PR fixes the issue and adds a warning against those that aren't warned against yet.

Also:

### Happy new year!

### DIFF
--- a/plugins/TagFix_BadValue.py
+++ b/plugins/TagFix_BadValue.py
@@ -127,6 +127,13 @@ However, this should probably still conform to the typical format used for value
         for k in keyss:
             if tags[k] in ("unknown", "*"):
                 err.append({"class": 40613, "subclass": stablehash64(k), "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])) })
+            elif len(tags[k].strip()) == 0 and k not in check_list_open and k not in self.check_list_closed:
+                err.append({
+                    "class": 3040,
+                    "subclass": stablehash64(k),
+                    "text": T_("Concerns tag: `{0}`", '='.join([k, tags[k]])),
+                    "fix": {"-": [k]}
+                })
 
         return err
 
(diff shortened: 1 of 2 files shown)


frodrigo:
As this issue my consern any plugin, maybe we should fix this as the parser level ?

Famlam:
> As this issue my consern any plugin, maybe we should fix this as the parser level ?

Not sure. I'm active for Osmose since 2020 and this is the first time I'm encountering it, so I guess it's pretty rare that analysers/plugins error on it. It was just a mistake of mine to not cover for this case. (I also guess there aren't too many plugins that get string indices without any prior check?)
Stripping those tags in the pre-parser would however prevent warnings being issued. So I would be in favor of just fixing any possible plugins that may error on it, and warning about this bad data.

---

p.s. I've reported it to Merkaartor too https://github.com/openstreetmap/merkaartor/issues/310. 

For AED maps I wasn't able to reproduce (couldn't even find an option to add opening_hours or images), it seems the last bad changeset was several months ago: https://www.openstreetmap.org/node/12171268601.

frodrigo:
You are, right. I misunderstoot the issue.

Issue #2442, opened 2025-02-03
Detect Zawgyi (Burmese pseudo-Unicode)


brawer:
Could Osmose detect Burmese text in Zawgyi encoding?

[Zawgyi](https://en.wikipedia.org/wiki/Zawgyi_font) is a historic font encoding that looks like Unicode, and can therefore be passed around as UTF-8, but abuses certain Burmese codepoints in a non-standard way. When rendered as a Unicode string, Zawgyi-encoded text will show up with garbled characters. Although Zawgyi is on the decline, there’s still a fair amount of Zawgyi-encoded text in OpenStreetMap as of early 2025. A current list of bad strings in OSM is found by [this tool](https://bdon.github.io/OpenStreetMap-BurmeseEncoding/).

This tool is written in Go and calls Google’s probabilistic open-source Zawgyi detector, plus a custom open-source Zawgyi-to-Unicode converter. Perhaps the above encoder could upload its findings to Osmose via the XML API. But actually, it should be trivial to implement a new analyzer from scratch as a Python plugin for Osmose. Zawgyi dection and conversion also exist in Python, published as PyPI packages `myanmartools` respectively `PyICU`. A plug-in for Osmose might look roughly like this. What do you think, would you be OK with this? What error class should be used, and how to get an ID for it?

```python
import myanmartools
import PyICU

class TagFix_ZawgyiBurmese(Plugin):

    def init(self, logger):
        Plugin.init(self, logger)
        self.detector = myanmartools.ZawgyiDetector()
        self.converter = PyICU.Transliterator.createInstance('Zawgyi-my')

    def node(self, data, tags):
        err = []
        for key, value in tags.items():
            if not any(0x1000 <= ord(c) <= 0x109F for c in value):
                continue
            score = self.detector.get_zawgyi_probability(value)
            if score < 0.8:
                continue
            fixed_value = self.converter.transliterate(value)
            if value == fixed_value:
                continue
            err.append({
               "class": ????, "subclass": ????,
               "text": T_("Value contains Zawgyi-encoded Burmese"),
               "fix": {key: fixed_value},
            })
        return err

    def way(self, data, tags, nds):
        return self.node(data, tags)

    def relation(self, data, tags, members):
        return self.node(data, tags)

###########################################################################                              
from plugins.Plugin import TestPluginCommon

class Test(TestPluginCommon):
    def test(self):
        a = TagFix_ZawgyiBurmese(None)
        a.init(None)
        for name in [
[...]

PR #2443, opened 2025-02-04
Add plugin for detecting values in Burmese Zawgyi encoding

brawer:
Background: [Zawgyi](https://en.wikipedia.org/wiki/Zawgyi_font) is an obsolete font encoding that is incompatible with proper Unicode. Structurally, Zawgyi strings look like Unicode (they can be passed around and stored as UTF-8, UTF-16, etc.), but when rendered, text gets displayed as garbled characters unless the user happens to have a non-standard font installed. With this non-standard font, the system is able to render Zawgyi, but properly encoded Unicode strings look broken. Also, because the Zawgyi encoding abuses codepoints intended for Myanmar’s minority languages, installing a Zawgyi font breaks the display of text in those minority languages.  The situation is a bit like with ISO 8859 in the 1980s, but worse because Zawgyi text and fonts pretend to be Unicode. Because of the structural similarity to Unicode, detecting Zawgyi is non-trivial and can only be done probabilistically. As of early 2025, Zawgyi is on the decline on the general Internet. However, OpenStreetMap still contains thousands of objects with tag values that are encoded in Burmese Zawgyi instead of proper Unicode.

This Osmose plugin makes use of Google’s open-source Zawgyi detector, which uses Markov chains to estimate the likelihood of a Burmese string being Zawgyi-encoded versus proper Unicode. The Osmose plugin suggests fixes by converting mis-encoded strings from Zawgyi to Unicode using the Unicode ICU library, which comes with a built-in converter for this purpose.

Fixes https://github.com/osm-fr/osmose-backend/issues/2442.

### DIFF
--- a/plugins/TagFix_ZawgyiBurmese.py
+++ b/plugins/TagFix_ZawgyiBurmese.py
@@ -1,0 +1,102 @@
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyrights Sascha Brawer 2025                                         ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
+##                                                                       ##
+## You should have received a copy of the GNU General Public License     ##
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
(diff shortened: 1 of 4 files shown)


plugins/TagFix_ZawgyiBurmese.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def init(self, logger):
+        Plugin.init(self, logger)
+        self.errors[50706] = self.def_class(
+            item = 5070,
+            level = 2,
+            tags = ['value', 'fix:chair'],
+            title = T_('Value contains Zawgyi-encoded Burmese characters'),
+            detail = T_(
+'''Tag values in OpenStreetMap should be stored in Unicode. However, this
```
frodrigo, line 43:
```suggestion
'''Tag values should be stored in Unicode. However, this
```

brawer:
`./tools/pytest.sh sax` fails for me, but these breakages look unrelated, as far as I can tell.

plugins/TagFix_ZawgyiBurmese.py
brawer: Done.

brawer:
OK to merge, or would you like to see some changes?

PR #2435, opened 2025-01-26
Add merge analyser for French Enedis power distribution poles

flacombe:
It is proposed to add a new analyser to merge Enedis open data regarding distribution power poles in France.

As original dataset contains more than 1.5M of suitable features, a manual preprocess is done to restrict to first 300k rows for now. Thus it will generate less than 300k warnings.
Dataset will be regularly updated with new features as the amount of osmose warnings get lower.

This analyser is planed to be updated later with poles material once available in Enedis dataset at the end of the year.

PR #2444, opened 2025-02-09
Set different classes from item 8290 analysers

flacombe:
Fix issue shown by #2435.

All analysers in item 8290 will now have different classes


frodrigo:
Ok. Thank you.

Once deployed, same change has to done into the database.

Issue #2447, opened 2025-02-11
False positive for bridge structure missing when bridge outline is separately mapped


andrebk:
There seems to be a consistent false positive for "bridge structure missing" when `bridge:structure` is added to a separately mapped `man_made=bridge` outline, instead of to the `highway=*` way. Osmose detects that there is a way with `bridge=yes` but no `bridge:structure` tag, but ignores that the way is attached to an outline that _does_ have the tag.

According to the [wiki page for the `bridge=*` tag](https://wiki.openstreetmap.org/wiki/Key:bridge), adding the bridge properties to the outline is the correct mapping approach:
> When [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[bridge](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge) is used for the bridge outline, it should hold all bridge specific-attributes like [name](https://wiki.openstreetmap.org/wiki/Key:name)=*, reference, etc.

This is also described on the [wiki page for the `bridge:structure` tag](https://wiki.openstreetmap.org/wiki/Key:bridge:structure):
> For bridges of a single type, for example an arch bridge it can be useful to add bridge:structure=* on the outline of the entire bridge ([man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[bridge](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge)). It is useful especially if single bridge has large number of ways tagged with [bridge](https://wiki.openstreetmap.org/wiki/Key:bridge)=[yes](https://wiki.openstreetmap.org/wiki/Tag:bridge%3Dyes) - multiple footways, carriageways, maybe split as different attributes change.

The correct mapping is for the way to be attached to the outline, could this connection be used to filter out bridges with the correct tagging on the outline?

**Note:** There is a gotcha here for bridges where the way is split because of changing tagging, e.g. `maxspeed`. In that case the longest way might not be connected to the outline directly, but via other ways. An example of this is [here](https://osmose.openstreetmap.fr/en/issue/b10a5715-6074-b840-056e-21ec72cc2e7c).

Item: 7012
Class: 1
Example: https://osmose.openstreetmap.fr/en/issue/7ce6de68-aaca-0afe-d917-be1a3cb7d22b

Famlam:
Fix deployed

PR #2450, opened 2025-02-14
Disable warning for bridge details for highways/railways in a bridge contour

Famlam:
Disable the warning for missing `bridge:structure=*` or more detailed `bridge=*` if the highway/railway is contained within* (one or more) `man_made=bridge` objects. 

See #2447

*Technically, also if it is only partly within. This keeps the code simple and also allows for the highway/railway to stick out a little bit. Since a single highway can be within multiple `man_made=bridge` objects, the code would become much more complex if I'd have to ensure the entire linestring is within `man_made=bridge` objects.

### DIFF
--- a/analysers/analyser_osmosis_highway_tunnel_bridge.py
+++ b/analysers/analyser_osmosis_highway_tunnel_bridge.py
@@ -25,20 +25,26 @@ from .Analyser_Osmosis import Analyser_Osmosis
 
 sql10 = """
 SELECT
-    id,
-    ST_AsText(way_locate(linestring))
+    ways.id,
+    ST_AsText(way_locate(ways.linestring))
 FROM
-    {0}ways
+    ways
+    LEFT JOIN polygons AS contours ON
+        ST_Contains(contours.poly, ways.linestring) AND
+        contours.tags != ''::hstore AND
+        contours.tags?'man_made' AND
+        contours.tags->'man_made' = 'bridge'
 WHERE
-    tags != ''::hstore AND
+    ways.tags != ''::hstore AND
     (
-        tags?'railway' OR
-        (tags?'highway' AND tags->'highway' IN ('motorway', 'trunk', 'primary', 'secondary'))
+        ways.tags?'railway' OR
+        (ways.tags?'highway' AND ways.tags->'highway' IN ('motorway', 'trunk', 'primary', 'secondary'))
     ) AND
-    tags?'bridge' AND
-    tags->'bridge' = 'yes' AND
-    ST_Length(linestring::geography) > 500 AND
-    NOT tags?'bridge:structure'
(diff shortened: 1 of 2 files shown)


analysers/analyser_osmosis_highway_tunnel_bridge.py
```diff
@@ -25,20 +25,26 @@
...
 SELECT
-    id,
-    ST_AsText(way_locate(linestring))
+    ways.id,
+    ST_AsText(way_locate(ways.linestring))
 FROM
-    {0}ways
+    {0}ways AS ways
+    LEFT JOIN {1}polygons AS contours ON
+        ST_Intersects(contours.poly, ways.linestring) AND
```
frodrigo, line 33: But you implemented it with an `ST_Intersects` and not a `ST_WithIn` ?

analysers/analyser_osmosis_highway_tunnel_bridge.py
```diff
@@ -197,14 +205,15 @@ def __init__(self, config, logger = None):
...
+        self.run(sql10.format("", ""), self.callback10)
         #self.run(sql20.format("", ""))
         #self.run(sql21, self.callback20)
         #self.run(sql30.format("", ""), self.callback30)
         self.run(sql40.format("", ""), self.callback40)
 
     def analyser_osmosis_diff(self):
-        self.run(sql10.format("touched_"), self.callback10)
+        self.run(sql10.format("touched_", ""), self.callback10)
+        self.run(sql10.format("not_touched_", "touched_"), self.callback10)
```
frodrigo, line 216: We cannot do this with a LEFT JOIN... IS NULL. `{1}` could not be restricted.

analysers/analyser_osmosis_highway_tunnel_bridge.py
Famlam: Yes, I think a single road can pass through multiple man_made=bridge sections, e.g. half arch bridge, half suspension bridge. So ST_WithIn wouldn't catch these. But the wiki isn't fully clear to me about this, and I guess these cases are rare...

On the other hand, ST_Intersects also has its flaws (see the footnote in the original post).

You can choose: which one would you prefer?

analysers/analyser_osmosis_highway_tunnel_bridge.py
Famlam: Oh right... And because the man_made can be added without modifying the highway, the `{0}`  should also go

analysers/analyser_osmosis_highway_tunnel_bridge.py
Famlam: I've just changed it to the more restrictive case. This covers at least the vast majority of the cases

Issue #2432, opened 2025-01-11
False positives


ComradeRamen:
A number of newer [motorways](https://osmose.openstreetmap.fr/en/map/#item=1210&zoom=17&lat=20.7805863&lon=-105.2454578&loc=12/20.80826/-105.26329&level=1&class=3) were incorrectly detected.

Famlam:
I guess it's the same issue as #2162

Famlam:
The extract seems fine so my initial suspicion was wrong. (Removed it to avoid confusion)
 
@ComradeRamen do you know why there's `noexit=yes` on https://www.openstreetmap.org/node/9414184192 ? I think this may cause a very rare edge case with an invalid noexit (invalid because there's two connected highways, so regardless of whether it's _allowed_ to drive, it's certainly not a noexit=yes) in combination with the extract border crossing.
If not, I suspect that removing `noexit=yes` will fix this issue.

ComradeRamen:
I've removed the `noexit=yes` tags.
With what mechanism would `noexit=yes` cause these detections?
Note that these detections first occurred in January while the roads were edited December 22nd.
![image](https://github.com/user-attachments/assets/278616d4-1e16-480d-b9d0-0b406ad3033d)

Famlam:
Hmm, yeah on second thought it indeed shouldn't cause this. It now continues further.

Famlam:
This issue has been resolved, the issue was due to a removed relation which we previously used as a border.

PR #2451, opened 2025-02-15
Use new relation for Nayarit

Famlam:
Probably the cause for #2432

It seems the boundary relation of Nayarit that we used was removed:
https://www.openstreetmap.org/relation/7695827
https://www.openstreetmap.org/changeset/161122699

Hence this part of the code didn't work anymore as boundary_ids is obtained from `self.config.polygon_id`
https://github.com/osm-fr/osmose-backend/blob/master/analysers/analyser_osmosis_highway_deadend.py#L179-L189

Note: the actual issues were marked as false positive by someone

@jocelynj not sure if something needs to be changed on the server?


frodrigo:
No need to change sever setup. It is only config.

Nice to spot that. Thank you.

Note, we have a task to check the polygons, but no one check the output. I just look at it, and it is a nightmare :-/

PR #2452, opened 2025-02-16
Recycling BM now include food_waste

trouyer:
[Bordeaux Métropole dataset](https://opendata.bordeaux-metropole.fr/explore/dataset/en_empac_p/) now include food_waste recycling container.

Actual analyser have bug since all entries are mapped with `recycling:glass_bottles=true`.

**Changelogs:**
- filter source that is not  `type_emplacement` `PROJET` 
- map `glass_bottles` and `food_waste` recycling
- add `location=underground` when `type_emplacement` is  `ENTERRE` or `SEMI_ENTERRE`


frodrigo:
Nice, thank you.

trouyer:
You're welcome @frodrigo.

PR #2459, opened 2025-02-22
Improve powerline osmosis analyser with line_management suggestion

### LINKED ISSUE #457: French power TSO location:transition towers

### LINKED ISSUE #2004: New osmosis analysis for location:transition, line_management for power supports
Following #457, processing opendata to provide a file to qualify existing osm transmission towers gives me ideas.
It could be possible to analyze OSM distribution supports against existing osm power lines.
Lines are more widely drawn than [line_management](https://wiki.openstreetmap.org/wiki/Key:line_management) on supports.

The following processing is applicable on any OSM power line and OSM power support, from osmosis database:
- Cut lines/minor_lines/cables at each intersection, default `circuits`=1
- Remove out of service (both planned and disused) underground cables (usually disconnected from overhead lines and dismounted from towers)
- Extrapolate each edge (start/end) of each line and keep `circuits` tag from lines.
- aggregate overlapping edges on the same environment (overhead with overhead and underground with underground), make an ordered list of `circuits` from aggregated edges (for instance, a node connecting a line with circuits=2 and two lines with circuit=1 leads to `2-1-1` on the resulting aggregated edge).
- Intersect overhead and underground edges, preserve overhead and underground circuits configs as separate attributes
- Make a difference between overhead and underground to keep edges covered by a single environment (overhead only and underground only)
- Union intersect and difference edges to get the final file with underground only, overhead only and common ones in the same table. Default circuit configurations to NULL.

It can be achieved by sequential SQL queries only, no need for a Python script.

Two OSM tagging additions can be deduced:
- [location:transition=yes](https://wiki.openstreetmap.org/wiki/Tag:location:transition%3Dyes) on any edge connecting overhead and underground
- [line_management](https://wiki.openstreetmap.org/wiki/Key:line_management) got values as following cases, defined bot for transmission and distribution grids

[Line_management](https://wiki.openstreetmap.org/wiki/Key:line_management) values (to be completed continuously):
```sql
CASE
WHEN config_circuits_overhead='1' AND config_circuits_underground='1' THEN 'transition'
WHEN config_circuits_overhead='2' AND config_circuits_underground='1-1'  THEN 'transition'
WHEN config_circuits_overhead='1-1' AND config_circuits_underground='1' THEN 'branch'
WHEN config_circuits_overhead='1-1' AND config_circuits_underground='1-1' THEN 'split'
WHEN config_circuits_overhead='2-1-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='2-2-2' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='2-2-1-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='3-1-1-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='3-2-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='4-1-1-1-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='4-2-1-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='4-2-2' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='4-3-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='5-2-2-1' AND config_circuits_underground IS NULL THEN 'split'
WHEN config_circuits_overhead='2-2-1' THEN 'straight|branch'
WHEN config_circuits_overhead='2-2' AND config_circuits_underground='1' THEN 'straight|branch'
WHEN config_circuits_overhead='2-1-1-1' AND config_circuits_underground IS NULL THEN 'straight|branch'
WHEN config_circuits_overhead='1-1-1' AND config_circuits_underground IS NULL THEN 'branch'
WHEN config_circuits_overhead='2-1' AND config_circuits_underground='1' THEN 'split|transition'
WHEN config_circuits_overhead='1' AND config_circuits_underground NULL THEN 'termination'
ELSE NULL
END
```

This will lead to warning on power supports nodes respectively missing [location:transition](https://wiki.openstreetmap.org/wiki/Tag:location:transition%3Dyes) or [line_management](https://wiki.openstreetmap.org/wiki/Key:line_management)
They will look like _Ỳou could add location:transition=yes here_ or _You may be interested by adding line_management=branch on this T-connection node_.
It may be required to create a supplementary item beside 7040, up to @frodrigo 

**It is not possible to make the reverse analysis**, like removing `location:transition=yes` or `line_management` when lines don't match the corresponding situation as those attributes may be used to state the presence of unknown/not drawn features, particularly underground.

flacombe:
It is proposed following improvement of osmosis_powerline analyser:
* Fix #2004
* Fix #457 with line_management suggestion from power lines nodes in osm topology
* Removal of #2007 merge analyser dedicated to France (will drop item 8290 / class 2)
* Minor improvement of item 7040 / class 3 with line_management=split exclusion from same voltage connection (implies less warnings)
* Some SQL renaming

Tests done on france_nord_pas_de_calais_nord and took < 5 seconds to find 272 occurrences of item 7040 / class 8.


analysers/analyser_osmosis_powerline.py
```diff
@@ -421,6 +532,17 @@ def analyser_osmosis_diff(self):
         self.run(sql51)
         self.run(sql52, self.callback50)
 
+    def __callback80_fix(self, res):
```
flacombe, line 535: My fix strategy is probably not good, as it produces results like:
```xml
<error class="8">
<node id="993023726" lat="50.5270043" lon="3.0664694000000003" version="2" user="InfosReseaux">
<tag k="power" v="tower" />
<tag k="operator" v="RTE" />
<tag k="design:ref" v="L1" />
<tag k="line_attachment" v="anchor" />
<tag k="operator:wikidata" v="Q2178795" />
</node>
<location lat="50.5270043" lon="3.0664694" />
<fixes>
<fix>
<node id="993023726">
<tag action="create" k="line_management" v="branch" />
</node>
</fix>
<fix>
<node id="993023726">
<tag action="modify" k="line_management" v="branch" />
</node>
</fix>
</fixes>
</error>
```

How can I prevent `modify` item when tag is created in the same warning?

flacombe:
More testing, modifications looks good:
* class 1 => 149
* class 2 => 22
* class 3 => 17 (currently 21, splits removed and additional lacks of transformer=* discovered)
* class 4 => 26
* class 5 => 0
* class 6 => 193 (currently 189, 4 legit adds)
* class 7 => 0
* class 8 => 272

analysers/analyser_osmosis_powerline.py
flacombe: Found out, now fixed

PR #2460, opened 2025-02-22
Remove unnecessary check for transformer frequency tag

russss:
This check claims to come from JOSM but it seems to have been removed from JOSM.

In general, we don't require that frequency is tagged on power objects unless it differs from the standard grid frequency.

This also agrees with the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:power=transformer)

cc @flacombe

### DIFF
--- a/plugins/Power.validator.mapcss
+++ b/plugins/Power.validator.mapcss
@@ -43,12 +43,3 @@ node[power=transformer][voltage] {
     -osmoseItemClassLevel: "9100/91002/2";
     -osmoseTags: list("tag");
 }
-
-node[power=transformer][!frequency] {
-    throwWarning: tr("Power Transformers should have a frequency tag");
-    -osmoseItemClassLevel: "9100/91003/3";
-    -osmoseTags: list("tag");
-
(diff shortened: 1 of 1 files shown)


Famlam:
I have no objections against the patch, but where does it claim to originate from JOSM?

russss:
> I have no objections against the patch, but where does it claim to originate from JOSM?

It's in the "90" (JOSM validator) category although maybe that was an error.

https://osmose.openstreetmap.fr/api/0.3/items/9100/class/91003

frodrigo:
cc @flacombe plase, can you review, as I think, you are the one that make it.

flacombe:
I can't find any history in JOSM tickets where frequency is enforced in [combinations.mapcss](https://josm.openstreetmap.de/log/josm/trunk/resources/data/validator/combinations.mapcss) nor in [french specific rules](https://josm.openstreetmap.de/wiki/Rules/FranceSpecificRules).

So ok to remove.
It will lead to remove 133 k warnings in France

PR #2462, opened 2025-02-23
Adjust polygons using deleted relations

Famlam:
### germany_niedersachsen
- Polygon https://www.openstreetmap.org/relation/454192 was deleted (image: https://polygons.openstreetmap.fr/get_image.py?id=454192&params=0). 
- Propose to replace by https://www.openstreetmap.org/relation/62771

On first glance, the difference seems to be that the original polygon only contained land area, while the new polygon contains the water areas too. Note however that there is overlap with part of NL in the border dispute area in the water... not sure if this has any side effects?

### usa_georgia_southwest
- Polygon https://www.openstreetmap.org/relation/11739369 was deleted (image: https://polygons.openstreetmap.fr/get_image.py?id=11739369&params=0).
- Propose to remove since it's part of https://www.openstreetmap.org/relation/1027982 which is already in the list

For the other failing polygons... I often have no clue why they're failing


frodrigo:
> germany_niedersachsen

With water is better.

Thank you.

> For the other failing polygons... I often have no clue why they're failing

Do you think the polygon base may be corrupted ?

Famlam:
> Do you think the polygon base may be corrupted ?

Not sure, I don't really know how to debug it well as I don't know what the polygon-backend does. I loaded a few polygon-forming-relations in JOSM and didn't see any obvious issues with those, they looked valid to me. But it may be the code behind it that generates something invalid (just like we initially had some difficulties with connected outers in multipolygons)

Issue #2469, opened 2025-03-07
Tests failing due to `{{Taginfo entry|*}}` in tables: template is a cell


Famlam:
Currently tests for TagWatchFrViPofm.py are failing because of this table row:
```
|School:FR||school:FR || typo-tag || FR
{{Taginfo entry|School:FR}}
```

The reason for failing is because FR isn't detected is because the `{{Taginfo entry|*}}` template is defined as a cell, e.g. it seems to insert something like `| 123` (if the taginfo count is 123). Hence, people editing the table omit the (normally) preceeding `|`.

Probably the safest way to deal with this (assuming we'll never actually need the count of taginfo via this template) is to just replace `{{Taginfo entry|*}}` by `|` or `||` using read_wiki_templates.

frodrigo:
That a so frequent issue that we need to deal with it on the code ?

Can't we just fix the wiki ?

Famlam:
It's a [template](https://wiki.openstreetmap.org/wiki/Template:Taginfo_entry) on the osm wiki. I don't think we can delete the template? The wiki isn't broken, this is the intended behavior of the template (even though I dislike it)

frodrigo:
Ok, I understand better. So, yes, I also think what you suggest is a good idea.

PR #2470, opened 2025-03-08
Fix table reading with templates that imply cells

Famlam:
See #2469 

I'm using `read_wiki_templates` because it takes care of all possible lower/uppercase and whitespaces, as well as potential nested templates.

Since I don't change the template name, I need to filter for duplicates prior to calling replace. Otherwise, having the same template twice within the table would insert a `\n|` twice as well.


frodrigo:
Thank you

PR #2476, opened 2025-03-15
Implement multipolygon support for table buildings

Famlam:
### Implement multipolygon support for table `buildings` and convert the analysers accordingly.

Notes:
- For the two cases of `b1.id > b2.id` in analyser_osmosis_building_overlaps I assume this was just such that comparing [A B C] with [A B C] would go like [BA, CA, CB] (n!-n comparisons) rather than [BA, CA, AB, CB, AC, BC] (n^2-n comparisons, if b1!=b2), e.g. it doesn't have to know the numerical value. This makes it much easier to keep it as it is as it may have to compare relations with ways.
- I dropped the `buildings.relation` property. Based on the commit history this was added a decade ago in case a way with tag `building=*` was part of a multipolygon relation. However, nowadays the `building=*` tag should be on the multipolygon relation itself, not on the (outer) way. If it's on the inner way, that's a valid building. (Additionally, this inadvertently also blocked detections involving buildings in e.g. associatedStreet relations)
- In analyser_osmosis_building_in_polygon I switched the other multipolygon implementation to table polygon to simplify it, but forgot to make this a separate commit... apologies
- [offtopic] I'm not sure what analyser osmosis_building_3nodes is supposed to detect, but it seems like an "overlapping buildings" detection dedicated to 3-node buildings and hasn't detected anything since [2018](https://osmose.openstreetmap.fr/nl/issues/graph.png?item=1&source=&class=1&username=&bbox=). Probably it's redundant due to the other overlapping buildings analyser.

### DIFF
--- a/analysers/Analyser_Osmosis.py
+++ b/analysers/Analyser_Osmosis.py
@@ -242,34 +242,25 @@ ANALYZE {0}.polygons;
     sql_create_buildings = """
 CREATE UNLOGGED TABLE {0}.buildings AS
 SELECT
-    *,
-    CASE WHEN polygon_proj IS NOT NULL AND wall THEN ST_Area(polygon_proj) ELSE NULL END AS area
-FROM (
-SELECT DISTINCT ON (id)
     id,
+    type || id AS type_id,
     tags,
-    linestring,
-    CASE WHEN ST_IsValid(linestring) = 't' AND ST_IsSimple(linestring) = 't' AND ST_IsValid(ST_MakePolygon(ST_Transform(linestring, {1}))) THEN ST_MakePolygon(ST_Transform(linestring, {1})) ELSE NULL END AS polygon_proj,
+    poly,
+    poly_proj,
     (NOT tags?'wall' OR tags->'wall' != 'no') AND tags->'building' != 'roof' AS wall,
     tags?'layer' AS layer,
-    ST_NPoints(linestring) AS npoints,
-    relation_members.relation_id IS NOT NULL AS relation
+    ST_NPoints(poly) AS npoints,
+    ST_Area(poly_proj) AS area
 FROM
-    ways
-    LEFT JOIN relation_members ON
-        relation_members.member_type = 'W' AND
-        relation_members.member_id = ways.id
+    polygons
 WHERE
-    tags != ''::hstore AND
     tags?'building' AND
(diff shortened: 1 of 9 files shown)


frodrigo:
> [offtopic] I'm not sure what analyser osmosis_building_3nodes is supposed to detect, but it seems like an "overlapping buildings" detection dedicated to 3-node buildings and hasn't detected anything since [2018](https://osmose.openstreetmap.fr/nl/issues/graph.png?item=1&source=&class=1&username=&bbox=). Probably it's redundant due to the other overlapping buildings analyser.

I guess it was to detected small spited building part imported from French cadastre.

If now it is broken or useless we can remove it, rather than run it every days!

@didier2020 It looks like you made this analyser.

Famlam:
> I guess it was to detected small spited building part imported from French cadastre.

I suspect these cases are covered nowadays by the building_overlaps analyser (class 1 + 2 for overlaps and 6 for "triangles" that only touch). I'll remove it if desired (in a separate PR, after this one)

didier2020:
> > [offtopic] I'm not sure what analyser osmosis_building_3nodes is supposed to detect, but it seems like an "overlapping buildings" detection dedicated to 3-node buildings and hasn't detected anything since [2018](https://osmose.openstreetmap.fr/nl/issues/graph.png?item=1&source=&class=1&username=&bbox=). Probably it's redundant due to the other overlapping buildings analyser.
> 
> I guess it was to detected small spited building part imported from French cadastre.
> 
> If now it is broken or useless we can remove it, rather than run it every days!
> 
> @didier2020 It looks like you made this analyser.
hello, it's old ...

dans le cadaste, les traits noir des delimitations de parcelles "coupaient" les batiments en orange ce qui générait des batiments morcelés. les batiments a 3 points (en triangle) cela n'existe pas trop dans la realité, c'était donc un moyen de detecter ces zones de batiments pour les corriger. Depuis, les exports du cadastres ont été totalement revus, cette analyse n'a plus de sens

In the land registry, the black lines of the plot boundaries "cut" the orange buildings, resulting in fragmented buildings.
Buildings with 3 points (in a triangle) don't really exist in reality, so this was a way to detect these building zones and correct them.
Since then, the land registry exports have been completely revised; this analysis no longer makes sense.

frodrigo:
Thank you

PR #2486, opened 2025-04-01
Remove analyser osmosis_building_3nodes

Famlam:
See discussion in https://github.com/osm-fr/osmose-backend/pull/2476


frodrigo:
Need frontend manual clean after deploy and update.

PR #2504, opened 2025-05-01
Remove unused maxspeed types for Austria

wolfbert:
at:urban30 and at:urban40 have been superseded by at:city_limit30 and at:city_limit40 (which will be checked in a separate plugin).

Part of https://github.com/osm-fr/osmose-backend/issues/2503

PR #2510, opened 2025-05-18
Map AT abbreviations to Osmose equivalent

Famlam:
NSI uses the country codes as defined in https://en.wikipedia.org/wiki/ISO_3166-2:AT
As the country codes were redefined in #2504 we need to map them to the ISO equivalent

@wolfbert

### DIFF
--- a/osmose_config.py
+++ b/osmose_config.py
@@ -1818,15 +1818,15 @@ de_state("thueringen", 62366, "DE-TH")
 
 at_state = gen_country('europe', 'austria', download_repo=OSMFR, language='de', proj=32633)
 
-at_state("niederosterreich", 77189, "AT-3")
-at_state("burgenland", 76909, "AT-1")
-at_state("karnten", 52345, "AT-2")
-at_state("oberosterreich", 102303, "AT-4")
-at_state("salzburg", 86539, "AT-5")
-at_state("steiermark", 35183, "AT-6")
-at_state("tirol", 52343, "AT-7")
-at_state("wien", 109166, "AT-9")
-at_state("vorarlberg", 74942, "AT-8")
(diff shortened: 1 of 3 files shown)


wolfbert:
@Famlam 
Sorry for the additional work, I wasn't aware that the "AT-1" etc. are official (numbers for Austrian provinces aren't used in everyday life, abbreviations are understood). So just in case this creates problems, we can revert this.

Famlam:
> I wasn't aware that the "AT-1" etc. are official (numbers for Austrian provinces aren't used in everyday life, abbreviations are understood). So just in case this creates problems, we can revert this.

Are these well-known abbreviations, as I'm unable to find them on e.g. Wikipedia or so?
(I _think_ that they are only used internally in Osmose (the part after the `-` at least), or am I wrong and did you see AT-1/2/3/... somewhere in the UI?)

wolfbert:
> id you see AT-1/2/3/... somewhere in the UI?)

No, I didn't. When I wrote TagFix_Maxspeed_AT.py, I figured that for further use in AT specific plugins, it would be handy to have codes that are meaningful (believing that the AT-1 and such were OSmose internal). According to [ISO-3166-2:AT](https://en.wikipedia.org/wiki/ISO_3166-2:AT) however, some bureaucrat decided that meaning is overrated. Don't stress over it, feel free to change the codes back to the original (of course I can do that as well, just let me know).

Famlam:
@frodrigo your preference?

Famlam:
Ok, will make the revert commit tonight or Wednesday :)

Famlam:
I made the revert PR as #2511 because that was easier. Closing this PR

PR #2527, opened 2025-06-20
Temporarily disable French power opendata from ODRE

flacombe:
The French opendata provider ODRE recently removed datasets regarding french power transmission network.

This PR disables several analyzers as to prevent any 404 errors during further runs.

@frodrigo I guess current warnings should be removed as they won't be refreshed.
But is it possible to make backend run on cache forever instead?


flacombe:
Given problem is datasets has been restored without geometry features, so probably Osmose will update its cache with bogus files.

Towers and branches hasn't been restored (and probably won't be) but lines and substation have:
https://odre.opendatasoft.com/explore/dataset/lignes-aeriennes-rte-nv/information
https://opendata.reseaux-energies.fr/explore/dataset/postes-electriques-rte

frodrigo:
If you still want this to me merged, please, rebase.

PR #2533, opened 2025-07-01
Update analyser_merge_college_FR.py

Olyon:
Data source update. The old is out of date. The new is almost the same, but there are more establishments, so we need to sort them to keep only amenity=college and amenity=university.
I'm not sure which ones to keep, all the values are in "select = {"type d'établissement":...", lycées and maison familiale rurale are managed by another analysis, those I've excluded are commented.

### DIFF
--- a/analysers/analyser_merge_college_FR.py
+++ b/analysers/analyser_merge_college_FR.py
@@ -37,8 +37,8 @@ class Analyser_Merge_College_FR(Analyser_Merge_Point):
             title = T_('College update'))
 
         self.init(
-            u"https://www.data.gouv.fr/fr/datasets/ideo-structures-denseignement-superieur/",
-            u"Idéo-Structures d'enseignement supérieur",
+            "https://www.data.gouv.fr/fr/datasets/ideo-structures-denseignement-superieur/",
+            "Idéo-Structures d'enseignement supérieur",
             CSV(
                 SourceDataGouv(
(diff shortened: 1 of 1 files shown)


analysers/analyser_merge_college_FR.py
```diff
@@ -37,26 +37,63 @@ def __init__(self, config, logger = None):
...
                 conflationDistance = 500,
                 mapping = Mapping(
-                    static1 = {"amenity": "college"},
-                    static2 = {"source": self.source},
+                    static1 = {"source": self.source},
                     mapping1 = {
+                        "amenity": lambda res: "university" if res["type d'établissement"] in [u"institut national supérieur du professorat et de l'éducation", u"institut universitaire de technologie", u"unité de formation et de recherche", u"université"] else "college",
                         "ref:UAI": "code UAI",
-                        "operator:type": lambda res: "private" if res["statut"] in [u"Privé hors contrat", u"Privé reconnu", u"Privé sous contrat"] else None,
+                        "operator:type": lambda res: "private" if res["statut"] in [u"privé hors contrat", u"privé reconnu par l’Etat", u"privé sous contrat", u"privé"] else None,
```
frodrigo, line 96:
```suggestion
                        "operator:type": lambda res: "private" if res["statut"] in ["privé hors contrat", "privé reconnu par l’Etat", "privé sous contrat", "privé"] else None,
```

frodrigo:
Thank you, merged, but not deployed yet.

PR #2540, opened 2025-07-04
Fix CI and docker - remove hidden char

flacombe:
Remove an hidden char in the CI and docker-compose files.

This changes cause a test to end up in error:
`analysers.analyser_osmosis_highway_deadend.Test testMethod=test_classes`

But the outcome looks right to me, no oneway looks inaccessible in the test case


frodrigo:
Thank you

PR #2546, opened 2025-07-06
Update analyser tests after postgis update

Famlam:
The recursive check now seems to follow a different path, leading to different outcomes for the tests involving circular deadends. This is not entirely unexpected, it happened before if the query changed just a little bit. The results are still valid (there's just multiple possible "good" outcomes)


jocelynj:
@Famlam : could you rebase your commit on `main`, so that we can be sure that it is fixing the last CI issue?

Famlam:
Well, we also need this fix :)
The roundabout one _probably_ only needs a coordinates update

PR #2547, opened 2025-07-07
test: Fix comparison of lat/lon

jocelynj:
Floating-point values must be rounded if we want to compare them, as there is no guarantee on the final digits between postgis and python.


jocelynj:
Note that we also need the fix from #2546 to make all tests passing.

PR #2553, opened 2025-07-12
source update. same data but organized differently

Olyon:


Olyon:
The analysis no longer works. https://osmose.openstreetmap.fr/fr/issues/open?source=7333&item=8200&class=1
Some things have disappeared from the openfuelmap.net file, and there is now a usable data source on data.gouv.fr.

frodrigo:
Ok. Thank you.

PR #2571, opened 2025-08-05
enrich irve socket max output

tykayn:
removes values that are more than the max known of 400kW to be set as the max power output of a charging station, adds a tag to correct values.
Also handles enumerations of this values to keep the maximum value.

The column evaluated in the dataset of French IRVE is "puissance_nominale" by the function "getPuissanceNominaleInKw".

Tested locally with docker on Monaco and Ile de France zones, returns proposals for the "charging_station:output" tag as "n kW" when finding good values.


analysers/analyser_merge_charging_station_FR.py
```diff
@@ -38,6 +39,95 @@ class Analyser_Merge_Charging_station_FR(Analyser_Merge_Point):
...
+            return f"{v:.6f}".rstrip("0").rstrip(".")
+        except Exception:
+            return str(v)
+
+    def getPuissanceNominaleInKw(self, raw):
+        """Computes nominal power in kW from a possible enumeration of values."""
+        if raw is None:
+            return None
+
+        s = str(raw).strip()
```
frodrigo, line 91: raw is not alredy a string ?

frodrigo:
@tykayn please, can you review the comments.

tykayn:
hi @Marc-marc-marc 
it is true, the field seem to be missing in the pre-processed.csv, we should add it.

in the original open data it apprears to be associated to maybe one or more sockets, so it is per stall. But open data providers seem to disagree about wether a line in the spreadsheet is a stall or a pool of stalls.
in the wiki i noticed that charging_station:output is giving the max kW of a stall, but if it is not used to give this info we have a problem and maybe we need some cleanup or an other tag to clearly state that the number is a sum of the charge points, something like "charging_station:output:sum", because knowing the max kW of a stall is very useful, where knowing the sum of a pool is not useful at all for most consumers who want to know if their vehicle will charge at a reasonable speed at one stall.

it didnt make crash my dev env when i ran the analysis, so that is weird.

the PR of @guiohm is interesting, i hope it will bring some clarity.

PR #2573, opened 2025-08-06
Update analyser_merge_school_FR.py

nlehuby:
Change source to another file from the Minister, with less guessing needed

https://forum.openstreetmap.fr/t/school-fr-et-osmose/12801/18


frodrigo:
Thank you.

Ping @vinber

nlehuby:
thanks, could you please let me know when it's up on osmose ?

Issue #2563, opened 2025-07-23
"Close similar name" errors for "NNth Street"


BrianMacIntosh:
The "Close similar name" test turns up errors for nearby streets named in the pattern "NNth Street", but with different numbers. Names in this format might warrant a special case in the test, since this is a very common pattern.

Here is an example. There are a few more nearby:
https://osmose.openstreetmap.fr/en/issue/29e7e0aa-32a2-2411-c65d-b04f385f4148 ([way 13311349](https://www.openstreetmap.org/way/13311349) / [way 13355182](https://www.openstreetmap.org/way/13355182))

Famlam:
Seems this regex needs a little tweak
https://github.com/osmose-qa/osmose-backend/blob/master/analysers/analyser_osmosis_highway_name_close.py#L30

It converts 8th to `_th` and 14th to `1_th` and then it detects that there's only 1 character difference for otherwise identical names.

PR #2581, opened 2025-08-23
Avoid close name for NNth

Famlam:
See #2563 

This avoids e.g. `8th`, `14th` and `22nd` from being converted to `_`, `1_` and `2_` respectively, after which a 1-character difference remains (which is subsequently reported as a close name)

~Not sure why all tests fail but this seems unrelated~

### DIFF
--- a/analysers/analyser_osmosis_highway_name_close.py
+++ b/analysers/analyser_osmosis_highway_name_close.py
@@ -79,7 +79,7 @@ class Analyser_Osmosis_Highway_Name_Close(Analyser_Osmosis):
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
 
-        # Check langues for country are writen with alphabets
+        # Check that languages for countries are written with alphabets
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_highway_name_close.py
```diff
@@ -27,7 +27,7 @@
 sql10_regex = """regexp_replace(regexp_replace(regexp_replace(regexp_replace(regexp_replace(regexp_replace(regexp_replace(regexp_replace({0},
 '([0-9]+-[а-я]{{2}})( |$)', '_', 'g'), -- Bulgaria
 '[-\\[\\]\\{{\\}}\\(\\)\"\\\\/]', '', 'g'),
-'(1st|2nd|3rd|[04-9]th)( |$)', '_', 'g'),
+'[0-9]*(1st|2nd|3rd|[04-9]th)( |$)', '_', 'g'),
```
frodrigo, line 30: Please, can you also do it for Bulgaria and Spanish line.

PR #2589, opened 2025-08-30
Simplify unclosed area-style way checks

Famlam:
Presently, the check for ways that should always be areas is a combination of a plugin (Structural_UnclosedArea) and an SQL analyser (part of analyser_osmosis_relation_multipolygon).

- The plugin is made redundant by the analyser (or vice versa) as they report on the same objects, see for instance [this](https://osmose.openstreetmap.fr/nl/issue/b744cbdb-5135-5714-c744-fc90e0d9e1c5) and [this](https://osmose.openstreetmap.fr/nl/issue/51c3df48-dbfb-d16e-1973-17774da41a17) for the same object.
- The analyser currently JOINs on relation_members to filter out unclosed ways in a multipolygon relation. However, in case of a multipolygon relation, the area-specific tags shouldn't be on the individual ways anymore. Instead, they should be on the multipolygon relation. This additional JOIN is thus unnecessary.

To make it easy to adjust the rules and to allow usage in JOSM (although there is some overlap with [JOSMs rules](https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/data/validation/tests/UnclosedWays.java)) I've made it into a mapcss-based plugin instead. I've also extended it with various other area-only tags and incorporated the exceptions from [JOSMs validator](https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/data/validation/tests/UnclosedWays.java)

Multipolygons and nodes aren't affected.


frodrigo:
Ok. nevertheless, I'm unsure about using exclusion list of values rather than inclusion. I'm worried that raise too much false positives.

Famlam:
I can test more extracts if desired?

Any particular key that worries you?

I already had these rules in my local mapcss file for several years, and also checked the JOSM issue tracker whether they had issues on theirs, but it seems to be quite robust.
(I excluded 'sport' and 'boundary' though)

Famlam:
I've updated the PR after testing it on:
```
netherlands_flevoland
mexico_puebla
vietnam
germany_niedersachsen
france_centre_indre_et_loire
france_languedoc_roussillon_aude
usa_washington
venezuela
japan_chubu
germany_saarland
slovenia
usa_iowa
netherlands_gelderland
sweden_gotland
switzerland_basel_stadt
san_marino
jersey
belgium_wallonia_german_community
monaco
```
and have whitelisted `amenity=hitching_post`, `military=obstacle_course` and `tourism=yes`. I've also removed `information=*` as it's always together with `tourism=information` already.

frodrigo:
Ok. Thank you.

PR #2599, opened 2025-10-01
Add cuisine=açaí to the allowed exceptions from the usual format.

kvibber:
Several JOSM/Vespucci brand-name presets use this tag value.

https://taginfo.openstreetmap.org/tags/cuisine%3Da%C3%A7a%C3%AD


frodrigo:
Maybe the tag itself should be fixed, rather than Osmose-QA ?

PR #2609, opened 2025-10-27
Improve basque multilingual

Crashillo:
No further check whenever the name is already the combined form


frodrigo:
Thank you.

Tests fixed upstream.

PR #2611, opened 2025-10-27
Update Phone.py to add belgium

Marc-marc-marc:
add config for Belgium


Marc-marc-marc:
let's move the "-" issue in another PR
ok for merging BE config

frodrigo:
Thank you

PR #2628, opened 2025-11-21
Update analyser_merge_poste_FR : encoding et columns change in data source

Olyon:
encoding : now it's utf.
wheelchair : The columns have changed. There are now many columns for disabilities, but all have the same value for wheelchair.
"No" value : "Non" or nothing, depending on the column.


Olyon:
We have the accessibility columns with only "Oui" and "Non" and the other columns that interest us with only "Oui" and blanks.
So now the analysis will suggest assigning a value to wheelchair=*, but for atm, stamping_machine, and change_machine, it will only suggest "yes", which corresponds to blanks meaning we don't know.

PR #2629, opened 2025-11-21
Modify mapping for underwear shops

lmagreault:
shop=underwear is deprecated.
Use shop=clothes with clothes=underwear instead.

### DIFF
--- a/merge_data/shop_FR.mapping.json
+++ b/merge_data/shop_FR.mapping.json
@@ -477,7 +477,7 @@
       }
     ],
     "generate": {
-      "shop": "fabric;wool;household_linen;underwear;clothes;sewing",
+      "shop": "fabric;wool;household_linen;clothes;sewing",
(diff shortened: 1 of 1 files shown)


merge_data/shop_FR.mapping.json
```diff
@@ -468,12 +468,16 @@
...
+        "shop": "clothes",
+        "clothes": "underwear"
+      },
       {
         "shop": "sewing"
       }
     ],
     "generate": {
-      "shop": "fabric;wool;household_linen;underwear;sewing",
+      "shop": "fabric;wool;household_linen;underwear;clothes;sewing",
```
frodrigo: As deprecated, we should remove `underwear` form produced values?
```suggestion
      "shop": "fabric;wool;household_linen;clothes;sewing",
```

frodrigo:

PR #2636, opened 2025-11-26
Update speed limit for cz:trunk to 110, to match reality and OSM wiki 

Piskvor:
Maxspeed at highway=trunk in Czechia is 110, minspeed is 80. Fixed from previous 130 (which has been wrong for years now, and keeps flagging correctly tagged ways).

See also https://wiki.openstreetmap.org/wiki/Default_speed_limits#cite_ref-:41_158-0 - the PR fixes this value so that it syncs with reality and the wiki maxspeed.


frodrigo:
Thank you. Merged but not deployed yet.

PR #2661, opened 2025-12-28
Update mapcss

Famlam:
As I was recompiling mapcss for #2660 and #2659 anyway let's also update the other (unrelated) mapcss rules :)

Likely class 9019008 will have to be manually cleared from the server?


frodrigo:
> [9019008](https://osmose.openstreetmap.fr/nl/issues/open?item=9019&class=9019008)

Done.

PR #2688, opened 2026-01-18
follow wiki redirect

Marc-marc-marc:


Famlam:
I'd say that - if this is implemented, it shouldn't be limited to this specific plugin but a generic function (in wikiReader) for all (OSM) wiki readers.

(Would actually be nice if the tests would still error, or if it would be a more visible error in the buildbot logs, such that it wouldn't go unnoticed for very long)

Marc-marc-marc:
> I'd say that - if this is implemented, it shouldn't be limited to this specific plugin but a generic function (in wikiReader) for all (OSM) wiki readers.

done.. is the new code ok ?

> (Would actually be nice if the tests would still error, or if it would be a more visible error in the buildbot logs, such that it wouldn't go unnoticed for very long)

I added a logger.warning but got CI repport :
plugins/modules/wikiReader.py:27:1 W0611 'modules.OsmoseLog' imported but unused [pyflakes]
plugins/modules/wikiReader.py:51:13 E0602 undefined name 'logger' [pyflakes]

plugins/modules/wikiReader.py
```diff
@@ -38,6 +39,26 @@
...
+    # Check for wiki redirect and follow it automatically
+    redirect_pattern = re.compile(r'^\s*#REDIRECT\s*\[\[([^\]]+)\]\]', re.IGNORECASE | re.MULTILINE)
+    max_redirects = 5
+
+    for i in range(max_redirects):
+        match = redirect_pattern.match(wikitext.strip())
+        if match:
+            from modules.downloader import urlread
+            redirect_target = match.group(1)
+            logger.warn(f"Wiki redirect detected: following to '{redirect_target}'")
```
Famlam: I don't think plugins have default support for logging.

```suggestion
            OsmoseLog.logger().warn(f"Wiki redirect detected: following to '{redirect_target}'")
```

Famlam (changes requested):
I think the suggested fix should resolve the linter issues.

The current fix will only fix `read_wiki_table`, but not the other wiki-reading functions.

I would create a new function `wikiurlread` (in this module) that replaces the call to `urlread` in the plugins, having the same arguments as urlread (and internally still calling urlread), but following redirects when encountered.

Marc-marc-marc:
I have created a _follow_redirects function that handles redirection in the four existing functions.

plugins/modules/wikiReader.py
```diff
@@ -24,8 +24,35 @@
...
 
+def _follow_redirects(wikitext: str) -> str:
+    """
+    Internal function to follow wiki redirects in wikitext content.
+    Args:
+        wikitext: The wiki text content that might contain a redirect
+    Returns:
+        The final content after following any redirects
+    """
+    from modules.downloader import urlread
```
frodrigo: Please, do not import from function.

Marc-marc-marc: done but unrelated error in CI : ModuleNotFoundError: No module named 'pkg_resources' (previously part of setuptools)

frodrigo:
@Marc-marc-marc please, can you rebase

Marc-marc-marc:
@frodrigo done

Famlam (changes requested):
As suggested before I'd go for a `urlwikiread` function rather than the current implementation.

Although the current implementation will work for _some_ cases, there are two problems with the current implementation:
1. it ignores the cache time of the original request, instead it always uses 1 day for the redirect
2. some wikireaders do not read wiki.openstreetmap.org, but e.g. Wikipedia pages (tagfix_postcode for instance). The redirect however always becomes a hardcoded wiki.openstreetmap.org URL.

Hence I would certainly go for an approach like
```py
def urlwikiread(url: str, cacheInterval: int = X, maxRedirects: int = 5):
  wikitext = urlread(url, cacheInterval)
  # detect redirects
  # if redirect detected
    # provide log, determine new URL (independent of domain)
    wikitext = urlwikiread(new_url, cacheInterval, maxRedirects - 1)
  return wikitext
```
and call this instead of urlread in the plugins

Marc-marc-marc:
> 2\. some wikireaders do not read wiki.openstreetmap.org, but e.g. Wikipedia pages (tagfix_postcode for instance).

I dind't see tagfix_postcode in https://github.com/osmose-qa/osmose-backend/tree/dev/analysers
where can I find it ?

Famlam:
https://github.com/osmose-qa/osmose-backend/tree/dev/plugins 
(It's only plugins that use the wikireader, no analysers)

Marc-marc-marc:
I wanted to avoid having to touch the calls in analysers/*, but 

> 2\. some wikireaders do not read wiki.openstreetmap.org, but e.g. Wikipedia pages (tagfix_postcode for instance). The redirect however always becomes a hardcoded wiki.openstreetmap.org URL.

calling this code in a plugin for Wikipedia would cause a regression, as it would redirect to an incorrect URL (note that the current situation is not good either, as such a redirection would also cause a crash).
I will look into modifying the calls to resolve this.

frodrigo:
Closed in favor of #2828

PR #2699, opened 2026-02-14
Switch pylam fork

frodrigo:


Famlam (approved):
To me this looks like a good temporary solution till we switch to an actively maintained plugin. (I hope to have time for that soon)

Issue #2560, opened 2025-07-19
False positive Power object intersecting building


AurimasF:
https://osmose.openstreetmap.fr/en/issue/bde6bf0b-c8f0-6e66-faa3-c579487543c4
https://osmose.openstreetmap.fr/en/issues/open?source=6649&item=1070&class=6

Power tower is located on the roof of the power plant building, is tagged with layer=1 and location=roof, but still reported as an issue.

frodrigo:
For exception, just mark it as false positive ?

cc @flacombe

Famlam:
I'd be in favor of whitelisting these cases. In the EU alone, there's already a few hundred of these.
https://overpass-turbo.eu/s/28b3

A simple `WHERE`-statement here should be enough to fix it
https://github.com/osmose-qa/osmose-backend/blob/dev/analysers/analyser_osmosis_highway_vs_building.py#L238-L250

flacombe:
Hello

Indeed `location=roof` looks like the best option to state they are on the building's roof.
We should also exclude power supports on ground versus underground building that may be located underneath (building with `location=underground`)

zigad:
This should be whitelisted as location=roof is correct one. I have many of this issues and have to manually tag them as false positives in Osmose. So +1 from my side to whitelist this

PR #2704, opened 2026-02-24
Don't complain about power poles on rooftops

Famlam:
See #2560

A power pole shouldn't be too close to a building, except when it's actually located on a rooftop.

Also exclude power poles on top of underground buildings as requested here: https://github.com/osmose-qa/osmose-backend/issues/2560#issuecomment-3094649686


frodrigo:
Thank you

Issue #2705, opened 2026-02-25
Phone.py : config in unity test is not the same as outsite it


Marc-marc-marc:
for ex 
[test_BE have phone_format](https://github.com/osmose-qa/osmose-backend/blob/dev/plugins/Phone.py#L364C105-L364C117)
[be_part doesn't have it](https://github.com/osmose-qa/osmose-backend/blob/dev/osmose_config.py#L750)
short_lengh is not the same

same issue with other country unity tests

I plan a two-step solution:
- synchronise the configuration differences between the two
- modify the unit tests so that they do not require a specific configuration, but use the actual configuration instead

PR #2706, opened 2026-02-26
phone : sync conf be_part and test_BE

Marc-marc-marc:
related #2705
also add unity test for phone_len 9


Famlam:
FYI, there's now 50k issues in Belgium #2722

Marc-marc-marc:
thanks for the typo fix

PR #2728, opened 2026-04-21
NZ is a bilingual country

### LINKED ISSUE #2473: Osmose default and local language name not the same NZ
In New Zealand, due to having two offical languages, the offical name of places often include both languages in the name, which causes issues with Osmose as when each name is also added as name:en and name:mi (which are parts of the main name) they are flagged as not matching the local language name. The offical name will never match the local language name (i assume is implied english but doesn't matter) as the offical name contains two languages. Sorry if i explained this badly, please see attatched screenshots. There are many cases of this, mostly around schools and goverment facilities.

Image

k-yle:
Closes #2473

osmose currently generates thousands of false-positive errors because it assumes that english is the only language spoken in New Zealand, which is not correct.

The convention used for bilingual names in NZ is too complex to model in [Name_Multilingual.py](https://github.com/osmose-qa/osmose-backend/blob/dev/plugins/Name_Multilingual.py), see https://github.com/osmose-qa/osmose-backend/issues/2473#issuecomment-3876959945 for a detailed explantation.

### DIFF
--- a/osmose_config.py
+++ b/osmose_config.py
@@ -1512,7 +1512,7 @@ default_country("central-america", "usa_virgin_islands", 286898, {"country": "VI
 
 #########################################################################
 
-default_country("australia-oceania", "new-zealand", 556706, {"country": "NZ", "language": "en", "proj": 32759, "driving_side": "left", "addr:street_distance": 2000})
+default_country("australia-oceania", "new-zealand", 556706, {"country": "NZ", "language": ["mi", "en"], "proj": 32759, "driving_side": "left", "addr:street_distance": 2000})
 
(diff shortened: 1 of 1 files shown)


frodrigo:
Ok, Thank you.
Note, merged, but not yet deployed.

Famlam:
This causes one of the tests to fail (hmm, seems they didn't run in this PR?)

The reason is that there's no script mapped for "mi" here:
https://github.com/osmose-qa/osmose-backend/blob/dev/modules/languages.py

k-yle:
oh sorry, I think the CI [doesn't run for first-time contributors](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories)... I've opened #2729 to fix this, and the CI has passed on that PR

PR #2729, opened 2026-04-22
add missing script definition for language mi

k-yle:
technically there are [only 15 letters](https://www.maorilanguage.net/maori-alphabet/), not 26. But it is common to use loan-words from English, so I've included all `A-Za-z`

### DIFF
--- a/modules/languages.py
+++ b/modules/languages.py
@@ -73,6 +73,7 @@ language2scripts = {
     'lt': ['Latin'],
     'lv': ['Latin'],
     'mg': ['Latin'],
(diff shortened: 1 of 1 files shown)


modules/languages.py
```diff
@@ -73,6 +73,7 @@
     'lt': ['Latin'],
     'lv': ['Latin'],
     'mg': ['Latin'],
+    'mi': [u'[A-Za-zĀĒĪŌŪāēīōū]'],
```
frodrigo:
```suggestion
    'mi': ['[A-Za-zĀĒĪŌŪāēīōū]'],
```

PR #2738, opened 2026-04-28
Update TagFix_BadValue.py - exclude crossing=separate

rskedgell:
Add recently documented tag crossing=separate to list in allow_closed

https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dseparate

### DIFF
--- a/plugins/TagFix_BadValue.py
+++ b/plugins/TagFix_BadValue.py
@@ -95,7 +95,7 @@ However, this should probably still conform to the typical format used for value
         self.allow_closed = { "area": ( "yes", "no", ),
                             "backrest": ( "yes", "no", ),
                             "conveying": ( "yes", "forward", "backward", "no", "reversible", ),
-                            "crossing": ( "traffic_signals", "uncontrolled", "unmarked", "no", "marked", "zebra", "informal", ),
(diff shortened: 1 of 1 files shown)


frodrigo:
I am unsure about that.
The addition to the wiki is from one month ago only, the dedicated page was create by the same user adding it to
- https://wiki.openstreetmap.org/wiki/Key:crossing
- https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dseparate

https://taginfo.openstreetmap.org/tags/crossing%3Dseparate#overview

There is some discussion about this tag value at some place ?

@Famlam what do you think about that?

Famlam:
I'm not a big fan of `=separate` as you effectively map something that's not there but somewhere else. Usage is fairly localized, so I agree, some link to a discussion approving or recommending this tag would be preferred, unless the tag really catches on and becomes more popular

frodrigo:
I close for now. But could reopen if you come back with more details.

rskedgell:
Thanks, that's fine. I'll take a look in a few months to see if it's been more widely adopted (i.e. by mappers other than me and whoever added the tag to the wiki). If it hasn't been, I'll probably revert all my uses of it.

PR #2740, opened 2026-05-03
Update analyser_osmosis_duplicated_geotag.py - False positive

didier2020:
False positive for
natural=cliff, man_made=embankment or barrier=guard_rail when different ordering

### DIFF
--- a/analysers/analyser_osmosis_duplicated_geotag.py
+++ b/analysers/analyser_osmosis_duplicated_geotag.py
@@ -191,7 +191,10 @@ WHERE
     (NOT b1.tags?'level' AND NOT b2.tags?'level' OR b1.tags->'level' = b2.tags->'level') AND
     (NOT b1.tags?'addr:floor' AND NOT b2.tags?'addr:floor' OR b1.tags->'addr:floor' = b2.tags->'addr:floor') AND
     (NOT b1.tags?'min_height' AND NOT b2.tags?'min_height' OR b1.tags->'min_height' = b2.tags->'min_height') AND
-    (NOT b1.tags?'ele' AND NOT b2.tags?'ele' OR b1.tags->'ele' = b2.tags->'ele')
+    (NOT b1.tags?'ele' AND NOT b2.tags?'ele' OR b1.tags->'ele' = b2.tags->'ele') AND
+    (b1.tags?'natural' AND  b1.tags->'natural' = 'cliff' AND b2.tags?'natural' AND  b2.tags->'natural' = 'cliff' ) OR
+    (b1.tags?'man_made' AND b1.tags->'man_made' = 'embankment' AND b2.tags?'man_made' AND b2.tags->'man_made' = 'embankment') OR
+    (b1.tags?'barrier' AND b1.tags->'barrier' = 'guard_rail' AND b2.tags?'barrier' AND b2.tags->'barrier' = 'guard_rail')))    
 """
 
 sql30 = """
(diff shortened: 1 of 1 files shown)


analysers/analyser_osmosis_duplicated_geotag.py
```diff
@@ -191,7 +191,11 @@
     (NOT b1.tags?'level' AND NOT b2.tags?'level' OR b1.tags->'level' = b2.tags->'level') AND
     (NOT b1.tags?'addr:floor' AND NOT b2.tags?'addr:floor' OR b1.tags->'addr:floor' = b2.tags->'addr:floor') AND
     (NOT b1.tags?'min_height' AND NOT b2.tags?'min_height' OR b1.tags->'min_height' = b2.tags->'min_height') AND
-    (NOT b1.tags?'ele' AND NOT b2.tags?'ele' OR b1.tags->'ele' = b2.tags->'ele')
+    (NOT b1.tags?'ele' AND NOT b2.tags?'ele' OR b1.tags->'ele' = b2.tags->'ele') AND
+    NOT (ST_OrderingEquals(b1.linestring, b2.linestring) AND
```
frodrigo: Why do you check about `ST_OrderingEquals`?

analysers/analyser_osmosis_duplicated_geotag.py
```diff
@@ -191,7 +191,11 @@
...
+    ((b1.tags?'natural' AND  b1.tags->'natural' = 'cliff' AND b2.tags?'natural' AND  b2.tags->'natural' = 'cliff' ) OR
+    (b1.tags?'man_made' AND b1.tags->'man_made' = 'embankment' AND b2.tags?'man_made' AND b2.tags->'man_made' = 'embankment') OR
+    (b1.tags?'barrier' AND b1.tags->'barrier' = 'guard_rail' AND b2.tags?'barrier' AND b2.tags->'barrier' = 'guard_rail')))
```
frodrigo: I think the logic you want to write is more like this ?
```suggestion
    (NOT b1.tags?'ele' AND NOT b2.tags?'ele' OR b1.tags->'ele' = b2.tags->'ele') AND
    NOT (b1.tags->'natural' = 'cliff' AND b2.tags->'natural' = 'cliff' AND ST_OrderingEquals(b1.linestring, b2.linestring)) AND
    NOT (b1.tags->'man_made' = 'embankment' AND b2.tags->'man_made' = 'embankment' AND  ST_OrderingEquals(b1.linestring, b2.linestring)) AND
    NOT (b1.tags->'barrier' = 'guard_rail' AND b2.tags->'barrier' = 'guard_rail' AND ST_OrderingEquals(b1.linestring, b2.linestring))
```

`ST_OrderingEquals` Cannot be at first level, because it will also apply to previous lines.

`ST_OrderingEquals` could be refactored, but will result in less readable code.

didier2020:
example of false positive issue : https://osmose.openstreetmap.fr/fr/issue/c80cc287-7b1d-2975-0a52-5874a3238598

frodrigo:

Famlam:
A bit too late but wouldn't this be incorrect? If the embankment or cliff has two (inverted) ways with exactly the same position, then the physical width of this object is exactly 0, which is impossible. A single-atom cliff for instance. Even if it's a really narrow ridge there should be at least some space between the ways, not?

didier2020:
Thanks we will see that soon +1

Famlam:
> Thanks we will see that soon +1

You mean if the issue mentioned above (https://osmose.openstreetmap.fr/fr/issue/c80cc287-7b1d-2975-0a52-5874a3238598, involving https://www.openstreetmap.org/way/1368215997) disappears this should be reversed? Otherwise I'm not sure what you mean

didier2020:
There is issues with this update : duplicated way are no longer detected.
I tried to find it but without success.

Famlam:
Could it be related to #2799?

Seems there can be a duplicate entry which crashes frontend

didier2020:
je ne sais pas repondre, je ne reproduit pas cette erreur avec docker

didier2020:
with local docker it's now ok
but not there 
https://github.com/osmose-qa/osmose-backend/pull/2810

didier2020:
see https://github.com/osmose-qa/osmose-backend/pull/2823 for fixing duplicated way are no longer detected.

PR #2744, opened 2026-05-06
Improve osmosis_powerline unfinished test

### LINKED ISSUE #2743: [item 7040, class 6] Unfinished power distribution line
https://osmose.openstreetmap.fr/en/issues/open?item=7040&source=&class=6

The detail of this reads as follows: https://osmose.openstreetmap.fr/en/issue/7205f6f8-5e6c-3dd2-173c-342f815bcef5

"The line ends in a vacuum, and should be connected to another line or a substation (`power=substation`), a transformer (`power=transformer`), a generator (`power=generator`), marked as transitioning into ground (`line_management=transition + location:transition=yes`) or explicitely terminated (`line_management=termination`)."

Part of the above statement may be incorrect. I think "a transformer (`power=transformer`)" should read "a transformer (`transformer=*`)" as pole- and tower-mounted transformers don't use power=transformer. On superficial inspection, this seems to be the test that is actually applied.

I think this class is unfair and inappropriate, as it may not be readily known or knowable how or where the line ends. Satellite photos will often tell us where there are tower and poles, and sometimes lines. Aerial photos might even show minor_lines, sometimes even to individual houses. Street level photos might show you all sorts of details. Or they might show none. Or the line ends somewhere near a farm 2km from the nearest public road and there are no street level photos.

This may put unnecessary stress on some contributors, as they feel the list of issues represents something they have done "wrong". Similar tests do not seem to be applied to other features, e.g. "Unfinished railway - no buffer stop" or "There are houses 30 metres from this residential road, why haven't you mapped the driveways?".

There is a similar class, [item 7040, class 2] Unfinished power transmission line, where it might not be unreasonable to retain it, given the general importance of such lines and the fact that they are less likely to just end 'randomly'.

flacombe:
This PR aims to improve osmosis_powerline unfinished power lines warnings.

Fix #2743

### DIFF
--- a/analysers/analyser_osmosis_powerline.py
+++ b/analysers/analyser_osmosis_powerline.py
@@ -262,10 +262,10 @@ FROM
         line_ends.nid = nodes.id
     WHERE
         NOT (nodes.tags?'location:transition' AND nodes.tags->'location:transition' = 'yes') AND
-        NOT (nodes.tags?'transformer' AND nodes.tags->'transformer' in ('distribution', 'main')) AND
+        NOT (nodes.tags?'transformer' AND nodes.tags->'transformer' IN ('distribution', 'main')) AND
         NOT (nodes.tags?'substation' AND nodes.tags->'substation' = 'minor_distribution') AND
         NOT (nodes.tags?'line_management' AND nodes.tags->'line_management' IN ('transition','termination')) AND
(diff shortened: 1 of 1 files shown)


flacombe:
Lint errors on `mapcss/mapcss2osmose.py` are not on me

Famlam (approved):

frodrigo:
Ok, Thank you

PR #2759, opened 2026-06-07
Fix class 1 already used

didier2020:
class 1 is already used by Osmosis_Duplicated_Geotag


Famlam (approved):
Looks good to me. Glad I didn't start marking false positives yet as they'll probably be reset ;)

frodrigo:
Thank you

PR #2760, opened 2026-06-12
Detect incorrect `crossing:island=yes`

KTibow:
## Summary

Add an Osmosis analyzer that reports split crossing ways tagged `crossing:island=yes` when they connect to a separately mapped `footway=traffic_island`.

The check covers crossing ways tagged with any of:

- `footway=crossing`
- `cycleway=crossing`
- `path=crossing`

The issue reports both involved ways so mappers can review the crossing and the mapped traffic island together.

For some examples of what this targets, see https://maproulette.org/browse/challenges/55365.

## Rationale

When a traffic island is mapped as its own way, the short crossing ways leading to it usually should not also be tagged as having an island themselves. In those cases `crossing:island=yes` is redundant or misleading on the split crossing segment.

https://wiki.openstreetmap.org/wiki/Key:crossing:island

## Testing

- `python -m py_compile analysers/analyser_osmosis_highway_traffic_island.py osmose_config.py`
- `uv run --with polib --with psycopg2-binary --with python-dateutil --with lockfile --with shapely --with regex --with requests --with pyproj --with Unidecode --with osmium --with xmltodict --with termcolor python -c "import analysers.analyser_osmosis_highway_traffic_island as m; print(m.sql10.format('', ''))"`
- `uv run --with polib --with psycopg2-binary --with python-dateutil --with lockfile --with shapely --with regex --with requests --with pyproj --with Unidecode --with osmium --with xmltodict --with termcolor python -m osmose_config`
- `uv run --with 'git+https://github.com/ccxcz/pylama.git@14200a96c46acde592ee3effc24bc82a4217026d' --with polib --with psycopg2-binary --with python-dateutil --with lockfile --with shapely --with regex --with requests --with pyproj --with Unidecode --with osmium --with xmltodict --with termcolor pylama analysers/analyser_osmosis_highway_traffic_island.py osmose_config.py`
- `git diff --check`

## Disclosure

Generated by GPT 5.5 with human oversight.

### DIFF
--- a/analysers/analyser_osmosis_highway_traffic_island.py
+++ b/analysers/analyser_osmosis_highway_traffic_island.py
@@ -58,7 +58,7 @@ class Analyser_Osmosis_Highway_Traffic_Island(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
-        self.classs_change[11] = self.def_class(item = 3040, level = 3, tags = ['tag', 'highway', 'footway', 'fix:survey'],
+        self.classs_change[1] = self.def_class(item = 3040, level = 3, tags = ['tag', 'highway', 'footway', 'fix:survey'],
             title = T_('Crossing island tag on a split crossing'),
             detail = T_(
 '''A crossing way is tagged with `crossing:island=yes` and connects to a
@@ -77,7 +77,7 @@ as one continuous crossing, keep `crossing:island=yes`.'''),
 '''If you are not sure whether the crossing way ends at the island or continues
 over it, do not change the tag.'''))
 
-        self.callback10 = lambda res: {"class": 11, "data": [self.way_full, self.way_full, self.positionAsText]}
+        self.callback10 = lambda res: {"class": 1, "data": [self.way_full, self.way_full, self.positionAsText]}
 
     def analyser_osmosis_full(self):
         self.run(sql10.format("", ""), self.callback10)
@@ -108,7 +108,7 @@ class Test(TestAnalyserOsmosis):
             a.analyser()
 
         self.root_err = self.load_errors()
-        self.check_err(cl="11", elems=[("way", "100"), ("way", "200")])
-        self.check_err(cl="11", elems=[("way", "110"), ("way", "210")])
-        self.check_err(cl="11", elems=[("way", "120"), ("way", "220")])
+        self.check_err(cl="1", elems=[("way", "100"), ("way", "200")])
+        self.check_err(cl="1", elems=[("way", "110"), ("way", "210")])
+        self.check_err(cl="1", elems=[("way", "120"), ("way", "220")])
         self.check_num_err(3)
--- a/analysers/analyser_osmosis_highway_traffic_island.py
+++ b/analysers/analyser_osmosis_highway_traffic_island.py
@@ -52,7 +52,7 @@ class Analyser_Osmosis_Highway_Traffic_Island(Analyser_Osmosis):
 
     def __init__(self, config, logger = None):
         Analyser_Osmosis.__init__(self, config, logger)
-        self.classs_change[1] = self.def_class(item = 2090, level = 3, tags = ['tag', 'highway', 'footway', 'fix:survey'],
+        self.classs_change[11] = self.def_class(item = 2090, level = 3, tags = ['tag', 'highway', 'footway', 'fix:survey'],
             title = T_('Crossing island tag on a split crossing'),
             detail = T_(
 '''A crossing way is tagged with `crossing:island=yes` and connects to a
@@ -71,7 +71,7 @@ as one continuous crossing, keep `crossing:island=yes`.'''),
 '''If you are not sure whether the crossing way ends at the island or continues
 over it, do not change the tag.'''))
 
-        self.callback10 = lambda res: {"class": 1, "data": [self.way_full, self.way_full, self.positionAsText]}
+        self.callback10 = lambda res: {"class": 11, "data": [self.way_full, self.way_full, self.positionAsText]}
 
     def analyser_osmosis_full(self):
         self.run(sql10.format("", ""), self.callback10)
@@ -101,7 +101,7 @@ class Test(TestAnalyserOsmosis):
             a.analyser()
 
         self.root_err = self.load_errors()
-        self.check_err(cl="1", elems=[("way", "100"), ("way", "200")])
-        self.check_err(cl="1", elems=[("way", "110"), ("way", "210")])
-        self.check_err(cl="1", elems=[("way", "120"), ("way", "220")])
+        self.check_err(cl="11", elems=[("way", "100"), ("way", "200")])
+        self.check_err(cl="11", elems=[("way", "110"), ("way", "210")])
+        self.check_err(cl="11", elems=[("way", "120"), ("way", "220")])
         self.check_num_err(3)
--- a/analysers/analyser_osmosis_highway_traffic_island.py
+++ b/analysers/analyser_osmosis_highway_traffic_island.py
@@ -49,8 +49,7 @@ WHERE
         crossing.tags->'path' = 'crossing'
     )
 ORDER BY
-    crossing.id,
-    traffic_island.id
+    crossing.id
 """
 
 
@@ -67,10 +66,10 @@ class Analyser_Osmosis_Highway_Traffic_Island(Analyser_Osmosis):
 '''A crossing way is tagged with `crossing:island=yes` and connects to a
 separately mapped `footway=traffic_island`.
 
-When a traffic island is mapped as its own way, the short crossing ways leading
+When a traffic island is mapped as its own way, the crossing ways leading
 to it usually should not also be tagged as having a crossing island.'''),
             fix = T_(
-'''Check the highlighted crossing way. If it is only a short segment leading to
+'''Check the highlighted crossing way. If it is a segment leading to
 the separately mapped traffic island, remove `crossing:island=yes`. You may add
 `crossing:island=no` for clarity.
 
--- a/analysers/analyser_osmosis_highway_traffic_island.py
+++ b/analysers/analyser_osmosis_highway_traffic_island.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+
+###########################################################################
+##                                                                       ##
+## Copyright Osmose Contributors 2026                                    ##
+##                                                                       ##
+## This program is free software: you can redistribute it and/or modify  ##
+## it under the terms of the GNU General Public License as published by  ##
+## the Free Software Foundation, either version 3 of the License, or     ##
+## (at your option) any later version.                                   ##
+##                                                                       ##
+## This program is distributed in the hope that it will be useful,       ##
+## but WITHOUT ANY WARRANTY; without even the implied warranty of        ##
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         ##
+## GNU General Public License for more details.                          ##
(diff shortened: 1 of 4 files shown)


frodrigo:
Have you tested it with a meduim sized extract ?

KTibow:
Just did against a medium Geofabrik extract: **Estonia** (`europe/estonia-latest.osm.pbf`, 116 MiB). Ran only the `osmosis_highway_traffic_island` analyser with `--skip-upload --skip-frontend-check --result-format=geojson`. First run (cold, with DB import) took 8m32s total. Second run (already-loaded DB, `--skip-init`) took **10.26s** - the SQL + result generation itself was about 1s; most overhead was polygon service retries/grants.

**How many found in Estonia:**  
**88 issues** (class 11). Verified with an independent pyosmium scan and the count matched exactly.

**One example found:**  
Crossing way https://www.openstreetmap.org/way/51972104 (footway=crossing, crossing:island=yes) connected to island way https://www.openstreetmap.org/way/1163414159 (footway=traffic_island). Good hit for the rule.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,107 @@
...
+SELECT DISTINCT
+    crossing.id,
+    traffic_island.id,
+    ST_AsText(way_locate(crossing.linestring))
+FROM
+    {0}ways AS crossing
+    JOIN {1}ways AS traffic_island ON
+        traffic_island.id != crossing.id AND
+        traffic_island.nodes && crossing.nodes AND
+        traffic_island.tags->'footway' = 'traffic_island'
```
frodrigo, line 40: Before using tags-> you should use tag? to trigger the index.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,107 @@
...
+sql10 = """
+SELECT DISTINCT
+    crossing.id,
+    traffic_island.id,
+    ST_AsText(way_locate(crossing.linestring))
+FROM
+    {0}ways AS crossing
+    JOIN {1}ways AS traffic_island ON
+        traffic_island.id != crossing.id AND
+        traffic_island.nodes && crossing.nodes AND
```
frodrigo, line 37: You probably need the include a spatial index.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,107 @@
...
+## along with this program.  If not, see <http://www.gnu.org/licenses/>. ##
+##                                                                       ##
+###########################################################################
+
+from modules.OsmoseTranslation import T_
+from .Analyser_Osmosis import Analyser_Osmosis
+
+
+sql10 = """
+SELECT DISTINCT
```
frodrigo: Propably need a distind on to avoid duplicate result

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -25,23 +25,32 @@
 
 
 sql10 = """
-SELECT DISTINCT
+SELECT DISTINCT ON (crossing.id)
     crossing.id,
     traffic_island.id,
```
frodrigo, line 30: But it should rather be uniq on `traffic_island`, no ?

KTibow: Well the problematic tag is on the crossing, right? So we count each crossing that needs to be updated as 1 issue. Should it dedupe by island instead?

frodrigo: You are right.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,116 @@
...
+    crossing.tags->'crossing:island' = 'yes' AND
+    crossing.tags?|ARRAY['footway', 'cycleway', 'path'] AND
+    (
+        crossing.tags->'footway' = 'crossing' OR
+        crossing.tags->'cycleway' = 'crossing' OR
+        crossing.tags->'path' = 'crossing'
+    )
+ORDER BY
+    crossing.id,
+    traffic_island.id
```
frodrigo, line 53:
```suggestion
    crossing.id
```

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, config, logger = None):
+        Analyser_Osmosis.__init__(self, config, logger)
+        self.classs_change[11] = self.def_class(item = 3040, level = 3, tags = ['tag', 'highway', 'footway', 'fix:survey'],
+            title = T_('Crossing island tag on a split crossing'),
+            detail = T_(
+'''A crossing way is tagged with `crossing:island=yes` and connects to a
+separately mapped `footway=traffic_island`.
+
+When a traffic island is mapped as its own way, the short crossing ways leading
```
frodrigo: There is no check about "short"
```suggestion
When a traffic island is mapped as its own way, the crossing ways leading
```

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,116 @@
...
+to it usually should not also be tagged as having a crossing island.'''),
+            fix = T_(
+'''Check the highlighted crossing way. If it is only a short segment leading to
```
frodrigo:
```suggestion
'''Check the highlighted crossing way. If it is a segment leading to
```

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+ORDER BY
+    crossing.id
+"""
+
+
+class Analyser_Osmosis_Highway_Traffic_Island(Analyser_Osmosis):
+
+    requires_tables_full = ['ways']
+    requires_tables_diff = ['ways', 'touched_ways', 'not_touched_ways']
```
frodrigo: No. That base table.

KTibow: Right, those go in the analyzer.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+ORDER BY
+    crossing.id
+"""
+
+
+class Analyser_Osmosis_Highway_Traffic_Island(Analyser_Osmosis):
+
+    requires_tables_full = ['ways']
+    requires_tables_diff = ['ways']
```
frodrigo: But nothing at all here

KTibow: Right, fixed.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,114 @@
...
+ORDER BY
+    crossing.id
+"""
+
+
+class Analyser_Osmosis_Highway_Traffic_Island(Analyser_Osmosis):
+
+    def __init__(self, config, logger = None):
+        Analyser_Osmosis.__init__(self, config, logger)
+        self.classs_change[11] = self.def_class(item = 3040, level = 3, tags = ['tag', 'highway', 'footway', 'fix:survey'],
```
frodrigo: Why do you set class=11 ? You can use class=1.

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+sql10 = """
+SELECT DISTINCT ON (crossing.id)
+    crossing.id,
+    traffic_island.id,
+    ST_AsText(way_locate(crossing.linestring))
+FROM
+    {0}ways AS crossing
+    JOIN {1}ways AS traffic_island ON
```
Famlam: Probably better to use table `highways`? Or is there a good reason for using the larger table `ways`?

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def analyser_osmosis_full(self):
+        self.run(sql10.format("", ""), self.callback10)
+
+    def analyser_osmosis_diff(self):
+        # Match issues when either the crossing way or the traffic island way is touched.
+        self.create_view_touched('ways', 'W')
+        self.create_view_not_touched('ways', 'W')
+        self.run(sql10.format("touched_", ""), self.callback10)
+        self.run(sql10.format("not_touched_", "touched_"), self.callback10)
```
Famlam: I think the (not) touched views already exist for both `ways` as well as (per my suggestion) `highways`?

tests/osmosis_highway_traffic_island.osm
```diff
@@ -0,0 +1,87 @@
...
+  <!-- Negative: shared traffic island, but crossing:island is not yes. -->
+  <way id='140' version='1' timestamp='2014-03-31T22:00:00Z'>
+    <nd ref='41' />
+    <nd ref='42' />
+    <tag k='footway' v='crossing' />
+  </way>
+  <way id='240' version='1' timestamp='2014-03-31T22:00:00Z'>
+    <nd ref='42' />
+    <nd ref='43' />
+    <tag k='footway' v='traffic_island' />
```
Famlam, line 95: In view of possible future expansions, can you make sure that the tagging is at least valid, meaning that there's also a relevant `highway` tag?

Famlam (changes requested):

analysers/analyser_osmosis_highway_traffic_island.py
```diff
@@ -0,0 +1,114 @@
...
+    crossing.tags?'crossing:island' AND
+    crossing.tags->'crossing:island' = 'yes' AND
+    crossing.tags?|ARRAY['footway', 'cycleway', 'path'] AND
+    (
+        crossing.tags->'footway' = 'crossing' OR
+        crossing.tags->'cycleway' = 'crossing' OR
+        crossing.tags->'path' = 'crossing'
+    )
+ORDER BY
+    crossing.id
```
Famlam: Shouldn't it also order by the island to ensure it always picks the same "other" way?

analysers/analyser_osmosis_highway_traffic_island.py
KTibow: Well I'm trying to match
- footway=crossing
- cycleway=crossing
- path=crossing

even if they don't have highway={footway,cycleway,path}.

analysers/analyser_osmosis_highway_traffic_island.py
KTibow: Legacy change trying to prevent collisions, fixed.

analysers/analyser_osmosis_highway_traffic_island.py
Famlam: I don't think we should consider such mistagged items here (especially not if it requires a larger table), but rather have such cases first pop up as their own dedicated error via the mapcss rules of JOSM (see https://josm.openstreetmap.de/browser/josm/trunk/resources/data/validator/combinations.mapcss#L978), which we import. That way we also ensure that that much more significant tagging error is the only (and therefore most urgent) issue to fix.

KTibow: Somehow didn't see this comment. Fair point, I've switched to the highways table.

Famlam: Can you also do this for the other ways table?

KTibow: Fair enough, `highway=footway` is usual.

frodrigo:
It is now ok for me.

Famlam (approved):

PR #2763, opened 2026-06-13
Support Data-fair compatibility mode for merge analysers

flacombe:
This PR proposes to support the ODS compatibility mode for Data-Fair platforms.
- Files are got from the corresponding endpoint on Data-Fair platforms on which compatibility mode is enabled
- Last update date is got from `dataUpdatedAt` field in main metadata information for the given dataset.

It also needs to remove BOM from CSV headers

It will restore `merge_power_pole_FR_spec_eneids` and `merge_power_substation_minor_FR` analysers.

### DIFF
--- a/analysers/Analyser_Merge.py
+++ b/analysers/Analyser_Merge.py
@@ -534,6 +534,28 @@ class SourceDataFair(Source):
 
         return datetime.datetime.fromisoformat(matching_file["updatedAt"][0:10])
 
+class SourceDataFairCompatOds(Source):
+    def __init__(self,
+                 url: str,
+                 select: str,
+                 format: str = "csv",
+                 where: Optional[str] = None,
+                 **kwargs):
+        self.url_base = url.replace("/datasets/", "/data-fair/api/v1/datasets/")
+        kwargs.update({
+            "encoding": "utf-8-sig",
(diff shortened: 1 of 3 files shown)


frodrigo:
Interesting hack.

To remove the BOM in python, rather use the `utf-8-sig` encoding.

flacombe:
It looks like this PR is not responsible for test errors. It's ready to be reviewed

> rather use the utf-8-sig encoding.

On this line? `hdr = next(csv.reader(self.f, delimiter=self.separator, quotechar=self.quote))`

frodrigo:
> > rather use the utf-8-sig encoding.
> 
> On this line? `hdr = next(csv.reader(self.f, delimiter=self.separator, quotechar=self.quote))`

No, use the source encoding param.
```
hdr[0] = hdr[0].lstrip('\ufeff') # Remove UTF8 BOM from header
```

flacombe:
> the source encoding param.

Which class holds this param?

frodrigo:
https://github.com/osmose-qa/osmose-backend/blob/dev/analysers/Analyser_Merge.py#L336

flacombe:
Indeed, it works well

PR #2770, opened 2026-06-28
Split Denmark in Regioners

jocelynj:
Using this page as reference:
https://wiki.openstreetmap.org/wiki/Denmark/Da:Administrative_boundaries#Regioner

Note that extracts are not yet generated on download.openstreetmap.fr, but diffs have started:
https://download.openstreetmap.fr/replication/europe/denmark/

### DIFF
--- a/osmose_config.py
+++ b/osmose_config.py
@@ -732,7 +732,16 @@ default_country("europe", "united_kingdom_scotland", 58446, {"country": "GB-SCT"
 
 iceland = default_country("europe","iceland", 299133, {"country": "IS", "language": "is", "proj": 32627}, download_repo=GEOFABRIK) # 299133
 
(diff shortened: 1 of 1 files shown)


osmose_config.py
```diff
@@ -732,7 +732,16 @@ class gen(default_country):
 
 iceland = default_country("europe","iceland", 299133, {"country": "IS", "language": "is", "proj": 32627}, download_repo=GEOFABRIK) # 299133
 
-default_country("europe", "denmark",  50046, {"country": "DK", "language": "da","proj": 32632, "phone_code": '45', "phone_len": 8, "phone_international": '00'}, download_repo=GEOFABRIK)
+#########################################################################
+
+de_part = gen_country('europe', 'denmark', download_repo=OSMFR, language='da', proj=32632,
```
Famlam: Could it be dk_part to not confuse it with germany (de)?

jocelynj: Yes, that’s a good idea. I’ve renamed to `dk_part()`, as it is the ISO code for Denmark.

Issue #2725, opened 2026-04-07
Multiple identical `socket:*=*` values with different outputs


Oivo35:
The validator flags multiple `socket:*=*` connectors with the same number as duplicates, even if they have different outputs.

https://osmose.openstreetmap.fr/en/issue/2159d1db-7788-1392-2935-9559255e01eb
Duplicated values socket:type2_combo=10;10

Other tags of element:
`socket:type2_combo:output=50 kW;150 kW`
https://www.openstreetmap.org/way/1341272846

The validator should also take into account the `socket:*:output=*` tag.

frodrigo:
ping @nlehuby

Marc-marc-marc:
socket:type2_combo=10;10 is ambiguous: how many socket type2_combo ? 10;10 isn't a good value, it's 10 or 20 ?
socket:type2_combo:output=50 kW;150 kW is also wrong : :output describe "The maximum power", so if a socket is 50 kW and the other 150 kW, the value is 150 kW

so the two reports therefore represent two true anomalies in the data

Oivo35:
From OSM wiki:
>Note: Semi-colon separated values can be used if required. For example, use [socket:type2_combo](https://wiki.openstreetmap.org/wiki/Key:socket)=6;3 with [socket:type2_combo:output](https://wiki.openstreetmap.org/wiki/Key:socket)=350 kW;90 kW for a location with a total of 9x Type 2 CCS cables, six of which can deliver up to 350 kW, while the other 3 can deliver up to 90 kW.

Osmose validator doesn't flag, for example, `socket:type2_combo=3;5` and `socket:type2_combo=150 kW;50 kW`.

mattanelkaim:
> socket:type2_combo=10;10 is ambiguous: how many socket type2_combo ? 10;10 isn't a good value, it's 10 or 20 ? socket:type2_combo:output=50 kW;150 kW is also wrong : :output describe "The maximum power", so if a socket is 50 kW and the other 150 kW, the value is 150 kW
> 
> so the two reports therefore represent two true anomalies in the data

I must respectfully disagree with you @Marc-marc-marc. Looking at the rest of the [definition in wiki](https://wiki.openstreetmap.org/wiki/Key:socket:*#:~:text=The%20maximum%20power%20that%20can%20be%20sourced%20from%20a%20single%20socket%20of%20the%20specified%20type%20at%20that%20charging%20station.), the full quote is:
[...]

PR #2786, opened 2026-07-24
Allow `socket:*` to have duplicate values

mattanelkaim:
Please see #2725 

I agree that this issue shouldn't be flagged for the `socket:*` tag.

From OSM wiki ([`amenity=charging_station` page](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#:~:text=Note%3A%20Semi%2Dcolon%20separated%20values%20can%20be%20used%20if%20required.%20For%20example%2C%20use%20socket%3Atype2_combo%3D6%3B3%20with%20socket%3Atype2_combo%3Aoutput%3D350%20kW%3B90%20kW%20for%20a%20location%20with%20a%20total%20of%209x%20Type%202%20CCS%20cables%2C%20six%20of%20which%20can%20deliver%20up%20to%20350%20kW%2C%20while%20the%20other%203%20can%20deliver%20up%20to%2090%20kW.)):
>Note: Semi-colon separated values can be used if required. For example, use [socket:type2_combo](https://wiki.openstreetmap.org/wiki/Key:socket)=6;3 with [socket:type2_combo:output](https://wiki.openstreetmap.org/wiki/Key:socket)=350 kW;90 kW for a location with a total of 9x Type 2 CCS cables, six of which can deliver up to 350 kW, while the other 3 can deliver up to 90 kW.

### DIFF
--- a/plugins/TagFix_DuplicateValue.py
+++ b/plugins/TagFix_DuplicateValue.py
@@ -98,6 +98,7 @@ Ensure the interpretation of the tag does not change when you delete one item.''
             re.compile('lacounty:.+'),
             re.compile('turn:lanes.*'),
             re.compile('.+:conditional'), # More thorough check in ConditionalRestrictions plugin
+            re.compile('socket:.+'),
         ))
         self.WhitelistSimilarRegex = set(( # Regexes for keys that can have similar, but not equal values
             re.compile('.+_name'),


frodrigo:
@nlehuby @tykayn what do you think about this change and the linked issue ?

nlehuby:
I think in such cases, it would be better to add the `socket:*` tags on the individual man_made=charge_point because this tagging seems very confusing.

But anyway if it's documented, let's go for it.

tykayn:
Yup also finding that linking very confusing because we usually specify stuff in
the key part and do not expect values to be corresponding to some other
enumeration. 
So yeah, at least because it is documented we should not raise an alert. I do
not see any other tag doing this confusing behavior and do not understand how
the proposal to do so went accepted. 


Baptiste LEMOINE - Dirigeant de Cipher Bliss.com ,
-------

Tel 0185461173 / Signal 0627130837 , Telegram: Tykayn , Mastodon: @tykayn, Riot:
@tykaynchu:matrix.org N° SIRET: 79942416300035 GPG: 64A8 9B18 65E6 6523 FD86
7CB5 8796 1FCA F978 54FF





Sent from Proton Mail for Android.



-------- Original Message --------

frodrigo:
Ok. Thank you all.

Famlam:
This was unfortunately merged into the master rather than the dev branch

frodrigo:
> This was unfortunately merged into the master rather than the dev branch

Right. Thank you for report. I rebase dev on top of master and push force the dev barch.

PR #2823, opened 2026-09-12
Fix missing issues duplicated geotag

didier2020:
Add test and fix missing issues


didier2020:
can you revert (https://github.com/osmose-qa/osmose-backend/pull/2740)  ?
because it's no longer finds all the errors

